PR #31, opened 2021-09-15
fix:The dynamic size value is returned

DiegoContrerasA:
## [LEVEL]/[COMPONENT]
**TASK**: 

### Types of changes

- [ ] Documentation Only
- [ ] Refactor


jrgarciadev:
Hi! @DiegoContrerasA I appreciate your contribution but the border size can't be the same as the Avatar size because when the user puts a custom size like "40px" then the border size will be "40px"

Issue #32, opened 2021-09-16
Add ARIA attributes


jrgarciadev:
**Describe the bug**
Currently, the components don't have the ARIA attributes, this will have NextUI more accessible to people with disabilities.

**Expected behaviour**
Add Aria Attributes to components

**Additional info**
References:
[ARIA-Moz](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA)
[ARIA-Google](https://developers.google.com/web/fundamentals/accessibility/semantics-aria/)
[Chrome extension](https://chrome.google.com/webstore/detail/screen-reader/kgejglhpjiefppelpmljglcjbhoiplfn?hl=en)

DaniAcu:
I could take this 😉

jrgarciadev:
Thanks, @DaniAcu!

javierscode:
@jrgarciadev maybe we should split this issue by each component. Like this:

- Avatar: improve accessibility.
- Button: improve accessibility. 
- etc...

jrgarciadev:
I agree with you @javierscode I'm to create separate tasks for this

PR #70, opened 2021-10-07
fix(switch): focus access

DaniAcu:
## Switch

#32 

### Types of changes

- [ ] Documentation Only
- [ ] Refactor

Switches weren't accessible by focus, so I change a little bit the CSS to be able to do it. However, the focus continues not to show anything by design. We should find a way to show the focus stage in a visible way without breaking the system design


jrgarciadev:
Thank you @DaniAcu!

PR #105, opened 2021-10-31
Fix 38/button aria attributes

carlosriveroib:
## [fix]/[button]
**TASK**: [Fix 38 - Add aria attributes to button](https://github.com/nextui-org/nextui/issues/38)

### Types of changes

- [ ] Documentation Only
- [ ] Refactor

Add `aria-label` attribute to button with the value of its children
Add `role="presentation"` when there is only icon
Add `tabIndex=-1' if disabled or 0 is not passed as prop

### DIFF
--- a/packages/docs/content/docs/components/button.mdx
+++ b/packages/docs/content/docs/components/button.mdx
@@ -427,27 +427,28 @@ import { Button } from '@nextui-org/react';
 
 #### Button Props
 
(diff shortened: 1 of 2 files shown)


jrgarciadev:
Hey @carlosriveroib due to I needed this for the next NextUI version I implemented it for myself, thank you so much!

PR #200, opened 2022-01-31
Theme change button not work at first click

techhjork:
> localStorage.getItem("theme") checking is null or undefined in first render   
that's why i set default dark at render

## [LEVEL]/[COMPONENT]
**TASK**: 

### Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Documentation Only
- [x] Refactor

### Screenshots - Animations

Untitled design

### DIFF
--- a/apps/docs/src/pages/_app.tsx
+++ b/apps/docs/src/pages/_app.tsx
@@ -29,7 +29,10 @@ const KbarComponent = dynamic(() => import('../components/kbar'), {
 const Application: NextPage<AppProps<{}>> = ({ Component, pageProps }) => {
   const router = useRouter();
   const kbarActions = generateKbarActions(router);
-
+  
+  if(localStorage.getItem("theme") == undefined){
+    localStorage.setItem("theme","dark")
+  }
   return (
     <NextThemesProvider
       defaultTheme="system"


apps/docs/src/pages/_app.tsx
```diff
@@ -29,7 +29,10 @@ const KbarComponent = dynamic(() => import('../components/kbar'), {
 const Application: NextPage<AppProps<{}>> = ({ Component, pageProps }) => {
   const router = useRouter();
   const kbarActions = generateKbarActions(router);
-
+  
+  if(localStorage.getItem("theme") == undefined){
+    localStorage.setItem("theme","dark")
```
Joaovsa7, line 34: Just to prevent the application to throw some error if localStorage is not defined on SSR.

```suggestion
  if(localStorage?.getItem("theme") == undefined){
    localStorage.setItem("theme","dark")
```

erenkulaksiz: > Just to prevent the application to throw some error if localStorage is not defined on SSR.

jrgarciadev:
Hey @techhjork thanks for your contribution, however the problem was fixed here -> https://github.com/nextui-org/nextui/pull/209 and it's pending releasing

PR #276, opened 2022-02-12
fix(tooltip): Show no content if content is empty

erenkulaksiz:
## [LEVEL]/[COMPONENT]
**TASK**: Small fix that where tooltip is showed even when theres no content was inside it. Also added relevant story to storybook.

### Types of changes

- [ ] Documentation Only
- [ ] Refactor

The code

```javascript 
<Tooltip content={''} color="primary">
  <Button auto flat>
    Do hover here
  </Button>
</Tooltip>
```

Before the change 
tooltip-1
After the change
tooltip-2

### DIFF
--- a/packages/react/src/tooltip/__test__/index.test.tsx
+++ b/packages/react/src/tooltip/__test__/index.test.tsx
@@ -149,4 +149,13 @@ describe('Tooltip', () => {
     );
     expect(wrapper.find('#initial-visible').length).toBe(1);
   });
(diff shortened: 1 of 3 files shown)


jrgarciadev:
Hey @erenkulaksiz thank you for this PR, could you please write a test in order to not reduce the coverage and guarantee the functionality?

jrgarciadev (approved):

PR #300, opened 2022-02-24
Updated Avatar Component Style Sheet to fix Avatar Background Not Ada…

Dhananjay-JSR:
## [LEVEL]/[COMPONENT]
**TASK**: Update NextUI Avatar Background size

### Types of changes

- [ ] Documentation Only
- [x] Refactor

The avatar background is not adapting with the avatar size on custom sizes

### Screenshots - Animations

Untitled

### DIFF
--- a/packages/react/src/avatar/avatar.styles.ts
+++ b/packages/react/src/avatar/avatar.styles.ts
@@ -29,6 +29,13 @@ export const StyledAvatar = styled('span', {
     opacity: 1
   },
   '.nextui-avatar-bg': {
+    position: 'absolute',
+    top: 0,
+    left: 0,
+    right: 0,
+    bottom: 0,
+    zIndex: '$2',
+    transition: '$default',
     size: '100%'
   },
   '.nextui-avatar-icon': {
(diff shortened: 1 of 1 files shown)


packages/react/src/avatar/avatar.styles.ts
```diff
@@ -29,13 +29,7 @@ export const StyledAvatar = styled('span', {
...
   },
   '.nextui-avatar-bg': {
-    position: 'absolute',
-    top: 0,
-    left: 0,
-    right: 0,
-    bottom: 0,
-    zIndex: '$2',
-    transition: '$default'
+    size: '100%'
```
jrgarciadev, line 32: Sorry, the previous code didn't have to be removed, it has to be as follows

```
    position: 'absolute',
    top: 0,
    left: 0,
    right: 0,
    bottom: 0,
    zIndex: '$2',
    transition: '$default'
    size: '100%'
```

jrgarciadev (changes requested):

jrgarciadev (approved):

PR #325, opened 2022-03-03
Make `Pagination` properly controlled.

### LINKED ISSUE #324: [BUG] - Pagination is not properly controlled
### Describe the bug

It seems that the Pagination component is not properly controlled.  Even if I set the page props manually, it will use the inner state to render the active page.

https://codesandbox.io/embed/practical-knuth-5bty57?fontsize=14&hidenavigation=1&theme=dark

1. create a `Pagination` instance
2. create a `page` state
3. modify the `page` state from elsewhere
4. The `Pagination` will not change its' active page number.

### Expected behavior

The `Pagination` will sync to the `page` props.

### Screenshots or Videos

You can try in this simple [demo](https://codesandbox.io/embed/practical-knuth-5bty57?fontsize=14&hidenavigation=1&theme=dark) 

### Operating System Version

macOS Moneterey 12.2.1 (21D62)

### Browser

Edge

CNLHC:
## [LEVEL]/[COMPONENT]
**TASK**: Fix #324 

### Types of changes

- [ ] Documentation Only
- [ ] Refactor

it seems that the `page` prop is only used in a `useState` in the old version, which means the change of this prop will merely affect the render of the `Pagination` at the first time.

### DIFF
--- a/packages/react/src/use-pagination/index.ts
+++ b/packages/react/src/use-pagination/index.ts
@@ -25,7 +25,7 @@ const usePagination = ({
   const [activePage, setActivePage] = useState(page || initialPage);
 
   const onChangeActivePage = (newPage: number) => {
-    setActivePage(page ?? newPage);
+    setActivePage(newPage);
     onChange && onChange(newPage);
   };
 
--- a/packages/react/src/use-pagination/index.ts
+++ b/packages/react/src/use-pagination/index.ts
@@ -25,16 +25,12 @@ const usePagination = ({
   const [activePage, setActivePage] = useState(page || initialPage);
 
   const onChangeActivePage = (newPage: number) => {
-    setActivePage(newPage);
+    setActivePage(page ?? newPage);
     onChange && onChange(newPage);
   };
 
   const setPage = useCallback(
     (pageNumber: number) => {
-      if (page) {
-        setActivePage(page)
-        return
-      }
       if (pageNumber <= 0) {
         onChangeActivePage(1);
       } else if (pageNumber > total) {
--- a/packages/react/src/use-pagination/index.ts
+++ b/packages/react/src/use-pagination/index.ts
@@ -1,4 +1,4 @@
-import { useMemo, useCallback, useState } from 'react';
+import { useMemo, useCallback, useState, useEffect } from 'react';
 import { range } from '../utils/numbers';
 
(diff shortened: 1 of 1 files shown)


packages/react/src/use-pagination/index.ts
```diff
@@ -25,7 +25,7 @@ const usePagination = ({
   const [activePage, setActivePage] = useState(page || initialPage);
 
   const onChangeActivePage = (newPage: number) => {
-    setActivePage(newPage);
+    setActivePage(page ?? newPage);
```
jrgarciadev, line 28: Hey @CNLHC thanks for sending this PR, could you remove this?, instead add these lines:

```
import { useMemo, useCallback, useState, useEffect } from 'react';
....
  useEffect(() => {
    if (page && page !== activePage) {
      setActivePage(page);
    }
  }, [page]);
  ...
```
Because otherwise, the `activePage` will be out of date with respect to `page`

jrgarciadev (changes requested):

packages/react/src/use-pagination/index.ts
CNLHC: Thanks for your review. I think it is necessary to add a `useEffect` to set `activePage` when the `page` changes. 

But I think the `setActivePage(page ?? newPage);`  is also useful. 

For instance, if a user clicks the next button in a controlled `Pagination`, without this statement, the `activePage` will differ from the `page`. The expected way is to keep `page` and `activePage` synchronization and notify the `newPage` value using `onChange`, the parent component should provide an `onChange` callback that takes care of the state updating. Once the upper state is updated, the `useEffect` will  update the `activePage` to the `newPage`

jrgarciadev: Hey @CNLHC thank you for helping the project, I understand your point but the `activePage` won't defer from the `page` because `next` and `previous` methods use the `activePage` (which is synchronized in the `useEffect` with the `page` value) for changing the current page, so if the user wants to update its `page` has to listen to `onChange(newPage)` event, to sum up since the `useEffect` is updating the `activePage` when `page` changes, is not necessary to change it again in the `onChangeActivePage` method

jrgarciadev (changes requested):

jrgarciadev (approved):

PR #328, opened 2022-03-05
Disabled State should work as expected when declared in Button Group …

### LINKED ISSUE #319: `disabled` prop on a `Button` within a `Button.Group` does not show disabled visual state when using variants or colors
**Describe the bug**
When using a `Button` within a `Button.Group`, if the `Button` has a `disabled` prop, it will not show as visually-disabled, though it is still disabled in function.

**Steps to reproduce**
Reproduced using live editor:
```typescript
<>
    <Button.Group>
      <Button disabled>One</Button>
      <Button>Two</Button>
      <Button>Three</Button>
    </Button.Group>
    <Button.Group color="success">
      <Button disabled>One</Button>
      <Button>Two</Button>
      <Button>Three</Button>
    </Button.Group>
    <Button.Group color="gradient" ghost>
      <Button disabled>Action1</Button>
      <Button>Action2</Button>
      <Button>Action3</Button>
    </Button.Group>
    <Button.Group color="error" rounded>
      <Button disabled>One</Button>
      <Button>Two</Button>
      <Button>Three</Button>
    </Button.Group>
    <Button.Group color="primary" bordered>
      <Button disabled>Action1</Button>
      <Button>Action2</Button>
      <Button>Action3</Button>
    </Button.Group>
    <Button.Group color="warning" flat>
      <Button disabled>Action1</Button>
      <Button>Action2</Button>
      <Button>Action2</Button>
    </Button.Group>
    <Button.Group color="secondary" size="sm">
      <Button disabled>One</Button>
      <Button>Two</Button>
      <Button>Three</Button>
    </Button.Group>
    <Button.Group color="secondary" light>
      <Button disabled>One</Button>
      <Button>Two</Button>
      <Button>Three</Button>
    </Button.Group>
  </>
```

**Expected behaviour**
Buttons that are `disabled` will visually represent that state.

**Screenshots**
Using the code provided, the first button in each of these groups should show as disabled. Only the first one does this correctly and is the only one not using a `variant` or `color` prop.
image

**Environment / Context**
- Desktop or Mobile? Desktop
- Operating System: Windows 11
- Browser and version: Chrome 98
- Device: PC

**Additional info**
Setting colors for individual buttons also does not work at all in a button group.

Dhananjay-JSR:
## [LEVEL]/[COMPONENT]
**TASK**: This PR close  #319 

### Types of changes

- [ ] Documentation Only
- [x] Refactor

Whenever Disabled state was declared in Button Group Context , it wasn't working as intended as variants were given priority 

### Screenshots - Animations

Earlier :-
old

Now:- 

fixed

### DIFF
--- a/packages/react/src/button/button.styles.ts
+++ b/packages/react/src/button/button.styles.ts
@@ -180,8 +180,8 @@ export const StyledButton = styled(
       },
       disabled: {
         true: {
-          bg: '$accents2',
-          color: '$accents4',
+          bg: '$accents2 !important',
+          color: '$accents4 !important',
           cursor: 'not-allowed',
           pe: 'auto',
           '&:hover': {


packages/react/src/button/button.styles.ts
```diff
@@ -180,8 +180,8 @@ export const StyledButton = styled(
       },
       disabled: {
         true: {
-          bg: '$accents2',
-          color: '$accents4',
+          bg: '$accents2 !important',
```
jrgarciadev, line 183: hey @Dhananjay-JSR please let's try to avoid using `!import` since it limits the customization, instead we can use compound variants to guarantee the styles as follows

```
  // color=success && disabled
      {
        color: 'success',
        disabled: true,
        css: {
          bg: '$accents2',
          color: '$accents4'
        }
  },
  ... do the same with the rest of colors
```

On the other hand, we have to also create variants for "bordered", "gradient", "flat", "light" and "ghost" alongside "disabled" 🙏🏻

jrgarciadev (changes requested):

packages/react/src/button/button.styles.ts
```diff
@@ -763,7 +763,106 @@ export const StyledButton = styled(
...
+          color: '$accents4'
+        }
+      },
+      {
+        bordered: true,
+        disabled: true,
+        css :{
+          bg: '$accents2',
+          color: '$accents4',
+          borderStyle: 'none'
```
jrgarciadev, line 822: Instead of removing the `borderStyle` could you please add the `disabled` color to the `border` ?

Something like this:

```
 color: '$accents4',
 borderColor: '$accents2',
```

packages/react/src/button/button.styles.ts
```diff
@@ -763,7 +763,106 @@ export const StyledButton = styled(
...
+      {
+        flat: true,
+        disabled: true,
+        css :{
+          bg: '$accents2',
+          color: '$accents4'
+        }
+      },
+      {
+        color: 'light',
```
jrgarciadev, line 834: here is `light: true`

packages/react/src/button/button.styles.ts
```diff
@@ -763,7 +763,106 @@ export const StyledButton = styled(
...
+        disabled: true,
+        css :{
+          bg: '$accents2',
```
jrgarciadev, line 837: since `light` mode doesn't have a background is not necessary to apply a background color

packages/react/src/button/button.styles.ts
```diff
@@ -763,7 +763,106 @@ export const StyledButton = styled(
...
+          color: '$accents4'
+        }
+      },
+      {
+        rounded: true,
```
jrgarciadev, line 842: Remove this compound variant because `rounded` variant doesn't affect the colours

packages/react/src/button/button.styles.ts
```diff
@@ -763,7 +763,106 @@ export const StyledButton = styled(
...
+        css :{
+          bg: '$accents2',
+          color: '$accents4'
+        }
+      },
+      {
+        shadow: true,
+        disabled: true,
+        css :{
+          bg: '$accents2',
```
jrgarciadev, line 861: remove `bg` and `color` because they will be applied by the `color` & `disabled` compound variants

jrgarciadev (changes requested):

PR #417, opened 2022-04-19
Update customization.tsx

H-KE:
Saw a typo, so here's a proposed update to the copy.

## [LEVEL]/[COMPONENT]
**TASK**: 

### Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] Documentation Only
- [ ] Refactor

### DIFF
--- a/apps/docs/src/components/home/customization.tsx
+++ b/apps/docs/src/components/home/customization.tsx
@@ -37,7 +37,7 @@ const CustomizationSection = () => {
           <Title color="green">easy.</Title>
         </Row>
         <Subtitle>
-          Thanks to NextUI is based on the amazing CSS-in-JS library&nbsp;
+          Thanks to NextUI being based on the amazing CSS-in-JS library&nbsp;
           <Link
             href="https://stitches.dev/"
             rel="noreferer noopener"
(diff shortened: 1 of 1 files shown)


apps/docs/src/components/home/customization.tsx
```diff
@@ -37,7 +37,7 @@ const CustomizationSection = () => {
           <Title color="green">easy.</Title>
         </Row>
         <Subtitle>
-          Thanks to NextUI is based on the amazing CSS-in-JS library&nbsp;
+          Thanks to NextUI being based on the amazing CSS-in-JS library&nbsp;
```
jrgarciadev, line 40: Could we put here, `Is built on top of ... `?

apps/docs/src/components/home/customization.tsx
```diff
@@ -55,7 +55,7 @@ const CustomizationSection = () => {
           <NextLink href="/docs/theme/override-styles#using-the-styled-function">
             <Link css={{ color: '$green300' }}>styled&nbsp;</Link>
           </NextLink>
-          function or native css class names.
+          function or native css class name, style according to your style.
```
jrgarciadev, line 58: we could put here -> `function or native CSS selectors`

jrgarciadev (changes requested):
Hey @H-KE thanks for contributing to NextUI, I left you some comments ✋🏻

jrgarciadev:
Due to the delay the changes were applied here https://github.com/nextui-org/nextui/commit/f7df1e141f74adcbf75ae80d688f8b8e3da30f3a

Issue #439, opened 2022-04-28
[BUG] - Table with pagination is not rendered on the server side


tianenpang:
### Describe the bug

The server side does not render row data depending on `rowsPerPage`

### Your Example Website or App

https://codesandbox.io/s/next-ui-pagination-table-ssr-mt0boq

### Steps to Reproduce the Bug or Issue

1. Go to `index.js`
2. Open Devtools
3. Take look at the preview in Network

### Expected behavior

Get the correct SSR according to `rowsPerPage`

### Screenshots or Videos

![issue-screenshot](https://user-images.githubusercontent.com/32772271/165859643-d89b0f4f-9cc6-4e51-85cb-63e14d242a74.png)

### Operating System Version

Windows

### Browser

Chrome

PR #440, opened 2022-04-28
fix: #439 server side rendering of table with pagination

tianenpang:
Closes https://github.com/nextui-org/nextui/issues/439

## 📝 Description

Fix the issue that row data is not rendered on the server side.

Add `rowsPerPage` to the `initialValues` of `TableContext` so that `TableBody` can calculate `rows` correctly.

## 🚀 New behavior

Added `hasPaginationChild` method to check if there is a pagination and to get `rowsPerPage` by default `0`.

No

## 📝 Additional Information

Screenshot of table with pagination ssr

table-ssr-pr


jrgarciadev:
Huge thanks @tianenpang 🙏🏻

PR #463, opened 2022-05-09
Updated _document.tsx code to remove errors

NayamAmarshe:
Original Discussion: https://github.com/nextui-org/nextui/issues/441
and my original code: https://github.com/nextui-org/nextui/issues/441#issuecomment-1116955466

I posted the correct code but I suppose the PR by anoushk1234 forgot to include `ctx: any` at line 82.   
 Without `ctx: any`, the editor would throw an error because ctx requires a type.   
Explained here: https://stackoverflow.com/a/50414836/18995569


jrgarciadev:
Hey @NayamAmarshe, the current NextUI + Next.js example targets`Javascript` code, if you want to have the `_document` typescript implementation you can use the `Next.js` example https://nextjs.org/docs/advanced-features/custom-document#typescript, I hope to add the `Typescript` example soon as well

PR #466, opened 2022-05-10
fix(react-user): prop and link warnings

tianenpang:
Closes https://github.com/nextui-org/nextui/issues/474

## 📝 Description

- passed unnecessary props to `StyledUser`
> `Warning: Received true for a non-boolean attribute zoomed`
> `Warning: Received true for a non-boolean attribute pointer`

- `StyledUserLink` and `Link` both use the `a` tag
> `Warning: validateDOMNesting(...): <a> cannot appear as a descendant of <a>`

- structure `props` in order to pass the correct `otherProps`

- change the `a` tag of `StyledUserLink` to `span`

No

### DIFF
--- a/packages/react/src/user/user.styles.ts
+++ b/packages/react/src/user/user.styles.ts
@@ -38,7 +38,7 @@ export const StyledUserDesc = styled('span', {
   }
 });
 
-export const StyledUserLink = styled('a', {
+export const StyledUserLink = styled('span', {
   a: {
     '&:hover': {
       opacity: 0.7
--- a/packages/react/src/user/user.styles.ts
+++ b/packages/react/src/user/user.styles.ts
@@ -37,11 +37,3 @@ export const StyledUserDesc = styled('span', {
     mb: 0
   }
 });
-
-export const StyledUserLink = styled('span', {
-  a: {
-    '&:hover': {
-      opacity: 0.7
-    }
-  }
-});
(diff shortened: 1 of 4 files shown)


Cow258 (changes requested):

packages/react/src/user/user.styles.ts
```diff
@@ -38,7 +38,7 @@ export const StyledUserDesc = styled('span', {
   }
 });
 
-export const StyledUserLink = styled('a', {
+export const StyledUserLink = styled('span', {
   a: {
```
Cow258, line 42: If you change the tag into `span`, the style should also be changed together.

tianenpang: It' s already styled in the [`Link`](https://github.com/nextui-org/nextui/blob/3aa92a9e0b63a06821bded15c063620f46fc4535/packages/react/src/link/link.tsx#L26) component.

Hi @jrgarciadev We can remove the `StyledUserLink` and use `Link` component directly in the `UserLink` component, since the hover style already in [`StyledLink`](https://github.com/nextui-org/nextui/blob/3aa92a9e0b63a06821bded15c063620f46fc4535/packages/react/src/link/link.styles.ts#L11) and also get a cleaner dom, I can update it if it sounds good.

jrgarciadev: Hey @tianenpang yes, of course, it sounds good

tianenpang: Done.

jrgarciadev:
Huge thanks! @tianenpang 🙏🏻

jrgarciadev (approved):

PR #475, opened 2022-05-16
feat(avatar): allow to pass custom ref to img

### LINKED ISSUE #476: [Feature Request] Allow Avatar to pass custom ref to img element

I need access to the underlying `img` element. I want to use the `Avatar` component to be used with [`blurhash`,](https://blurha.sh) therefore access to the `img` element and `onLoad` event is required.

Use `useForkRef` hook that allows using two refs at once (as seen in [`@mui/utils`](https://github.com/mui/material-ui/blob/master/packages/mui-utils/src/useForkRef.ts) package). With this hook, ref from props and internal ref can be combined into one.

None

### Screenshots or Videos

_No response_

jansedlon:
Closes #476  

## 📝 Description

I have a use case where I need access to the underlying `img` element. At the moment, it's not possible to send own ref object.

There is no way to use own ref to get access to the `img` element.

## 🚀 New behavior

I used `useForkRef` hook from `@mui/utils` that allows to combine two refs into one. That enables components to use internal ref as well as ref passed from props.

No

### DIFF
--- a/packages/react/src/avatar/avatar.tsx
+++ b/packages/react/src/avatar/avatar.tsx
@@ -6,6 +6,7 @@ import { useDOMRef } from '../utils/dom';
 import { __DEV__ } from '../utils/assertion';
 import StyledAvatar, { AvatarVariantsProps } from './avatar.styles';
 import clsx from '../utils/clsx';
+import useForkRef from '../use-fork-ref';
 
 interface Props {
   text?: string;
@@ -13,6 +14,7 @@ interface Props {
   icon?: React.ReactNode;
   alt?: string;
   className?: string;
+  imgRef?: ReactRef<HTMLImageElement>;
   as?: keyof JSX.IntrinsicElements;
 }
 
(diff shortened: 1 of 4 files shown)


jrgarciadev:
Hey @jansedlon we have a function similar to `useForkRef` you can use instead, it's here https://github.com/nextui-org/nextui/blob/next/packages/react/src/utils/refs.ts#L38-L42, thanks a lot for contributing 🙏🏻

packages/react/src/avatar/avatar.tsx
```diff
@@ -32,14 +34,26 @@ const safeText = (text: string): string => {
...
+      ...otherProps
+    } = props;
 
     const domRef = useDOMRef(ref);
 
     const showText = !src;
     const [ready, setReady] = useState(false);
 
     const imgRef = useRef<HTMLImageElement>(null);
+    const handleImgRef = useForkRef<HTMLImageElement, HTMLImageElement>(
```
danielsimao, line 53: There is a function for this already in the utils folder called `useDomRef`. I would use that instead.

danielsimao (changes requested):

PR #502, opened 2022-05-28
Prevent anchor links from hiding behind navbar

imchell:
## 📝 Description

When clicking any section title (rendered by `LinkedHeading()`) in doc, the page jump to where a part of the section is hiding behind the navbar.

This is a common problem caused by sticky navbar, which can be solved by CSS scroll snap module.

Current behavior

Note that Size section actually hides behind navbar.

## 🚀 New behavior

New behavior


jrgarciadev (approved):
Huge thanks @imchell 🙏🏻

PR #508, opened 2022-06-05
fix: horizontal overflow

### LINKED ISSUE #281: UI bug in the home page with safari browser
**Describe the bug**
When I open the nextui.org website with Safari, there is a horizontal scroll on the home page that goes just to the right, and sometimes it gets stuck there so I need to refresh the page to fix it

**Screenshots**
Nextui

**Environment / Context**
- Desktop
- Operating System: [MacOs]
- Browser and version [Safari 15.3]
- Device: [Macbook air m1 2020]

imchell:
Closes #281

## 📝 Description

This was the problem mentioned in #281.

Homepage

## 🚀 New behavior

Homepage

### DIFF
--- a/apps/docs/src/styles/globalStyles.ts
+++ b/apps/docs/src/styles/globalStyles.ts
@@ -2,7 +2,8 @@ import { globalCss } from '@nextui-org/react';
 
 export default globalCss({
   html: {
-    scrollPaddingTop: '80px'
+    scrollPaddingTop: '80px',
+    overflowX: 'hidden'
   },
   // css baseline overrides
   blockquote: {


jrgarciadev:
Thanks a lot @imchell  🙏🏻

apps/docs/src/styles/globalStyles.ts
```diff
@@ -2,7 +2,8 @@ import { globalCss } from '@nextui-org/react';
 
 export default globalCss({
   html: {
-    scrollPaddingTop: '80px'
+    scrollPaddingTop: '80px',
+    overflowX: 'hidden'
```
tianenpang, line 6: The `overflow-x: hidden` causing the sticky navbar to not stick at the top.

It's possible to apply `overflow` to the [`main-container`](https://github.com/nextui-org/nextui/blob/f51149341bd5159ccd41b262befc9f70d780b65b/apps/docs/src/layouts/default.tsx#L25) to fix https://github.com/nextui-org/nextui/issues/281?

imchell: Thanks for pointing it out. 👍 I'm sorry I didn't test out for Chrome. Will deal with this later.

imchell: Putting `overflow-x` in `<Container />` is feasible. What do you think?

tianenpang: Sorry, I'm using Windows, can't test it for Safari 😅

PR #520, opened 2022-06-08
fix: card text selection

imchell:
## 📝 Description

Texts on card components can not be selected by mouse when they are not pressable.

Texts can not be selected nor copied.

## 🚀 New behavior

Texts can be selected and copied.

## 📝 Additional Information

This is a bug invoked by react-aria `usePress`. According to the author of `usePress()` ([explained here](https://github.com/adobe/react-spectrum/issues/2956#issuecomment-1069655946), which also involved the card implementation), we should avoid it when the card is not pressable. Correct me if I'm wrong 😂 not a pro of react-aria


jrgarciadev (approved):
Thanks! @imchell 🙏🏻

PR #542, opened 2022-06-12
Need Import React in Example

MrScrappy:
When trying to replicate the example it will fail since in the example the React instance is being requested in:

```jsx
     return {
                ...initialProps,
                styles: React.Children.toArray([initialProps.styles])
     //                  ^
};
  }
```

and this at no time matters

Closes # 

## 📝 Description

> When trying to replicate the example it will fail since in the example the React instance is being requested and this at no time matters

> I add the line import React from 'react'

## 🚀 New behavior

At the time of following the steps and copying and pasting the example, this error will stop appearing
image

## 📝 Additional Information

It's something simple that users can add, but for novice programmers it can be a headache to understand what is going wrong, and the idea that the learning curve or creating your project can be as simple as copy and paste.

### DIFF
--- a/apps/docs/content/docs/guide/nextui-plus-nextjs.mdx
+++ b/apps/docs/content/docs/guide/nextui-plus-nextjs.mdx
@@ -50,6 +50,7 @@ to create file `_document.js`.
 Then we add the following code to the file:
 
 ```jsx
+import React from 'react';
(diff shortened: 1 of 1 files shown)


RodrigoRVSN (approved):

jrgarciadev:
Hey @MrScrappy , thanks 🙌🏻

PR #549, opened 2022-06-17
fix(react-table): Table.Pagination component doesn't update with Table.Body

### LINKED ISSUE #543: [BUG] - Table.Pagination components doesn't update the number of pages displayed when the table data changes
### Describe the bug

I'm updating the data within a table through a dropdown selector for different sets of data.

Once the data being displayed updates, the number of pages shown by the pagination components doesn't reflect the change properly.

For example, if a dataset has less items than the previous, more pages than necessary are available for the user to choose from. Also, if the initial dataset had less items than the one selected by the user, additional data wouldn't span to extra pages and that date couldn't be accessed / displayed. I had to solve that issue by defaulting the data displayed to the dataset with more items in it, this way no data gets hidden but additional pages are shown on all of the other datasets.

https://buscavet.vercel.app/

1. Go to the website.
2. Click any of the following buttons: "Medicações", "Vacinas", "Suplementos" or "Desinfetantes".
3. Try changing the dataset displayed on the table below it through the dropdown menu.
4. Observe pages not updating accordingly to the ammount of rows being displayed.

### Expected behavior

The number of pages updates according to the number of rows contained in the table.
I'm not sure if this is something related to my implementation of that component, but I certainly expected it to work the way I did it...

### Screenshots or Videos

image
image
image

### Operating System Version

Linux (Fedora 36)

### Browser

Firefox

imchell:
Closes #543

## 📝 Description

Updating `<Table />` component can be a usual action, as mentioned in #543. However, `<Table.Pagination>` does not update synchronically with `<Table.Body />`, which results in two bugs:

1. `<Table.Pagination />` length doesn't change. So when rows number is larger after the update, some pages can not be selected via `<Table.Pagination />`.
2. When rows number is smaller comparing to the previous table, the original page of `<Table.Pagination />` is automatically selected and a blank table is showed.3. 

This is because:

1. The table context `collection` never updates when rows data change.
2. `<Pagination />` is not properly controlled as it does not receive the `page` prop.

#543 described it thoroughly. I've added a new story here so it would be clearer to test.

before

## 🚀 New behavior

after

### DIFF
--- a/packages/react/src/table/__tests__/index.test.tsx
+++ b/packages/react/src/table/__tests__/index.test.tsx
@@ -1,6 +1,8 @@
 import React from 'react';
 import { mount } from 'enzyme';
 import Table from '../index';
+import { updateWrapper } from 'tests/utils';
+import { getKeyValue } from '../../utils/object';
 
 const STITCHES_FACTOR = 2;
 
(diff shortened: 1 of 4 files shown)


packages/react/src/table/table-pagination.tsx
```diff
@@ -64,6 +65,7 @@ const TablePagination: React.FC<TablePaginationProps> = ({
       total={totalPagination}
       animated={animated}
       onChange={handlePageChanged}
+      page={currentPage}
```
danielsimao, line 68: How do you use this?

danielsimao (changes requested):

packages/react/src/table/table.stories.tsx
```diff
@@ -415,6 +417,63 @@ export const Pagination = () => {
   );
 };
 
+export const SwitchPagination = () => {
```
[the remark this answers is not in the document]
imchell: Thanks for reviewing it @danielsimao ! I think a test would be great but I'm not sure to what extent a test can do (forgive me I'm not a pro in test). I'm not a maintainer so you can send an individual PR I think?

packages/react/src/table/table-pagination.tsx
imchell: This is used to avoid losing control of `page` when switching between tables. The pagination would look like this without `page={currentPage}`:
test
Note that the pointer moves to the left arrow at first.

danielsimao:
As agreed with @imchell, I will add some test cases to his PR.

packages/react/src/table/__tests__/index.test.tsx
```diff
@@ -351,4 +353,69 @@ describe('Table', () => {
     expect(thirdColumnHeader.text()).toBe('DATE MODIFIED');
     expect(thirdColumnHeader.props()['aria-sort']).toBe('none');
   });
+
+  it('should reset pagination when amount of rows change', async () => {
```
danielsimao, line 357: @jrgarciadev probably this does not look great to you and I apologise, I'm not very familiar with enzyme. What I try to check here is that indeed the pagination reset when there is an items update.

jrgarciadev:
Fixed here https://github.com/nextui-org/nextui/pull/626

PR #563, opened 2022-06-22
fix(docs):default value of animated props in avatar group

shota8888:
## 📝 Description

Fix the docs from `false` to `true` because it is `animated: true` in avatar-group

No

## 📝 Additional Information
Fix the avatar-group.styles.ts file if the default value assumes false


danielsimao (approved):

jrgarciadev:
Thanks 🙏🏻  @shota8888

jrgarciadev (approved):

PR #580, opened 2022-06-27
fix(react-use-resize): add dependencies

Himself65:


tianenpang (commented):
Hi @Himself65 thanks for your contribution, but the dependencies cause `Tooltip` component to loop infinitely.

It seems to me that doesn't really need dependencies, as `useResize` jsut add/remove event listener when the component is mounted/unmounted or run `callback` immediately.

PR #602, opened 2022-07-06
bugfix:  input labelPlaceholder not center

KieSun:
## 📝 Description

Input `labelPlaceholder` not center. If I use `placeholder` props, the text is center, but `labelPlaceholder` not. I think them behavior should be same.

before

## 🚀 New behavior

image

No


KieSun:
@jrgarciadev

packages/react/src/input/input.styles.ts
```diff
@@ -567,7 +566,8 @@ export const StyledInputBlockLabel = styled('label', {
       underlined: true,
       asPlaceholder: true,
       css: {
-        left: '$space$2'
+        left: '$space$2',
+        top: '20%'
```
tianenpang, line 570: The `labelPlaceholder` is still a bit out of position (vertically) when it is `undelined`.

pr602-underline-label-placeholder

tianenpang (changes requested):
Hi @KieSun thank you for contributing, I left some comments.

packages/react/src/input/input.styles.ts
KieSun: Do you mean they all need to be centered?

tianenpang: Yep, it looks like all the Inputs that use `labelPlaceholder` have a shift of around 5 pixels up.

Hi @jrgarciadev is that a feature? 🤔

jrgarciadev: Hey guys sorry for the delay, that was done on purpose due to the top animation

PR #620, opened 2022-07-17
fix(docs): change media page title

imchell:
## 📝 Description

Original media page title was "Utilities".

## 🚀 New behavior

Rename media page title as "Media"


jrgarciadev (approved):
Thanks!! @imchell 🙌🏻

PR #626, opened 2022-07-19
fix(react-pagination): add active page to memoized callback deps

tianenpang:
Closes https://github.com/nextui-org/nextui/issues/619

## 📝 Description

Reading `state` in pagination's `onChange` will always return the initial value.

Add `activePage` to the memoized callback deps.

## 🚀 New behavior

Update or changes if `total, activePage` (deps) has been changed.

Read more: [`usecallback`](https://reactjs.org/docs/hooks-reference.html#usecallback).

No

## 📝 Additional Information

> **Note** The branch based on `next`.

### DIFF
--- a/packages/react/src/use-pagination/index.ts
+++ b/packages/react/src/use-pagination/index.ts
@@ -46,7 +46,7 @@ const usePagination = ({
         onChangeActivePage(pageNumber);
       }
     },
-    [total],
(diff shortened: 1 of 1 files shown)


honeysweetful:
onChange donot work when it receive a new value, and param page cannot bind a react state

PR #661, opened 2022-08-09
chore(docs) remove unused variable

TIMMLOPK:
Closes # 

## 📝 Description

> `selectedValue` never used and we don't need it

No


TIMMLOPK:
branch seem some problem🥲.I create another PR

PR #663, opened 2022-08-09
chore: upgrade react-aria

TIMMLOPK:
## 📝 Description

> new react-aria version has tree shaking improvements and has better bundle size.
> I think nextui also get benefits from it.

## 🚀 New behavior

Smaller bundle size 

## 💣 Is this a breaking change (Yes/No): No

ref discussion : https://github.com/nextui-org/nextui/discussions/611


TIMMLOPK:
I would try to fix build error 
ref https://github.com/adobe/react-spectrum/pull/3321

jrgarciadev (approved):
Thanks! 🙌🏻

PR #670, opened 2022-08-14
Pass input value to onContentClick handler

peteole:
I think it would be really convenient to pass the current value of an input field to the onContentClick handler for the use case that the content is a button that should submit the input value. I know there are other ways to solve this but directly passing the value in the handler seems by far the most simple version to me.

### DIFF
--- a/packages/react/src/input/input-props.ts
+++ b/packages/react/src/input/input-props.ts
@@ -48,7 +48,8 @@ export interface Props
   onBlur?: (e: React.FocusEvent<FormElement>) => void;
   onContentClick?: (
     key: ContentPosition,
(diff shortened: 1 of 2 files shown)


peteole:
What do you think about it?

jrgarciadev:
Hey @peteole I really appreciate your contribution, however, the input value is already being sent through the `onChange` function, on the other hand, the content element render will probably change once we migrate the input to `react-aria`, stay tuned 🙌🏼

peteole:
Ok, I get it. The idea was that the content would be used as a submit-button and then it would be useful to pass the content to be submitted to the callback. Maybe you can consider this in the migration.

PR #704, opened 2022-08-27
DOC: fix controlled / uncontrolled in Radio

yogabonito:
## 📝 Description

A component that gets a state and the corresponding change handler as a prop, is a controlled component. (Before this commit it was the other way around in the documentation for the Radio.Group component.)

No


jrgarciadev (approved):
Thank you! @yogabonito 🙌🏼

PR #741, opened 2022-09-06
fix: Input should not change value itself when value props is set (fix #737

### LINKED ISSUE #737: [BUG] - Components/Input
### Describe the bug

The input is uncontrolled input. When I defined value={'somethings'}  on input then I can change the value of input. That is not in normal input tag behaviour.

_No response_

The input is uncontrolled input. When I defined value={'somethings'}  on input then I can change the value of input. That is not in normal input tag behaviour.

### Expected behavior

The input is uncontrolled input. When I defined value={'somethings'}  on input then I can change the value of input. That is not in normal input tag behaviour.

### Screenshots or Videos

_No response_

### Operating System Version

Windows

### Browser

Chrome

sun0day:
Closes #737 

## 📝 Description

The controlled Input value should not be changed by Input when value prop is set.

When a non-undefined `value` prop is passed to `Input`, the `changeHandler` and `clearHandler` internal code will directly call `setSelfValue` to change `value`.

## 🚀 New behavior

When a non-undefined `value` prop is passed to `Input`, the `changeHandler` and `clearHandler` internal code will not call `setSelfValue` to change `value`.

No


OrhanTozan:
Can we please merge this? @jrgarciadev

jrgarciadev (approved):
Thanks a lot! @sun0day 🚀

PR #746, opened 2022-09-07
docs: make `invalid` radio code example more realistic

benmneb:
## 📝 Description

adds a more realistic example of "invalid" checkbox use

every option is invalid

## 🚀 New behavior

has both valid and invalid options

no

## 📝 Additional Information

thanks for the epic library


crypt096 (approved):

mel-habip (approved):
Indeed, this does make it more intuitive to comprehend & a more realistic use case as well.

jrgarciadev (approved):
Thanks! @benmneb

PR #748, opened 2022-09-07
fix(button): Button prop as in typescript

BlackSenPig:
## 📝 Description

> When we use button "as" in typescript ,the props will not extend.

>pass the wrong type
`Omit<HTMLNextUIProps<"button">` 

## 🚀 New behavior

>change to type button
`Omit<HTMLNextUIProps<button>` 

No


BlackSenPig:
update:
  Using forwardRef from import "/utils/system";

PR #749, opened 2022-09-07
Progress never full prop

OnlyCS:
## 📝 Description

> Added a prop that makes the progress bar never full. A use case would be when you set the max value to a really low number (say, 3), and you want to avoid the "why is it loading at 100%" response from users.


jrgarciadev:
Hey @OnlyCS thanks for contributing, unfortunately, I'll close this PR due to this can be handled by the developer outside the component

PR #783, opened 2022-09-22
Add name to switch input element.

rguarascia:
Useful when a switch appears in a form with conjunction with formData.

## 📝 Description
Add's the ability to pass in the `name` prop to `<Switch/>`, and passes it to the `input` element created inside the component. Doing this allows us to get the value using formData instead of relying on state controlled toggle. 

Cannot pass `name` prop to Switch so that the input has a name. 

## 🚀 New behavior
`<Switch name={"helloWorld"}/>` will render `<input name={"helloWorld"] ...restOfTheObject/>`. This is useful for when placing a Switch instead of a form, and need to extract the value with formData instead of using state + controlled.

no


jrgarciadev (approved):
Thanks! @rguarascia

PR #788, opened 2022-09-26
fix(react-radio) inaccessible prop types and duplicate class name

tianenpang:
Closes https://github.com/nextui-org/nextui/issues/759

## 📝 Description

- Export `radio` and `radio-group` prop types
- Fix duplicate class names

No


jrgarciadev:
🚀

PR #825, opened 2022-10-15
Broken Popover when disableAnimation = true

### LINKED ISSUE #797: [BUG] - Dropdown disableAnimation is broken
### Describe the bug

When disableAnimation is true, the dropdown menu just doesn't show, I've tested on the site as well as within my app across Chrome and Firefox just to make sure.

_No response_

Add the disableAnimation attribute.

### Expected behavior

As a user I expect the dropdown menu to open up without the animation.

### Screenshots or Videos

_No response_

### Operating System Version

Windows

### Browser

Firefox

0x2aff:
Closes #797

## 📝 Description

Unfortunately the popover without animation has problems with re-rendering after opening state change.

If we disable animation by setting arbitrary values in CSSTransition we enforce a re-render and react-aria is able to correctly define position state because the DOM element 'exists'.

When setting disableAnimation = true, popover / react-aria wont be able to define "left" / "right" positions because the DOM element is not rendered in time?! This means the opened content of the popover exists outside of the browser-view.

## 🚀 New behavior

Enforces re-rendering of Popover.Content while using arbitrary values with CSSTransition if disableAnimation = true.
Popover.Content will be visible with disableAnimation = true.

No

## 📝 Additional Information

This solution to the problem is as hacky as hacktober 2022 :)
Maybe it will help to find a proper solution or alteast provide a hackfix for the time being.

### DIFF
--- a/packages/react/src/popover/popover-content.tsx
+++ b/packages/react/src/popover/popover-content.tsx
@@ -1,13 +1,13 @@
-import React, {ReactNode} from "react";
-import {useModal, useOverlay, DismissButton} from "@react-aria/overlays";
-import {useDialog} from "@react-aria/dialog";
-import {FocusScope, useFocusRing} from "@react-aria/focus";
-import {mergeProps} from "@react-aria/utils";
+import React, { ReactNode } from "react";
+import { useModal, useOverlay, DismissButton } from "@react-aria/overlays";
+import { useDialog } from "@react-aria/dialog";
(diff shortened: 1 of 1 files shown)


jrgarciadev:
Hey @0x2aff I really appreciate you sending this PR, but unfortunately I'll close it due to we fix this internally https://github.com/nextui-org/nextui/pull/861, sorry 🙏🏻

PR #839, opened 2022-10-23
Fix: Update TableContext when items change (fix #798)

### LINKED ISSUE #798: [BUG] - Pagination Data Not Updated
### Describe the bug

When I change the page using state the new data is loaded using API but it's not updated in the table. It shows blank table.

_No response_

1. Implemented Pagination using Table

### Expected behavior

It should show the data of the different pages. It only shows data of the first page.

### Screenshots or Videos

Screenshot 2022-10-04 at 8 57 44 PM

### Operating System Version

macOS

### Browser

Chrome

Knorrke:
Closes #798 

## 📝 Description
Update `collection` in `TableContext` when items change to update dependent components like Table.Pagination.

Currently the `collection` are only initialized once, so even if you add items the Pagination won't get additional pages. Only Table.Body uses the `collections` from props.

## 🚀 New behavior
When `items` prop of Table.Body changes, `TableContext` is updated and Pagination will update the number of pages accordingly.

No


jrgarciadev:
Hey @tianenpang could you please help me check this? 🙏🏻

Knorrke:
If there's anything I can help with, please let me know.

tianenpang (commented):
Hi @Knorrke sorry for the late review and the fix looks great.

I found that a similar issue still exist when `Table.Pagination` has `total` prop configured, it may be related to [`handlePageChanged`](https://github.com/nextui-org/nextui/blob/f85a569b4782e99124c752269867272dc91f6313/packages/react/src/table/table-pagination.tsx#L52-L55) or [`setPage`](https://github.com/nextui-org/nextui/blob/f85a569b4782e99124c752269867272dc91f6313/packages/react/src/use-pagination/index.ts#L39-L50).

jrgarciadev (approved):
Huge thanks! @Knorrke 🙏🏻

PR #843, opened 2022-10-26
Add Table.Button component

### LINKED ISSUE #548: [Feature Request] Table pagination 'Load More' trigger

Currently we have two ways to paginate the table, using the Table.Pagination and infinity pagination.
I find myself needing a third way, an infinity pagination type of thing but instead of automatically fetching the data when reaching the bottom, displaying a 'Load More' button inside the table, this way I can read the last row of the table without making another unwanted request to the DB. 

I already made this myself but I can't put the load more button inside of the table so that is not always visible and you can only see it when you reach the bottom of the table.

Same behaviour as the infinity pagination but the trigger would be a 'Load More' button at the bottom of the table (like an extra row) instead of scrolling to the bottom.

Adding some kind of Table.Footer so that you can add a 'row' at the end of the table would solve this.

### Screenshots or Videos

Screen Shot 2022-06-16 at 15 48 22

Hdanzo:
Add documentation for new component
Add test for new component
Fix some warnings in table test file

Closes #548

## 📝 Description

Added the possibility to add a button to the end of a table.

## 🚀 New behavior

image

No

### DIFF
--- a/packages/react/src/table/table-button.tsx
+++ b/packages/react/src/table/table-button.tsx
@@ -1,0 +1,48 @@
+import React, {forwardRef} from "react";
+
+import {CSS} from "../theme/stitches.config";
+import {Button, ButtonProps} from "../index";
+import clsx from "../utils/clsx";
+import {useDOMRef} from "../utils/dom";
+import withDefaults from "../utils/with-defaults";
+
+import {useTableContext} from "./table-context";
+
+interface Props {
+  show?: boolean;
+}
+
+const defaultProps = {
+  show: true,
+};
+
+type NativeAttrs = Omit<React.HTMLAttributes<unknown>, keyof Props | "color">;
+
+export type TableButtonProps = Props & NativeAttrs & ButtonProps & {css?: CSS};
+
+const TableButton: React.FC<TableButtonProps> = forwardRef(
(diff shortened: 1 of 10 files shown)


packages/react/src/table/table-button.tsx
```diff
@@ -0,0 +1,48 @@
+import React, {forwardRef} from "react";
```
sun0day, line 1: Thank you for your contribution. But there is no need to add a `Table.Button`.  Just add a prop like `manually:boolean` to `Table` to control whether to manually trigger lazy loading.

Hdanzo: Okay, sure, I can work on adding it, are there any other criteria to take into account? I suppose the prop should be added to the `Table.Body` component

sun0day: I imagine the new API will be like `<Table.Body bottom={ReactNode} />` 

`bottom` only shows when `Table.Body` scrolls to the bottom. What do you think @jrgarciadev

Hdanzo: I am waiting for a reply to start working on this

jrgarciadev:
Hey @Hdanzo I appreciate you effort in sending this PR, however, the UI changes are made only by the `core` team since we discussed it the design and other stuff, sorry but I'll close it

PR #866, opened 2022-11-09
chore(react): react-aria dependecies upgraded in order to support react 18

jrgarciadev:
## 📝 Description

`react-aria` packages were upgraded


packages/react/package.json
```diff
@@ -54,37 +54,37 @@
...
-    "@react-aria/label": "3.4.0",
-    "@react-aria/link": "3.3.1",
-    "@react-aria/menu": "3.6.0",
-    "@react-aria/overlays": "3.10.0",
-    "@react-aria/radio": "3.3.0",
+    "@babel/runtime": "^7.20.1",
+    "@react-aria/button": "3.6.2",
+    "@react-aria/checkbox": "3.6.0",
+    "@react-aria/dialog": "3.4.0",
+    "@react-aria/focus": "3.9.0",
```
tianenpang, line 61: From [`v3.8.0`](https://github.com/adobe/react-spectrum/issues/3458) it will break the app and it could be useful to specify in the docs that strict-mode should be disabled.

packages/react/package.json
```diff
@@ -54,37 +54,37 @@
     "coverage:codecov": "yarn test --coverage && codecov"
   },
   "dependencies": {
-    "@babel/runtime": "^7.6.2",
-    "@react-aria/button": "3.6.0",
-    "@react-aria/checkbox": "3.5.0",
-    "@react-aria/dialog": "3.3.0",
-    "@react-aria/focus": "3.7.0",
```
tianenpang, line 61: Or still keep the version `3.7.0`, it works as usual.

tianenpang (changes requested):

PR #901, opened 2022-11-21
fix: resolve circle depents in docs

sun0day:
## 📝 Description

> Add a brief description

`yarn dev:docs` failed due to circle depents.

## 🚀 New behavior

error above disappeared.

### DIFF
--- a/apps/docs/content/landing/index.tsx
+++ b/apps/docs/content/landing/index.tsx
@@ -10,7 +10,7 @@ import {
   Maximize,
   CodeDocument,
   HtmlLogo,
-} from "@components";
+} from "@icons";
 
 export default {
   topFeatures: [
--- a/apps/docs/src/components/install-banner/install-banner.tsx
+++ b/apps/docs/src/components/install-banner/install-banner.tsx
@@ -1,6 +1,7 @@
 import React from "react";
 import {useRouter} from "next/router";
-import {FeaturesGrid, Note, NextJsLogo} from "@components";
+import {Note, NextJsLogo} from "@icons";
+import FeaturesGrid from '@components/features-grid'
 import {Button, Container, Grid, Snippet} from "@nextui-org/react";
 import {StyledCardBlur, Title, Subtitle} from "@primitives";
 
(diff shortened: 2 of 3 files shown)


jrgarciadev:
@tianenpang this issue happen to you? I couldn't replicate it 🤔

tianenpang:
I can't reproduce it either 🤔 hey @sun0day please remove `node_modules` and cache then try again.

```bash
# Remove all node_modules and cache include sub-dirs
find . '(' -name "node_modules" -o -name ".next" -o -name ".turbo" ')' -type d -prune -exec rm -rf '{}' +

# Install deps and build nextui
yarn && yarn build:nextui

# Restart docs dev server
yarn dev:docs
```

tianenpang (commented):
I can't reproduce it either 🤔 hey @sun0day please remove `node_modules` and cache then try again.

```bash
# Remove all node_modules and cache include sub-dirs
find . '(' -name "node_modules" -o -name ".next" -o -name ".turbo" ')' -type d -prune -exec rm -rf '{}' +

# Install deps and build nextui
yarn && yarn build:nextui

# Restart docs dev server
yarn dev:docs
```

sun0day:
> 

This error disappeared after cleaning cache and reinstalling dependencies.  But still, I think using circle depents is bad practice. We should avoid circle depents in case of some weird exceptions.

tianenpang:
I agree with you, but I think this error may be caused by other issues 🤔 and for the circular dependency, it's possible to use the webpack plugin to detect it automatically.

sun0day:
I am pretty sure this origin reason is circular depents due to the exception description and my debug process. But the circular depents exception not always occurs, it depends the node_modules' physical topology,  `yarn` can't guarantee node_modules' topology always be the same every time you install or add new dependencies, especially when you host has node_modules cache.

And maybe we can  use eslint rule [import/no-cycle](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-cycle.md) to discover the circular depends problems in dev stage.

jrgarciadev:
Hey @sun0day I'll close this PR until we can reproduce it in a better way

PR #923, opened 2022-11-28
fix: fixed issues of clear button inside inputs #920

### LINKED ISSUE #920: [BUG] - Accessibility issues of clear button inside inputs
### Describe the bug

- clear button does not have a focus state
- clear button does not have a label

_No response_

.

### Expected behavior

.

### Screenshots or Videos

.

### Operating System Version

.

### Browser

Chrome

jguddas:
Closes #920

## 📝 Description

1. this adds a focus indicator around the clear button for keyboards.
2. this adds a label to the clear button.

There is still an issue, the label for the clear button is hard code to "clear", and should really be localized to the language of the user.

No

## 📝 Additional Information

2022-11-28-221530_280x141_scrot


packages/react/src/input/input.styles.ts
```diff
@@ -1,5 +1,6 @@
 import {styled, VariantProps} from "../theme/stitches.config";
 import {sharedVisuallyHidden} from "../theme/shared-css";
+import {cssFocusVisible} from "../theme/shared-css";
```
gauravkrp, line 3: Can you move this import to the above? There is already an import from 'theme/shared-css'

jrgarciadev (approved):
Really appreciated 🙏🏻

PR #930, opened 2022-12-01
fix: fixed a11y role issue in progress bar component

jguddas:
Resolves the following violations:
- Elements must only use allowed ARIA attributes
- ARIA progressbar nodes must have an accessible name

2022-12-01-195721_1369x421_scrot


jrgarciadev (approved):
Thank you so much! @jguddas

PR #949, opened 2022-12-12
fix(react-theme): white border of the focus ring

tianenpang:
## 📝 Description

There is a white border in the animation of the focus ring caused by the `outline` property.

- Remove `outline` styles from the `isFocusVisible` variant.

No

## 📝 Additional Information

> **Note** Based on `next` branch.

Take `Radio`'s focus ring as an example, however, not all components have *white border* in the animation of the focus ring.

Before:

[before.webm](https://user-images.githubusercontent.com/32772271/207093070-c151dbf5-ad23-4df8-9cfa-7a6a48703306.webm)

After:

[after.webm](https://user-images.githubusercontent.com/32772271/207093110-cbb8c72c-f3d0-42dd-aaae-6ec710611610.webm)

### DIFF
--- a/packages/react/src/theme/shared-css.ts
+++ b/packages/react/src/theme/shared-css.ts
@@ -16,15 +16,13 @@ export const sharedFocus = css({
 });
 
 export const cssFocusVisible = css({
+  outline: "none",
   variants: {
     isFocusVisible: {
       true: {
         boxShadow: "0 0 0 2px $colors$background, 0 0 0 4px $colors$primary",
       },
-      false: {
-        outline: "none",
-        boxShadow: "none",
-      },
+      false: {},
     },
   },
 });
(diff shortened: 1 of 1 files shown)


jguddas:
There are a lot of unrelated changes in this PR that block me from testing that all the other components that use the focus ring still work.

packages/react/src/theme/shared-css.ts
```diff
@@ -19,12 +19,11 @@ export const cssFocusVisible = css({
   variants: {
     isFocusVisible: {
       true: {
-        outline: "transparent solid 2px",
```
jguddas, line 22: `outline: "none"` might be better, I don't see outline being unset anywhere in the button component, for example

jguddas (changes requested):

packages/react/src/theme/shared-css.ts
tianenpang: Thanks for the suggestion, I put `outline: "none"` outside the variants

jguddas (approved):

Issue #958, opened 2022-12-19
[BUG] - Progress bar component has clipping issue for small values


johannpayer:
### Describe the bug

For small values, the colored portion of the progress bar component clips out of the circular shape, making the colored area appear rectangular rather than round. For even smaller values, this issue becomes worse because the colored region begins to clearly "stick out" of the circular progress bar.

### Steps to Reproduce the Bug or Issue

StackBlitz is not loading for me, so I can't provide a link to an example, but the issue is very easy to replicate:

```tsx
<Progress value={5} />
```

### Expected behavior

For small values, the colored area of progress bar components should still be clipped to the circular region.

### Operating System Version

Windows, macOS

### Browser

Chrome

johannpayer:
I updated my bug report to link to an example on StackBlitz. Please let me know if you have any questions.

PR #961, opened 2022-12-21
fix: process overflow visible #958

wangly19:
## 📝 Description

The progress bar shows an exception.

exception: 

Kapture 2022-12-21 at 21 34 28

Correct
Kapture 2022-12-21 at 21 36 46

Setting small values makes the problem worse, as colored areas start to "highlight" circular progress bars visibly.

## 🚀 New behavior

Display normal

No

## 📝 Additional Information
For small values, the colored portion of the progress bar component clips out of the circular shape，
Is the design correct?

image

### DIFF
--- a/packages/react/src/progress/progress.styles.ts
+++ b/packages/react/src/progress/progress.styles.ts
@@ -14,7 +14,7 @@ export const StyledProgress = styled("div", {
   padding: 0,
   width: "100%",
   position: "relative",
-  overflow: "auto",
+  overflow: "hidden",
   variants: {
     color: {
       default: {
--- a/packages/react/src/progress/progress.styles.ts
+++ b/packages/react/src/progress/progress.styles.ts
@@ -14,7 +14,7 @@ export const StyledProgress = styled("div", {
   padding: 0,
   width: "100%",
   position: "relative",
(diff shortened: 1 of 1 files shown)


jrgarciadev:
@tianenpang could you please help me testing this?

packages/react/src/progress/progress.styles.ts
```diff
@@ -14,7 +14,7 @@ export const StyledProgress = styled("div", {
   padding: 0,
   width: "100%",
   position: "relative",
-  overflow: "visible",
+  overflow: "auto",
```
tianenpang, line 17: Hey @wangly19 thanks for your contribution. 

In my opinion, `overflow: "hidden"` would be more suitable.

tianenpang (changes requested):

packages/react/src/progress/progress.styles.ts
wangly19: I've revised it.

tianenpang:
Hi @jrgarciadev LGTM, ready to merge 🚀 thanks @wangly19.

Closes https://github.com/nextui-org/nextui/issues/958

PR #962, opened 2022-12-22
fix(docs-playground): dynamic live font styles

tianenpang:
## 📝 Description

The theme of sandpack makes `DynamicLive` applied `$sm` font-size and line-height, which is the font-size and line-height of `LiveCode`.

As the live preview is not the `$base` font-size and `$md` line-height, it causes some content shift.

https://github.com/nextui-org/nextui/blob/66153ecb82e4f4b838d41763e4f83c6f8ac95c2e/apps/docs/src/components/sandpack/themes.ts#L29-L34

Reapply the right font-size and line-height to `InView`.

No

## 📝 Additional Information

Before:

[before.webm](https://user-images.githubusercontent.com/32772271/209175226-288cfd1f-2fa4-4e3f-b8c3-886c679b5783.webm)

After:

[after.webm](https://user-images.githubusercontent.com/32772271/209175261-e579bfe0-cbd3-470c-8107-935d34029d62.webm)

### DIFF
--- a/apps/docs/src/components/playground/dynamic-live.tsx
+++ b/apps/docs/src/components/playground/dynamic-live.tsx
@@ -189,6 +189,12 @@ const LiveContainer = Components.styled(Box, {
   },
 });
 
+const StyledInView = Components.styled(InView, {
+  fontSize: "$base",
+  lineHeight: "$md",
+  fontFamily: "$sans",
+});
+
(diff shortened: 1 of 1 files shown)


jrgarciadev (approved):
Thanks! @tianenpang ❤️

PR #1041, opened 2023-02-11
fix: invalid role value

ImLunaHey:
Closes # 

## 📝 Description
While running a lighthouse test I noticed it warned about `"section"` not being valid.
Looking on [MDN](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/section_role) it looks like the correct value would be `"region"` instead.

Changes `role` from `"section"` to `"region"`.

## 🚀 New behavior
None

No


jrgarciadev (approved):
Thank you! @ImLunaHey

PR #1049, opened 2023-02-18
Feat/card tv migration

gijsbotje:
Hi there,

I wanted to contribute to the V2 rewrite and decided to start on the Card component. I'm still a little unsure if i'm on the right track, so I got a few questions.

Am I correct to use the tailwind state classes instead of the props?
For example
```
// packages/core/theme/src/components/card/index.ts
isHoverable: {
  true: "hover:drop-shadow-lg hover:-translate-y-0.5 transition",
},
isPressable: {
  true: "cursor-pointer active:scale-95",
},
```

I've just seen the changes for the AvatarGroup on the feat/v2 branch. Do you expect a hook for the `CardBody`, `CardFooter` and `CardHeader`? Or is just adding the classes with clsx, in for example `CardFooter.tsx`, enough?
Example:
```
// packages/components/card/src/card-footer.tsx
    <Component
      ref={domRef}
      {...otherProps}
      className={clsx(
        "w-full h-auto p-3 flex items-center overflow-hidden color-inherit rounded-b-xl",
        {
          "backdrop-blur-md backdrop-saturate-[1.8]": isBlurred,
        },
        className,
      )}
    >
      {children}
    </Component>
```

any other guidance you can give me on how to migrate the components would be also welcome, I would love to help out on the other components!


jrgarciadev:
Hey @gijsbotje thank you so much for helping us with the migration, it looks good to me, you're in the right direction, the idea is to use the TailwindCSS states `hover` `active` instead of the `react-aria` so in the future we will be able to use the same styles in another framework,  regarding if I expect a hook for each card part, no, I don't, but it would be great to have the card splitter by slots, for example:


```
import {tv} from 'tailwind-variants'

const card = tv({
  slots: {
     base: '....',
     header: '...',
     body: '...',
     footer: '...'
  }
})
```

You could pass the classes to the children by using the utility [`pickChild`](https://github.com/nextui-org/nextui/blob/main/packages/react/src/utils/collections.ts#L27-L46) or by using a context, let me know if you have any question, feel free to reach me on Discord

Thank you! 🙏🏻

gijsbotje:
@jrgarciadev aah yeah the slots was one of the things which I wasn't sure about. I'll change that up to use the slots. If I have more questions i'll ask them om Discord 👍🏻

jrgarciadev (approved):
Huge thanks @gijsbotje , I will continue the missing parts, really appreciated it 🙌🏻

jrgarciadev:
Hey @gijsbotje as I don't have access to your repository I had to commit directly to the feat/v2 branch with your changes, sorry about that https://github.com/nextui-org/nextui/commit/91cb12b8bdd8510933c0fecee23e95d8c3cb4206, I will include you in the release notes ❤️

PR #1051, opened 2023-02-22
Used labeled tuple for CalculatedNodeHeights instead of number array

theothirteen:
Closes ToDo item

## 📝 Description

> Used tuple instead of the number array

>  Returns number array

## 🚀 New behavior

> Returns labeled tuple


jrgarciadev (approved):
Thanks! @theothirteen

PR #1073, opened 2023-03-16
fix(react-link): update example by handle next 13

hyesungoh:
## 📝 Description

- Using `linkContent` at link - Next.js Link
  - there are already exist file, but not using it

- next.js 13 handling example
  - current example cause hydration error
  - there are two way to handle it, what I found
    - using `legacyBehavior` prop
    - using `Link` tags 'as' prop with `NextLink`

I found this PR, so I handle with this as prop https://github.com/nextui-org/nextui/pull/446

스크린샷 2023-03-16 오후 4 16 55

## 🚀 New behavior

image

No. just changin docs

## 📝 Additional Information

does not work example `Go to Button`, I do not know why 😢 

> I guess... are there not allowed importing 'next/link' where linkContent mdx files?

### DIFF
--- a/apps/docs/content/components/link/nextLink.ts
+++ b/apps/docs/content/components/link/nextLink.ts
@@ -1,12 +1,11 @@
 const App = `import { Link } from "@nextui-org/react";
+import NextLink from "next/link";
 
 export default function App() {
   return (
-    <NextLink href="/docs/components/button">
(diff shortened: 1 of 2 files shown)


jrgarciadev:
Closing this due to v1 will not receive more updates.

PR #1088, opened 2023-04-09
Fix: #1085 - Updated Textarea component to proper resize when controlled

### LINKED ISSUE #1085: [BUG] - Components/Textarea
### Describe the bug

Textarea autosize works bad if we pass onChange callback that changes state

https://codesandbox.io/p/sandbox/dreamy-faraday-b9gkdi?file=%2Fsrc%2FApp.tsx&selection=%5B%7B%22endColumn%22%3A14%2C%22endLineNumber%22%3A11%2C%22startColumn%22%3A14%2C%22startLineNumber%22%3A11%7D%5D

1. Go to provided codesanbox examle
2. Write some text in two fields, use linebreaks to enlarge textareas

### Expected behavior

As a user, I expect textearea with onChange(() => setState) callback enlarge the same way as textearea without it does.  

### Screenshots or Videos

image

### Operating System Version

Windows 10

### Browser

Chrome

nooikko:
Closes #1085 

## 📝 Description

> Updated the Textarea component to properly handle resizing when controlled by an outside state controller such as `useState` or `useInput`

> The current behavior causes the form to not properly handle resizing, resulting in a scrollbar appearing no matter the props (such as `maxRows`) passed in, when such props would typically cause the Textarea to expand accordingly.

## 🚀 New behavior

> The Textarea now correctly expands when utilizing props such as `maxRows`.

No

## 📝 Additional Information
As a consideration, the general typing of for `value` when using an input allows for `string | string[] | number | undefined`. This PR does cause this to be assumed to be `string` which should be fine as all three of the available types are translated to `string` under the hood.

### DIFF
--- a/packages/react/src/textarea/textarea.tsx
+++ b/packages/react/src/textarea/textarea.tsx
@@ -84,7 +84,7 @@ const Textarea = React.forwardRef<HTMLTextAreaElement, TextareaProps>(
 
       const [height, rowHeight] = calculateNodeHeight(
         nodeSizingData,
-        node.value || node.placeholder || "x",
+        (isControlled && (props.value as string)) || node.value || node.placeholder || "x",
         rows || minRows,
         rows || maxRows,
       );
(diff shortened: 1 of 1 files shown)


jrgarciadev (approved):
Thank you! @nooikko

PR #1169, opened 2023-06-20
fix: fixed bordered button styling issues inside group

jguddas:
#### Focus state
Before: image
After: image
#### Pressed state
Before: image
After: image
#### Sizing
If the content is the same, the buttons are all the same size.
image


jrgarciadev (approved):
Amazing! 🚀

PR #1172, opened 2023-06-21
 feat(tabs): improve tab accessibility

ryo-manba:
## 📝 Description

I've made improvements to the tab component by updating the HTML element from `<div>` to `<button>`. 
This change aligns with the accessibility guidelines outlined in the "Role, Property, State, and Tabindex Attributes" section of the following page: [Link to the ARIA tabs example](https://www.w3.org/WAI/ARIA/apg/patterns/tabs/examples/tabs-automatic/). 
By using the <button> element, we ensure better semantic representation and enhance keyboard accessibility for tab navigation.

The current behavior of the tab component is using a `<div>` element to represent each tab item.

## 🚀 New behavior

With this change, the tab component now uses the more appropriate `<button>` element for each tab item. This improves the accessibility and ensures proper keyboard interaction with the tab navigation.

No

## 📝 Additional Information

Using the getByRole function instead of getByTestId for retrieving the tab items is considered a best practice for testing accessibility.

### DIFF
--- a/packages/components/tabs/__tests__/tabs.test.tsx
+++ b/packages/components/tabs/__tests__/tabs.test.tsx
@@ -84,21 +84,21 @@ describe("Tabs", () => {
   test("should select the correct tab with keyboard navigation", async () => {
     const wrapper = render(
       <Tabs aria-label="Tabs static test">
-        <Tab key="item1" data-testid="item1" title="Item 1">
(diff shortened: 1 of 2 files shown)


jrgarciadev (approved):
Thank you! @ryo-manba 🙏🏻

PR #1174, opened 2023-06-21
fix: fixed pagination problems

jguddas:
This PR:
- removes the padding from the top and bottom (elements should not be padded IMO).
- makes it so the focus ring is not cut off anymore.
- makes it so the focus does not get lost when elements move.


jrgarciadev:
Hey @jguddas if you remove the padding the scale animation will be break since it needs some space to display it, I had to use it because the overflow need to be the scroll one

PR #1190, opened 2023-07-11
fix: progress bar with more than 100% is shown as full and with proper label

jguddas:
## Example with `value: 555`

https://storiesv2.nextui.org/?path=/story/components-progress--with-value-formatting&args=value:555

Now the label reflects the value and the progress bar is not stuck anymore.

The label is also fixed for circular progress by this change.

### Before
image

### After
image


jrgarciadev:
Could you please create a `clampPercentage` function inside `packages > utilities > shared-utilities > src > numbers`?

Something like this:

```ts
export function clampPercentage(percentage: number): number {
    return Math.min(Math.max(percentage, 0), 100);
}
```

Then apply it into the `use-progress`

```tsx
 style={{
     transform: `translateX(-${100 - clampPercentage(percentage)}%)`,
   }}
```

jrgarciadev (approved):
Thank you! @jguddas  🙏🏻

PR #1191, opened 2023-07-11
feat: pagination dots changes to arrows when focus visible

jguddas:
https://github.com/nextui-org/nextui/assets/25524993/249a0029-99ac-43cc-98c2-df0749c2284e


jrgarciadev (approved):
Good catch!, thank you! @jguddas

PR #1194, opened 2023-07-13
fix: pagination focus is now maintained after page switching

jguddas:
How it works now:

https://github.com/nextui-org/nextui/assets/25524993/fd67ae3b-d7eb-47e2-b8c1-1b9f35fc33c1


jrgarciadev (approved):
Nice! Thank you @jguddas

PR #1198, opened 2023-07-15
fix: added aria hide outside to popover

jguddas:
- Hides content outside of popover from screen readers when the popover is open.


jrgarciadev (approved):
Thanks!  @jguddas

PR #1199, opened 2023-07-15
fix: fixed navbar link problems in docs

jguddas:
- fixes the structure so it is now `ul > li > a` rather than just `ul > a`
- adds `aria-current="page"`
- fixes the issue that the page jumps when you click on a link that is inside an iframe

### DIFF
--- a/apps/docs/app/examples/navbar/disabled-blur/page.tsx
+++ b/apps/docs/app/examples/navbar/disabled-blur/page.tsx
@@ -1,6 +1,16 @@
 "use client";
 
-import {Navbar, NavbarBrand, NavbarContent, NavbarItem, Link, Button} from "@nextui-org/react";
+import {
+  Navbar,
+  NavbarBrand,
+  NavbarContent,
+  NavbarItem,
+  Link as NextUiLink,
+  LinkProps,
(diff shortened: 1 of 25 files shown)


apps/docs/app/examples/navbar/disabled-blur/page.tsx
```diff
@@ -1,6 +1,16 @@
...
+  Navbar,
+  NavbarBrand,
+  NavbarContent,
+  NavbarItem,
+  Link as NextUiLink,
+  LinkProps,
+  Button,
+} from "@nextui-org/react";
+
+const Link = (props: LinkProps) => <NextUiLink {...props} onClick={(e) => e.preventDefault()} />;
```
jrgarciadev, line 13: Please implement the `Link` directly from `@nextui-org/react` since preventing the event is only useful for the docs but might be confusing when someone using this implementation

jguddas: Not sure how to do that?

jrgarciadev (approved):

jrgarciadev (approved):
Amazing! thank you @jguddas

PR #1201, opened 2023-07-16
fix(stories): fixed password field story issues

jguddas:


packages/components/input/stories/input.stories.tsx
```diff
@@ -90,7 +90,12 @@ const PasswordTemplate: ComponentStory<typeof Input> = (args) => {
       <Input
         {...args}
         endContent={
-          <button className="focus:outline-none" type="button" onClick={togglePasswordVisibility}>
+          <button
+            aria-label="show password"
+            aria-pressed={isPasswordVisible}
```
jrgarciadev, line 95: This param here is not needed since this is a native button

packages/components/input/stories/input.stories.tsx
```diff
@@ -90,7 +90,12 @@ const PasswordTemplate: ComponentStory<typeof Input> = (args) => {
       <Input
         {...args}
         endContent={
-          <button className="focus:outline-none" type="button" onClick={togglePasswordVisibility}>
+          <button
+            aria-label="show password"
```
jrgarciadev, line 94: Would be nice to change the `aria-label` according to the `isPasswordVisible` state.

```js
aria-label={isPasswordVisible ? "hide password : "show password"}
```

jrgarciadev (approved):
Nice, just a few comments @jguddas

packages/components/input/stories/input.stories.tsx
jguddas: You can try it out, and you will see that it does not get read out if you change a label dynamically like this.

Even if it would work, `aria-pressed` is the right pattern for toggle buttons like this.

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-pressed

packages/components/input/stories/input.stories.tsx
jguddas: Native buttons do not have a toggle state, this would only not be needed if it would for example be a checkbox.

jrgarciadev (approved):

PR #1219, opened 2023-07-23
docs (pagination): use onPress instead of onClick

denishnavadiya123:
Closes # 

## 📝 Description

> In pagination docs, there are some button use `onClick` is deprecated. I change it to `onPress`

## ⛳️ Current & New behavior (updates)

> There is no change in behavior


jrgarciadev:
Hi @denishnavadiya123 thanks for sending this PR, the `onPress` event is only available on NextUI components Button/Card etc, on this example we used the native `button` so it doesn't have the `onPress` event only the `onClick` which is fine because the idea is to give users a custom items example

PR #1222, opened 2023-07-26
fix: resolved cursor issues in pagination

jguddas:
- The cursor does not animate anymore on initial render and non page change prop changes.
- The cursor hover state now looks good with `disableAnimation` set.
- The animated cursor is now transparent to the cursor (`pointer-events: none`).

### DIFF
--- a/.changeset/eighty-knives-heal.md
+++ b/.changeset/eighty-knives-heal.md
@@ -1,0 +1,10 @@
+---
+"@nextui-org/pagination": patch
+"@nextui-org/theme": patch
+---
+
+fix: resolved cursor issues in pagination
+
+- The cursor does not animate anymore on initial render and non page change prop changes.
+- The cursor hover state now looks good with disableAnimation set.
+- The animated cursor is now transparent to the cursor (pointer-events: none).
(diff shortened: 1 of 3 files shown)


jrgarciadev:
@jguddas could you please upload a video with the current issue?

jguddas:
> @jguddas could you please upload a video showing the issues?

### The cursor hover state now looks good with `disableAnimation` set.
#### After
This makes more sense as the hover state and is also the way it already looks like for `disableAnimation={false}`.

image

#### Before
image

### The animated cursor is now transparent to the cursor (`pointer-events: none`).
When you hover over the active item you now have the pointer just like you would expect and already have with `disableAnimation={true}`.

### The cursor does not animate any more on initial render and non page change prop changes.
#### After
https://github.com/nextui-org/nextui/assets/25524993/2edf3747-a853-4d7c-8837-76da9a9d7bc5
#### Before
Here you see when we load the table with pagination the purple cursor moves from the left to its initial position.

https://github.com/nextui-org/nextui/assets/25524993/3a6a99ab-8c78-4bd7-a7cf-43eaa64c04d6

jrgarciadev:
Hey @jguddas why did you remove the initial animation? does it have issues? If so please send post one

jrgarciadev (approved):

PR #1278, opened 2023-08-05
fix(tooltip): behavior of tooltip trigger with delay

tianenpang:
Closes https://github.com/nextui-org/nextui/issues/1168

## 📝 Description

The tooltip should not open again after the delay is not reached and the cursor moves out.

Fix the trigger behavior to prevent it from opening again after the delay is not reached and the cursor moves out.

## 🚀 New behavior

Use `triggerProps` instead of custom on-pointer events, it already includes the correct pointer behavior.

No.

#### Before

[before.webm](https://github.com/nextui-org/nextui/assets/32772271/2e1d08bc-d34c-4c85-8b9b-0341e9cd8e1b)

#### After

[after.webm](https://github.com/nextui-org/nextui/assets/32772271/f9d83313-9b6a-4098-9545-fc0ac650f289)

### DIFF
--- a/.changeset/olive-hounds-bake.md
+++ b/.changeset/olive-hounds-bake.md
@@ -1,0 +1,5 @@
+---
+"@nextui-org/tooltip": patch
+---
+
+Fix the trigger behavior to prevent it from opening again after the delay is not reached and the cursor moves out.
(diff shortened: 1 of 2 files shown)


jrgarciadev (approved):
Genius! @tianenpang 🎉

PR #1303, opened 2023-08-06
fix: show focus ring when table row is focused in read mode

jguddas:
A table is still focusable in read mode and should still show a focus ring.

Steps to reproduce:

1. Open https://storybook.nextui.org/iframe.html?args=&id=components-table--static&viewMode=story
2. Press <kbd>tab</kbd>
3. Press <kbd>left</kbd><kbd>left</kbd><kbd>left</kbd><kbd>left</kbd>

Fixed:

https://github.com/nextui-org/nextui/assets/25524993/dd3b1b24-4cf1-4719-9dd7-bb24ff49ee42

### DIFF
--- a/.changeset/witty-trees-buy.md
+++ b/.changeset/witty-trees-buy.md
@@ -1,0 +1,5 @@
+---
+"@nextui-org/table": patch
+---
+
+fix: show focus ring when table row is focused in read mode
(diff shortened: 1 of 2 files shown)


jrgarciadev (approved):
Huge thanks @jguddas Sorry the delay

PR #1313, opened 2023-08-08
Feat/select component

### LINKED ISSUE #1299: [BUG] - Weird keyboard behavior when combining Input with Downshift
### Describe the bug

Hello! First off, lovely component library. Been using the v2 in a small app and have been loving it! 

In that app, I'm also using `downshift` for an Autocomplete component. However, I've noticed a weird behavior after I switched from the old `<Downshift>` component to the `useCombobox` hook. When navigating through the results with the keyboard, every 2nd entry is being skipped. After trying to debug the issue, I noticed that this behavior is related to NextUI's `Input` component, which is also the reason I'm creating the issue here. When switching to a standard `<input>`, the issue disappears. My first naive guess is that the Input component somehow triggers the keyboard event twice.

I tried it in 3 different browsers (Firefox, Safari, Brave) and the issue was apparent in all of them.

If you think I should create the issue at https://github.com/downshift-js instead of here, just tell me and I will do that :)

- https://stackblitz.com/github/David3141/nextui-downshift-example?file=src%2Fcomponents%2FDownshiftExample.tsx
- https://github.com/David3141/nextui-downshift-example

Click into the input field and navigate through the items with your keyboard

### Expected behavior

I expect the keyboard navigation to not skip every other entry. 

### Screenshots or Videos

https://github.com/nextui-org/nextui/assets/13077061/98cea97a-1ed3-49df-a5ef-348a1fa73f35

### Operating System Version

macOS

### Browser

Firefox

jrgarciadev:
Closes #1299 

## 📝 Description

New components

- [x] Select 
- [x] Listbox
- [x] ScrollShadow  

## 🚀 New behavior

3 new components


iwillwen:
Hey @jrgarciadev thanks for your great job! I wanna know when will the beta version for select component be released?

PR #1323, opened 2023-08-08
fix(input): duplicate triggering of input events

tianenpang:
Closes https://github.com/nextui-org/nextui/issues/1308

## 📝 Description

This PR fixes the issue where native DOM events for the input were being triggered twice. This issue was caused by the duplication of `inputProps` and `otherProps` in the `mergeProps`.

Firstly, `inputProps` is derived from `originalProps` after being processed by `useTextField`, which already includes the native DOM events. Secondly, `otherProps` also contain a similar set of props.

In this scenario, there could be two identical DOM event functions present in `mergeProps`. However, within `mergeProps`, the identical functions are executed in a chain ([source](https://github.com/adobe/react-spectrum/blob/main/packages/%40react-aria/utils/src/mergeProps.ts#L56)), leading to the execution of the functions twice.

Filter out duplicate event functions in `filterDOMProps`.

## 🚀 New behavior

The `filterDOMProps` utility method now supports the `omitEventNames` option to prevent redundant event functions. (Skips only functions with names starting with `on[A-Z]`).

No.

## 📝 Additional Information

Test cases have been added and passed successfully.

### DIFF
--- a/.changeset/mean-parrots-whisper.md
+++ b/.changeset/mean-parrots-whisper.md
@@ -1,0 +1,5 @@
+---
+"@nextui-org/react-rsc-utils": patch
+---
+
+Support for omit-event-names option.
--- a/.changeset/new-zoos-hug.md
+++ b/.changeset/new-zoos-hug.md
@@ -1,0 +1,5 @@
+---
+"@nextui-org/input": patch
+---
+
+Fixed event functions call twice.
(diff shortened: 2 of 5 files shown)


jrgarciadev:
Huge thanks! @tianenpang

tianenpang:
Hi @jrgarciadev, this PR only updates the `filterDOMProps` for the input, perhaps there are other components that we should also address to prevent similar issues 👀

PR #1333, opened 2023-08-09
fix: moved sr only modal close buttons to not show unnecessary scroll bar

jguddas:
This pr removes the 2px overflow from the modal in mobile view and consequently removes the scroll bar.

This was the result of visually hidden elements being 1x1px in size.


jrgarciadev:
Good catch! @jguddas could you please create the changeset as well?

jrgarciadev (approved):

jrgarciadev (approved):

Issue #1295, opened 2023-08-06
[Feature Request] Add endContent to AccorionItem


7prog:
### Is your feature request related to a problem? Please describe.

Add an end content in addition to start content to the accordion item for buttons and actions

### Describe the solution you'd like

both start and end content for the accordion item

### Describe alternatives you've considered

I've used both the title and start content to add actions to the end of the trigger

Jlassiezzeddine:
I think this is useful to have both startContent and endContent on the accordion component,

I'm working on a PR

sescotti:
hi @Jlassiezzeddine, do you remember what happened with this in the end? saw a bunch of PRs that failed some checks and no further updates. did you find a workaround?

criskell:
What is the state of this issue?

PR #1337, opened 2023-08-10
feat: endContent prop added to AccordionItem

Jlassiezzeddine:
## 📝 Description

As requested in #1295 (feature request), endContent prop is added to the AccordionItem component

We can just pass startContent prop which usually renders an icon or avatar

## 🚀 New behavior

We can now pass endContent prop which can render an icon, an avatar or a button (that's what's generally needed)

No

## 📝 Additional Information

I added the Button component to the package as dev dependecy, updated the theme and created the stories for Accordion "With End Content" and "With Start And End Content"


packages/components/dropdown/stories/dropdown.stories.tsx
Jlassiezzeddine, line 1: This was committed unintentionally as I was experimenting with another issue.

Sorry it's my first contribution ever

PR #1340, opened 2023-08-10
feat: endContent prop added to AccordionItem component

Jlassiezzeddine:
📝 Description
As requested in #1295 (feature request), endContent prop is added to the AccordionItem component

⛳️ Current behavior (updates)
We can just pass startContent prop which usually renders an icon or avatar

🚀 New behavior
We can now pass endContent prop which can render an icon, an avatar or a button (that's what's generally needed)

💣 Is this a breaking change (Yes/No):
No

📝 Additional Information
I added the Button component to the package as dev dependecy, updated the theme and created the stories for Accordion "With End Content" and "With Start And End Content"


jguddas (changes requested):
I got some bad news, interactive elements can't be nested, no `button` inside `button`.

Screenshot of storybook accessibility panel showing the following error

https://dequeuniversity.com/rules/axe/4.7/nested-interactive

jguddas:
The way to solve this is IMO a redundant click handler on the h2 and a hidden button for a11y.

```tsx
import { usePress, VisuallyHidden } from "react-aria";

export default function Example({ onPress }) {
  let { pressProps: redundentPressProps } = usePress({ onPress });
  let { pressProps: buttonPressProps } = usePress({ onPress });
  return (
    <h2 {...redundentPressProps}>
      Hello World!
      <VisuallyHidden>
        <button {...buttonPressProps}>toggle</button>
      </VisuallyHidden>
    </h2>
  );
}
```

Jlassiezzeddine:
@jguddas  I can't find out how to do it the way your mentionned, I think it's impossible to do nesting interactive elements anyway.

I found a way to put the actions outside of the accordion item button (which is the trigger), but keep it inside the H2 (the header) and position it absolutely in the middle of the heading

accordion

It's accessible and working well.

What do you think ?

jguddas:
> What do you think ?

It's super flaky, put a Lorem Ipsum in there and you will see:

1. It looks funky
2. If you have non-interactive elements in there, the click does not get propagated.

I can take a look on the weekend and see if I can get my suggestion pushed.

packages/core/theme/src/components/accordion.ts
```diff
@@ -62,13 +62,14 @@ const accordionItem = tv({
       "group-[.is-splitted]:shadow-medium",
       "group-[.is-splitted]:rounded-medium",
     ],
-    heading: "",
+    heading: "relative",
     trigger: [
-      "flex py-4 w-full h-full gap-3 outline-none items-center tap-highlight-transparent",
+      "flex py-4 w-full h-full gap-3 outline-none items-center tap-highlight-transparent ",
```
mauriciabad, line 67:
```suggestion
      "flex py-4 w-full h-full gap-3 outline-none items-center tap-highlight-transparent",
```

Unnecessary white space.

Issue #1356, opened 2023-08-11
[BUG] - Images not loading correctly in Image component stories


Jlassiezzeddine:
### Describe the bug

ReferenceError: require is not defined in the Image component stories

### Steps to Reproduce the Bug or Issue

1. Fork the project
2. Follow development setup here : [https://github.com/nextui-org/nextui/blob/main/CONTRIBUTING.MD#development-setup](url)
3. run : pnpm run dev

### Expected behavior

Images should load without error

### Screenshots or Videos

![storybook](https://github.com/nextui-org/nextui/assets/46867906/dacf102f-f6e3-416b-abad-aad00fde8d11)

### Operating System Version

Windows 11

### Browser

Chrome

PR #1357, opened 2023-08-11
fix: storybook image stories ReferenceError: require is not defined

Jlassiezzeddine:
## 📝 Description

Issue : #1356

When we run storybook in dev static files are not loaded correctly using require("path/to/file")
I think the possible solution that I applied in this PR is to add the assets folder to .storybook/main.js :

`module.exports = {
...,
staticDirs: ["../public", "../../components/image/stories/assets"], 
...
}`

[ReferenceError: require is not defined]

storybook

## 🚀 New behavior

Images now load correctly

fix-storybook

No

## 📝 Additional Information

I'm not sure this is the right solution to include every static folder from each component that need images to the staticDirs array

### DIFF
--- a/packages/components/image/stories/image.stories.tsx
+++ b/packages/components/image/stories/image.stories.tsx
@@ -47,7 +47,7 @@ export default {
 
 const defaultProps = {
   ...image.defaultVariants,
-  src: require("./assets/local-image-1.jpeg"),
+  src: "local-image-1.jpeg",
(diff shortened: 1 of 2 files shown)


jrgarciadev:
Hey @Jlassiezzeddine sorry I didn't see you sent this PR, I already fixed here https://github.com/nextui-org/nextui/pull/1359

PR #1371, opened 2023-08-11
docs: On individual instalation instructions at tailwind setup, require package should be @nextui/theme

tuco5:
## 📝 Description

> There is an error in the documentation for installing individual components on the tailwind configuration, i is specified to require @nextui-org/react but in the previous step it is indicated to install "@nextui-org/system" and "@nextui-org/theme"

> Not modifying behavior

## 🚀 New behavior

> No new behavior

> No

## 📝 Additional Information

> This PR is of type docs.


jrgarciadev (approved):
Good catch! @tuco5 thank you

Issue #1395, opened 2023-08-14
[BUG] - `Input` is empty when using `react-hook-form` `defaultValues` or `setValue`


remorses:
### NextUI Version

2.0.19

### Describe the bug

When using `Input` with `react-hook-form` the input will not use `defaultValues`, instead it will always be empty

I think this is related to 

### Your Example Website or App

https://stackblitz.com/edit/github-osw1rc-6utbjj?file=pages%2Findex.js

### Steps to Reproduce the Bug or Issue

1. pnpm dev
2. See the empty input
3. even after a second, when the input should be updated with `setValue`, the input is empty

### Expected behavior

Input should have value 'hello' and 'xxx'

### Operating System Version

Mac

### Browser

Chrome

remorses:
I added a file `with-ref.js` in the reproduction, this bug also happens when doing `ref.current.value = 'value'`, the input is reset after a few milliseconds

PR #1406, opened 2023-08-15
fix(input): don't always use controlled mode

remorses:
closes: #1395 

## 📝 Description

Passing "" to react-aria forces the input to be in controlled mode, where there ref value is ignored

Input does not use ref value

## 🚀 New behavior

Input can be initialized with `ref.current.value = 'xx'`, which is what react-hook-form does

No

### DIFF
--- a/packages/components/input/src/use-input.ts
+++ b/packages/components/input/src/use-input.ts
@@ -125,7 +125,7 @@ export function useInput<T extends HTMLInputElement | HTMLTextAreaElement = HTML
         originalProps?.placeholder,
       ),
       inputElementType: isMultiline ? "textarea" : "input",
-      value: inputValue ?? "",
+      value: inputValue, // don't pass "" if value === undefined, it will switch the input to controlled mode
       onChange: setInputValue,
     },
     domRef,


remorses:
With this PR that issue should be solved too, the problem that caused the warning in the console was these lines:
```
 props.value ?? undefined,
props.defaultValue ?? undefined,
```
If you pass undefined `react-aria` will think that the input is not controlled, if instead you pass `""` it will switch the input to controlled mode, and not print the warning, 

it's the responsibility of the user to pass `""`, not Nextui, always passing `""` will cause much worse bugs like #1395

jrgarciadev:
> With this PR that issue should be solved too, the problem that caused the warning in the console was these lines:
> 
> ```ts
> props.value ?? undefined,
> props.defaultValue ?? undefined,
> ```
> 
> If you pass undefined `react-aria` will think that the input is not controlled, if instead you pass `""` it will switch the input to controlled mode, and not print the warning,
> 
> it's the responsibility of the user to pass `""`, not Nextui, always passing `""` will cause much worse bugs like #1395

The "Warning: A component is changing an uncontrolled input to be controlled. ..." is still there with your changes, I figured out that I was passing the value to the `useTextField` hook but that's not necessary, the solution is to remove the `value: inputValue` prop, could you update the PR please?, and test your use case

remorses:
Ok i removed the value from `useTextField` and my use case still works

I added a test for my use case, it does the check 2 times because previously the value was being reset asynchronously, it would first go to "value" and then ""

jrgarciadev:
Ok thanks!, please add the changeset as well to be able to publish the changes

PR #1408, opened 2023-08-15
fix/add import CardFooter at Card component composition

vab1997:
## 📝 Description

Add import CardFooter at Card component composition in the tab code


jrgarciadev (approved):
Thank you! @vab1997

PR #1412, opened 2023-08-15
fix: import the cardFooter in the card component doc

cinsuasty:
## 📝 Description

Added the missing import statement for CardFooter in the code as it was not present. This fix is necessary to prevent errors for users following the documentation.

Currently, the import statement for CardFooter is missing in the code. This leads to issues for users who are following the provided documentation.

## 🚀 New behavior

The new behavior involves including the necessary import statement for CardFooter in the codebase. This change ensures that users who follow the documentation will no longer encounter errors.

No


jrgarciadev (approved):
Thank you! @cinsuasty

PR #1427, opened 2023-08-18
fix(Button): button link onClick warning

camcam2605:
Closes https://github.com/nextui-org/nextui/issues/1423

## 📝 Description

> Add a brief description

## 🚀 New behavior


jguddas:
The warning is justified regardless of element IMO.

jguddas:
> @jguddas  we should ignore that warning if we don't use `onClick`
> 

If we have a link we should be using `useLink` not `useButton`.

PR #1431, opened 2023-08-18
fix: added value to aria label of select

jguddas:
Before it just read out the label i.e. `Favorite Animal`, now it reads out `Dog, Favorite Animal`.

I added a visually hidden comma so most screen readers put a liddle pause between the value and label.

### DIFF
--- a/packages/components/select/src/select.tsx
+++ b/packages/components/select/src/select.tsx
@@ -4,6 +4,7 @@ import {ScrollShadow} from "@nextui-org/scroll-shadow";
 import {ChevronDownIcon} from "@nextui-org/shared-icons";
 import {forwardRef} from "@nextui-org/system";
 import {FocusScope} from "@react-aria/focus";
+import {VisuallyHidden} from "@react-aria/visually-hidden";
 import {cloneElement, ForwardedRef, ReactElement, Ref, useMemo} from "react";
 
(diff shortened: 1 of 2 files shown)


packages/components/select/src/select.tsx
```diff
@@ -95,7 +96,10 @@ function Select<T extends object>(props: Props<T>, ref: ForwardedRef<HTMLSelectE
             {!shouldLabelBeOutside ? labelContent : null}
             <div {...getInnerWrapperProps()}>
               {startContent}
-              <span {...getValueProps()}>{renderSelectedItem}</span>
+              <span {...getValueProps()}>
+                {renderSelectedItem}
+                <VisuallyHidden>,</VisuallyHidden>
```
jrgarciadev, line 101: @jguddas the selected values are already joined by a "," https://github.com/nextui-org/nextui/pull/1431/files#diff-bdef9d0be5667754a2de1d88a28f30d57743ca21a6fb6f25cf8d1b2681987e98R86

jguddas: This is a visually hidden comma, it changes so the sr will get `Dog, Cat, Favorite Animals`  
instead of `Dog, Cat Favorite Animals`

jrgarciadev (approved):

PR #1436, opened 2023-08-18
fix(docs): replaced ghost by dot in chip variants

loicgreffier:
## 📝 Description

Replaced the unkown `ghost` variant by the `dot` variant for `Chip` component.

N/A

## 🚀 New behavior

N/A

No


jrgarciadev (approved):
Good catch thank you!

PR #1474, opened 2023-08-23
Fix/tooltip arrow position

jrgarciadev:
Closes # 

## 📝 Description

Tooltip arrow position fixed

Tooltip arrow is placed in front of the tooltip content

## 🚀 New behavior

Tooltip arrow is now placed behind the tooltip content


u3u:
image

Hello, this will result in the loss of arrow color because its background color is set to inherit.

PR #1491, opened 2023-08-25
fix(docs): table async pagination

CHENFANGC:
Closes https://github.com/nextui-org/nextui/issues/1466

## 📝 Description

Fixed pagination issue by changing from `useAsyncList` to `useSWR`. While useAsyncList is more suitable for infinite loading lists and could solve the problem of consecutive pagination clicks, it's not the optimal solution for jump-clicking. By using `useSWR`, we can request data only for the current page and cache previously requested data, providing a better user experience for interactions.

Change `useAsyncList` to `useSWR`.

## 🚀 New behavior

Add swr

No


jrgarciadev (approved):
Awesome!, thank you @CHENFANGC

jrgarciadev (changes requested):
@CHENFANGC for some reason when changing from a page to another page the Pagination disappear, could you please check that scenario?  Only happens when the page has not been loaded before.

We could disable the pagination when fetching new data.

CHENFANGC:
@jrgarciadev thank you for the review. IMO, disabling the `Pagination` can lead to a less smooth UX when users click on the next page button consecutively (requiring them to wait for the request to complete).

If disabling is necessary, developers can set the `isDisabled` prop of the `Pagination` manually.

BTW, the preview link mentioned above is not the latest commit deployed.

jrgarciadev:
> @jrgarciadev thank you for the review. IMO, disabling the `Pagination` can lead to a less smooth UX when users click on the next page button consecutively (requiring them to wait for the request to complete).
> 
> If disabling is necessary, developers can set the `isDisabled` prop of the `Pagination` manually.
> 
> BTW, the preview link mentioned above is not the latest commit deployed.


I agree with you @CHENFANGC , but could we just not hide the Pagination when changing the page? every time the pagination unmounts/mount triggers the animation to move the cursor to the selected page and it doesn't look good

CHENFANGC:
> > @jrgarciadev thank you for the review. IMO, disabling the `Pagination` can lead to a less smooth UX when users click on the next page button consecutively (requiring them to wait for the request to complete).
> > If disabling is necessary, developers can set the `isDisabled` prop of the `Pagination` manually.
> > BTW, the preview link mentioned above is not the latest commit deployed.
> 
> I agree with you @CHENFANGC , but could we just not hide the Pagination when changing the page? every time the pagination unmounts/mount triggers the animation to move the cursor to the selected page and it doesn't look good

In the 5672a8f commit, `keepPreviousData: true` was used in `useSWR`, which resolved the issue of hiding pagination. In my case, it doesn't hide pagination. Do you still experience pagination hiding on your end?

https://github.com/nextui-org/nextui/assets/29766018/48634d1d-2fe7-4b11-89ef-c172db8b52eb

jrgarciadev:
Nice thank you! @CHENFANGC

PR #1495, opened 2023-08-26
fix(shared-utils): expose getProp

bdbergeron:
## 📝 Description

Exposes the `getProp` helper function, which unlike `getKeyValue`, allows for retrieving properties from nested objects. This is helpful when working with dynamic `Table` instances.

`getProp` is not exported for consumers to use.

## 🚀 New behavior

`getProp` can be properly imported for consumers to use.

### DIFF
--- a/.changeset/chilly-taxis-matter.md
+++ b/.changeset/chilly-taxis-matter.md
@@ -1,0 +1,5 @@
+---
+"@nextui-org/table": patch
+---
+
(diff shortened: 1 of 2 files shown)


jrgarciadev:
Hey @bdbergeron I appreciate you effort in sending this PR, however, this function is meant to be used internally, if we expose it we will need to be careful when changing it, users have options as the `get` one from [lodash](https://lodash.com/docs/4.17.15#get) that offers better support and is well tested.

Thanks!

PR #1508, opened 2023-08-27
fix(card-storybook): replace require with static image reference

tbor00:
## 📝 Description
This PR is related to:

- fix with the require on card storybook
- add card assets to static dir on storybook config

In the storybook of the card, we use require for set images. I changed that to fix the error and improve static references on storybook config

imagen

imagen

## 🚀 New behavior
Don't have issue's with the images and we can refer to a card assets

imagen

imagen

No


jguddas:
> Why we should move the images?

Other storybooks and all refer to the public directory, and we need to stay consistent.

jrgarciadev:
Fixed

PR #1509, opened 2023-08-27
fix: some typo and renamed use infinite scroll call on async section …

MisterHims:
…of select docpage

## 📝 Description

- Fix some typo in Select component documentation page
- Change "use-infinity-scroll" by the correct import "use-infinite-scroll"


MisterHims:
Done. I intentionally don't let the lockfileVersion bumped to 6.1 (instead of 6.0) in the pnpm-lock.yaml file.

jrgarciadev (approved):
Huge thanks @MisterHims

PR #1513, opened 2023-08-28
feat(system): extend props for nextui provider

tianenpang:
Resolves https://github.com/nextui-org/nextui/issues/1494

## 📝 Description

Since the `OverlayProvider` wraps its children in a `div` container at the top level of the app, this can cause the children to not inherit the layout of the `body` element. 

While it's possible to set the layout for the container using selectors in a css file, but we could improve the DX by extending the props for `NextUIProvider` and passing them down to `OverlayProvider`. This way, we could directly set styles for the container by using `className` or `style` prop.

- Extend `NextUIProviderProps` from `ModalProviderProps` and pass these props down to `overlayProvider`.

## 🚀 New behavior

- Update the default locale to `en-US`.

No.

### DIFF
--- a/.changeset/three-shrimps-hope.md
+++ b/.changeset/three-shrimps-hope.md
@@ -1,0 +1,5 @@
+---
+"@nextui-org/system": patch
+---
+
+Extend props for NextUIProvider and update default locale from en to en-US.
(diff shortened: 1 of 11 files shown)


jrgarciadev (approved):
Thanks mate! @tianenpang

PR #1532, opened 2023-09-01
fix: #1531 title prop of a tab should be a string

### LINKED ISSUE #1531: [BUG] - When the `title` prop of Tab is not a string, it should not be passed to HTML elements.
### NextUI Version

2.1.7

### Describe the bug

It will cause the incorrect title `[object Object]` to be displayed when hovering over it.

image

_No response_

https://nextui.org/docs/components/tabs#with-icons

### Expected behavior

Only when `title` is a string can it be passed to HTML elements.

### Screenshots or Videos

_No response_

### Operating System Version

macOS

### Browser

Chrome

akshayaqburst:
Closes #1531 

## 📝When the title prop of Tab is not a string, it should not be passed to HTML elements

### DIFF
--- a/packages/components/tabs/src/tab.tsx
+++ b/packages/components/tabs/src/tab.tsx
@@ -85,6 +85,10 @@ const Tab = forwardRef<"button", TabItemProps>((props, ref) => {
     });
   };
 
+  if (otherProps && otherProps.title && typeof otherProps.title !== "string") {
+    otherProps.title = "";
+  }
+
   return (
     <Component
       ref={domRef}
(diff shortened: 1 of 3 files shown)


packages/components/tabs/src/tab.tsx
```diff
@@ -85,6 +85,10 @@ const Tab = forwardRef<"button", TabItemProps>((props, ref) => {
     });
   };
 
+  if (otherProps && otherProps.title && typeof otherProps.title !== "string") {
+    otherProps.title = "";
+  }
+
```
jguddas, line 91:
```suggestion
```

jguddas (changes requested):

PR #1533, opened 2023-09-01
fix: title prop of a tab should be a string

### LINKED ISSUE #1531: [BUG] - When the `title` prop of Tab is not a string, it should not be passed to HTML elements.

akshayaqburst:
Closes #1531 

## 📝 [Description](fix: title prop of a tab should be a string)

> Add a brief description

## ⛳️ It will cause the incorrect title [object Object] to be displayed when hovering over it.

## 🚀 New behavior

> Only when title is a string can it be passed to HTML elements.


jguddas (commented):
Please don't.

`title` should not be set to `''` by default and the title prop should probably be inherited from ItemProps by being removed from Omit.

PR #1588, opened 2023-09-09
fix: added input error message and description screen reader support

jguddas:
## 📝 Description

`useTextField` uses `useField` which adds `aria-describedby` only when the `errorMessage` or `description is provided.

`aria-describedby` does not include the error message or description and thus does not get read out by screen readers.

## 🚀 New behavior

It gets read out by screen readers.

No.

## 📝 Additional Information

https://github.com/adobe/react-spectrum/blob/main/packages/%40react-types/shared/src/inputs.d.ts#L77

### DIFF
--- a/.changeset/light-ads-smoke.md
+++ b/.changeset/light-ads-smoke.md
@@ -1,0 +1,5 @@
+---
+"@nextui-org/input": patch
+---
+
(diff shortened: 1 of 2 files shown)


jguddas:
Nvm, it actually works right now.

PR #1591, opened 2023-09-09
fix: resolved overlap issue with arrow in select

jguddas:
Before the content could grow until it was under the arrow.

Now:

image

### DIFF
--- a/packages/core/theme/src/components/select.ts
+++ b/packages/core/theme/src/components/select.ts
@@ -9,7 +9,7 @@ const select = tv({
     label: "block text-small font-medium text-foreground-500 pointer-events-none",
     mainWrapper: "w-full flex flex-col",
     trigger:
-      "relative px-3 gap-3 w-full inline-flex flex-row items-center shadow-sm outline-none tap-highlight-transparent",
+      "relative px-3 gap-3 w-full inline-flex flex-row items-center shadow-sm outline-none tap-highlight-transparent pr-10",
     innerWrapper:
       "inline-flex h-full w-[calc(100%_-_theme(spacing.unit-6))] items-center gap-1.5 box-border",
     selectorIcon: "absolute right-3 w-unit-4 h-unit-4",
--- a/packages/core/theme/src/components/select.ts
+++ b/packages/core/theme/src/components/select.ts
@@ -11,7 +11,7 @@ const select = tv({
     trigger:
       "relative px-3 gap-3 w-full inline-flex flex-row items-center shadow-sm outline-none tap-highlight-transparent pr-10",
     innerWrapper:
-      "inline-flex h-full w-[calc(100%_-_theme(spacing.unit-6))] items-center gap-1.5 box-border",
+      "inline-flex h-full items-center gap-1.5 box-border",
     selectorIcon: "absolute right-3 w-unit-4 h-unit-4",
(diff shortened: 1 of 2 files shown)


packages/core/theme/src/components/select.ts
```diff
@@ -9,7 +9,7 @@ const select = tv({
     label: "block text-small font-medium text-foreground-500 pointer-events-none",
     mainWrapper: "w-full flex flex-col",
     trigger:
-      "relative px-3 gap-3 w-full inline-flex flex-row items-center shadow-sm outline-none tap-highlight-transparent",
+      "relative px-3 gap-3 w-full inline-flex flex-row items-center shadow-sm outline-none tap-highlight-transparent pr-10",
     innerWrapper:
       "inline-flex h-full w-[calc(100%_-_theme(spacing.unit-6))] items-center gap-1.5 box-border",
```
jrgarciadev, line 14: @jguddas this line already handles the overlaps `innerWrapper: "... w-[calc(100%_-_theme(spacing.unit-6))]"`, we could use either the padding-right or the width

jguddas: I will build an example when I get time, I definitely have overlap when I don't add it.

jrgarciadev: Ok maybe we could remove the `width` then, and only use the padding

jguddas: Ideally we would use inline rather than absolute, this would be a bit bigger change tho.

jrgarciadev: What do you mean by inline?

jguddas: Remove display: absolute

jrgarciadev: Oh got it you mean in the `selectorIcon`, It will be more complex because the `innerWrapper` slot can grow with the content you we need to keep the selector icon always in the correct position, Let's go first with padding and removing the `w-[calc(100%_-_theme(spacing.unit-6))]`

jguddas, line 14:
```suggestion
      "inline-flex h-full items-center gap-1.5 box-border",
```

packages/core/theme/src/components/select.ts
```diff
@@ -9,9 +9,9 @@ const select = tv({
     label: "block text-small font-medium text-foreground-500 pointer-events-none",
     mainWrapper: "w-full flex flex-col",
     trigger:
-      "relative px-3 gap-3 w-full inline-flex flex-row items-center shadow-sm outline-none tap-highlight-transparent",
+      "relative px-3 gap-3 w-full inline-flex flex-row items-center shadow-sm outline-none tap-highlight-transparent pr-10",
```
jrgarciadev, line 12: @jguddas let's reduce the pr to `6`, please check this line https://github.com/nextui-org/nextui/pull/1591/files#diff-cf57aaf38eadf5556e28ffe5f5cd170c1b121821a248be7d551d02c13025b0c4L88 

The `px` overrides the `pr-` on `sm` sizes

jrgarciadev (changes requested):

wingkwong:
Closing due to inactivity.

PR #1614, opened 2023-09-12
Update standalone.ts

verheyenkoen:
Skeleton standalone example does not need to import the Card component as it's not being used in this example.

## 📝 Description

Small update/fix to the Skeleton docs.

n/a

## 🚀 New behavior

n/a

No

## 📝 Additional Information

None


jrgarciadev:
Thank you! @verheyenkoen

PR #1643, opened 2023-09-16
fix(docs): button custom implement

li-jia-nan:
## 📝 Description

Button custom implementation example fixed


jrgarciadev (approved):

jrgarciadev:
Good catch @li-jia-nan  thanks!

PR #1644, opened 2023-09-17
docs: fix code Sizes demo overflow

li-jia-nan:
## 📝 Description

docs: fix code Sizes demo overflow

## before:

image

## after:

image

### DIFF
--- a/apps/docs/content/components/code/sizes.ts
+++ b/apps/docs/content/components/code/sizes.ts
@@ -2,11 +2,11 @@ const App = `import {Code} from "@nextui-org/react";
 
 export default function App() {
   return (
-    <div className="flex gap-4">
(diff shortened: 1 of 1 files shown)


apps/docs/content/components/code/sizes.ts
```diff
@@ -2,11 +2,11 @@ const App = `import {Code} from "@nextui-org/react";
 
 export default function App() {
   return (
-    <div className="flex gap-4">
+    <div className="flex flex-col gap-4">
```
jguddas, line 5: This is what we use for `Snippet`
```suggestion
    <div className="flex flex-wrap gap-4">
```

jguddas (approved):

jrgarciadev (approved):
Thank you! @li-jia-nan

PR #1651, opened 2023-09-18
fix: add peer dependencies

syi0808:
## 📝 Description
Screenshot 2023-09-18 at 2 31 19 PM
I encountered an error starting a new Next 13 project. This was a problem with yarn PnP and there was no similar case. As a result of analyzing the error, it was an error that the react-dom package could not be found in the react-aria/overlay package. 

Looking for a little more, the peer dependency of the package used by next-ui was not declared as the peer dependency of next-ui.

The following solutions are being used temporarily.

Change peer dependency forced injection or pnpFallbackMode: "all" using packageExtensions option in yarnrc.yml

It is enough for me already, but I made a PR because I thought it would be nice if a more fundamental solution could be prepared in next-ui.

Module not found because peer dependency does not have react-dom and framer-motion in nextui packages. (only yarn PnP , not node_modules)

## 🚀 New behavior
Add as peer dependency to packages requiring react-dom or framer-motion.

No

### DIFF
--- a/packages/components/accordion/package.json
+++ b/packages/components/accordion/package.json
@@ -41,6 +41,7 @@
   },
   "peerDependencies": {
     "react": ">=18",
+    "react-dom": ">=18",
     "framer-motion": ">=4.0.0"
   },
   "dependencies": {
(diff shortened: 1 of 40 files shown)


jrgarciadev (approved):
Great work, thank you! @syi0808

syi0808:
I didn't update Yarnlock. Sorry. Update within 30 minutes

PR #1674, opened 2023-09-22
fix(select): setting padding top 0 if label and placeholder are both falsy

ChiuMungZitAlexander:
## 📝 Description

Adjust the padding of innerWrapper of select when `label` is undefined/falsy.

Due to we set `innerWrapper: "pt-4",` if `label` and `placeholder` are both falsy, there will still have a padding top.

```
{
  isLabelPlaceholder: true,
  labelPlacement: "inside",
  size: "sm",
  class: {
    label: ["group-data-[filled=true]:-translate-y-[calc(50%_+_theme(fontSize.tiny)/2_-_3px)]"],
    innerWrapper: "pt-4",
  },
},
{
  isLabelPlaceholder: true,
  labelPlacement: "inside",
  size: "md",
  class: {
    label: [
      "group-data-[filled=true]:-translate-y-[calc(50%_+_theme(fontSize.small)/2_-_4px)]",
    ],
    innerWrapper: "pt-4",
  },
},
```

⬇️ In this case, both `placeholder` and `label` are not provided.

```
<Select classNames={{ label: 'hidden', popover: 'bg-white' }}>
 {animals.map(animal => (
    <SelectItem
      key={animal.value}
      value={animal.value}
    >
      {animal.label}
    </SelectItem>
  ))}
</Select>
```
image

## 🚀 New behavior

Remove the padding top of innerWrapper if `label` and `placeholder` are both falsy, as neither of them is required.

image

No

## 📝 Additional Information

No

### DIFF
--- a/packages/components/select/src/use-select.ts
+++ b/packages/components/select/src/use-select.ts
@@ -463,7 +463,7 @@ export function useSelect<T extends object>(originalProps: UseSelectProps<T>) {
       return {
         ...props,
         className: slots.innerWrapper({
-          class: clsx(classNames?.innerWrapper, props?.className),
+          class: clsx(classNames?.innerWrapper, !label && "pt-0", props?.className),
         }),
       };
     },
--- a/packages/components/select/stories/select.stories.tsx
+++ b/packages/components/select/stories/select.stories.tsx
@@ -601,6 +601,23 @@ const LabelPlacementTemplate = ({color, variant, ...args}: SelectProps) => (
         </Select>
       </div>
     </div>
+    <div className="w-full max-w-2xl flex flex-col gap-3">
+      <h3>With falsy label</h3>
+      <div className="w-full flex flex-row items-end gap-4">
+        <Select
+          className="w-1/3"
+          color={color}
+          placeholder="Select an animal"
+          variant={variant}
+          {...args}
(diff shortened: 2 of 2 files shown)


packages/components/select/src/use-select.ts
```diff
@@ -463,7 +463,7 @@ export function useSelect<T extends object>(originalProps: UseSelectProps<T>) {
       return {
         ...props,
         className: slots.innerWrapper({
-          class: clsx(classNames?.innerWrapper, props?.className),
+          class: clsx(classNames?.innerWrapper, !label && "pt-0", props?.className),
```
jrgarciadev, line 466: Hey @ChiuMungZitAlexander we cannot add styles inside the components logic/hooks because tailwindcss will not take them

All the styles should be added into the `theme/components` packages

jrgarciadev: You could use a data property `data-has-label` check the use-select code

jguddas:
We have labels marked as optional even tough that violate [WCAG 3.3.2](https://www.w3.org/TR/WCAG21/#labels-or-instructions), we should throw atleast a type if not a proper error when `label`, `aria-label` and `aria-labledby` is not set.

PR #1677, opened 2023-09-22
button: add letter-spacing style for button children

li-jia-nan:
## 📝 Description

When the content of the button is in Chinese and has two characters, we can add a letter-spacing in the characters, be like https://ant.design/components/button#faq

image

no have letter-spacing

## 🚀 New behavior

have letter-spacing

No


jguddas:
Why not create a new component called `SmartZnSpace` and use it inside the button?

Maybe a bit out of scope for this library.

jguddas:
> I agree with you, I can definitely implement it externally, but it would be great if this feature could be added to the UI library

Can you do a little write-up or post a link to why there should be a spacing between two Chinese characters?

PR #1682, opened 2023-09-23
fix(select): add placeholder dep for renderSelectedItem

### LINKED ISSUE #1678: [BUG] - Select placeholder not updated
### NextUI Version

2.1.13

### Describe the bug

In the Select component, I have `t('xxx')` in label and `t('yyy')` in placeholder props. 

The t function is from useTranslation  (next-i18next)

**The label is able to be changed when the locale change but the placeholder is not able to be changed**

I have the react code, like below

```jsx
<Select
  label={t('xxx')}
  placeholder={t('yyy')}
>
....
</Select>
```

This is probably due to the useMemo in `renderSelectedItem` https://github.com/nextui-org/nextui/blob/main/packages/components/select/src/select.tsx#L75

NA

1. Install next-i18next
2. create a Select component
3. use t function with useTranslation in label prop and placeholder prop

### Expected behavior

As a user, the placeholder should be updated when locale is changed

### Screenshots or Videos

_No response_

### Operating System Version

macOS

### Browser

Chrome

markyong:
Closes #1678

## 📝 Description

> Add a brief description

## 🚀 New behavior

### DIFF
--- a/.changeset/forty-vans-refuse.md
+++ b/.changeset/forty-vans-refuse.md
@@ -1,0 +1,5 @@
+---
+"@nextui-org/select": patch
+---
+
+fix #1678 Select component placeholder prop should be fresh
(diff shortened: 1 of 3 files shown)


jguddas (approved):

jrgarciadev (approved):
Thank you! @markyong

PR #1686, opened 2023-09-24
feat: Slider Component

### LINKED ISSUE #172: New Component: Range-slider
Would love to see something like 1. & animatable like 2.
1. https://chakra-ui.com/docs/form/range-slider#customizing-the-slider
2. https://www.framer.com/components/slider/

### LINKED ISSUE #888: [Feature Request] New Component Slider

NextUI does not have a slider component.

Slider

Slider

### Screenshots or Videos

_No response_

jrgarciadev:
Closes #172 
Closes #886 
Closes #888  

## 📝 Description

New component added: Slider

### Breaking Changes

- Popover API changed to improve the arrow implementation, arrow is now a pseudo-element, which allows the popover to also move the arrow altogether, this change impacts the Popover, Tooltip and Select implementations.

Popover changes:

```diff
  <Popover
      showArrow
      backdrop="opaque"
      placement="right"
      classNames={{
-        base: "py-3 px-4 border border-default-200 bg-gradient-to-br from-white to-default-300 dark:from-default-100 dark:to-default-50",
-        arrow: "bg-default-200",
+        base: [
+          // the "before" pseudo element is now the popover' arrow
+          "before:bg-default-200"
+        ],
+        content: [ // now we need to use the "content" slot to actually modify the popover' content styles
+          "py-3 px-4 border border-default-200",
+          "bg-gradient-to-br from-white to-default-300",
+          "dark:from-default-100 dark:to-default-50",
        ],
      }}
    >
      <PopoverTrigger>
        <Button>Open Popover</Button>
      </PopoverTrigger>
      <PopoverContent>
        {(titleProps) => (
          <div className="px-1 py-2">
            <h3 className="text-small font-bold" {...titleProps}>
              Popover Content
            </h3>
            <div className="text-tiny">This is the popover content</div>
          </div>
        )}
      </PopoverContent>
  </Popover>
```

Tooltip changes:

```diff
 <Tooltip
      showArrow
      placement="right"
      content="I am a tooltip"
      classNames={{
-        base: "py-2 px-4 shadow-xl text-black bg-gradient-to-br from-white to-neutral-400",
-        arrow: "bg-neutral-400 dark:bg-white",
+        base: [
+          // the "before" pseudo element is now the popover' arrow
+          "before:bg-neutral-400 dark:before:bg-white",
+        ],
+        content: [ // now we need to use the "content" slot to actually modify the popover' content styles
+          "py-2 px-4 shadow-xl",
+          "text-black bg-gradient-to-br from-white to-neutral-400",
+        ],
      }}
    >
      <Button variant="flat">Hover me</Button>
</Tooltip>
```

Select changes:

```diff
 <Select
      items={users}
      label="Assigned to"
      className="max-w-xs"
      variant="bordered"
      classNames={{
        label: "group-data-[filled=true]:-translate-y-5",
        trigger: "min-h-unit-16",
        listboxWrapper: "max-h-[400px]",
      }}
      listboxProps={{
        itemClasses: {
          base: [
            "rounded-md",
            "text-default-500",
            "transition-opacity",
            "data-[hover=true]:text-foreground",
            "data-[hover=true]:bg-default-100",
            "dark:data-[hover=true]:bg-default-50",
            "data-[selectable=true]:focus:bg-default-50",
            "data-[pressed=true]:opacity-70",
            "data-[focus-visible=true]:ring-default-500",
          ],
        },
      }}
      popoverProps={{
        classNames: {
-          base: "p-0 border-small border-divider bg-background",
-          arrow: "bg-default-200",
+          base: "before:bg-default-200", // the before pseudo element controls the popover's arrow
+          content: "p-0 border-small border-divider bg-background", // now instead of the "base" slot we use the "content" slot
        },
      }}
      renderValue={(items) => {
        return items.map((item) => (
          <div key={item.key} className="flex items-center gap-2">
            <Avatar
              alt={item.data.name}
              className="flex-shrink-0"
              size="sm"
              src={item.data.avatar}
            />
            <div className="flex flex-col">
              <span>{item.data.name}</span>
              <span className="text-default-500 text-tiny">({item.data.email})</span>
            </div>
          </div>
        ));
      }}
    >
      {(user) => (
        <SelectItem key={user.id} textValue={user.name}>
          <div className="flex gap-2 items-center">
            <Avatar alt={user.name} className="flex-shrink-0" size="sm" src={user.avatar} />
            <div className="flex flex-col">
              <span className="text-small">{user.name}</span>
              <span className="text-tiny text-default-400">{user.email}</span>
            </div>
          </div>
        </SelectItem>
      )}
    </Select>
  );
}`;
```

Dropdown Changes

```diff
    <Dropdown
      showArrow
      classNames={{
-        base: "py-1 px-1 border border-default-200 bg-gradient-to-br from-white to-default-200 dark:from-default-50 dark:to-black",
-        arrow: "bg-default-200",
+        content: "py-1 px-1 border border-default-200 bg-gradient-to-br from-white to-default-200 dark:from-default-50 dark:to-black",  // now instead of the "base" slot we use the "content" 
+        base: "before:bg-default-200",  // the before pseudo element controls the popover's arrow
      }}
    >
```


packages/components/slider/stories/slider.stories.tsx
```diff
@@ -13,7 +13,7 @@ export default {
     },
     color: {
       control: {type: "select"},
-      options: ["default", "primary", "secondary", "success", "warning", "danger"],
+      options: ["foreground", "primary", "secondary", "success", "warning", "danger"],
```
jguddas, line 16: `foreground` might be a better name, but keeping it consistent is IMO more important.

jrgarciadev: We use `foreground` in some components https://nextui.org/docs/components/link#colors

packages/core/theme/src/components/slider.ts
```diff
@@ -110,11 +111,20 @@ const slider = tv({
         thumb: "cursor-default",
       },
     },
+    disableAnimation: {
+      true: {
+        thumb: "data-[dragging=true]:after:scale-100",
+      },
+      false: {
+        thumb: "after:transition-all",
```
jguddas, line 119:
```suggestion
        thumb: "motion-reduce:after:transition-all",
```

packages/components/slider/stories/slider.stories.tsx
jguddas: Okay :+1:

packages/components/slider/src/use-slider.ts
```diff
@@ -0,0 +1,188 @@
...
+          ? numberFormatter.format(state.values[0])
+          : numberFormatter.formatRange(state.values[0], state.values[state.values.length - 1]),
+    };
+  };
+
+  const getTrackProps: PropGetter = (props = {}) => {
+    return {
+      ref: trackRef,
+      className: slots.track({class: classNames?.track}),
+      "data-thumb-count": state.values.length,
```
jguddas, line 143:
```suggestion
      "data-thumb-count": fillOffset !== undefined ? state.values.length : undefined,
```

packages/core/theme/src/components/slider.ts
```diff
@@ -0,0 +1,134 @@
...
+      "rounded-full",
+      "top-1/2",
+      "flex",
+      "justify-center",
+      "items-center",
+      "before:absolute",
+      "before:w-11",
+      "before:h-11",
+      "before:rounded-full",
+      "data-[dragging=true]:after:scale-80",
```
jguddas, line 41: What about something like this?
```suggestion
      "data-[dragging=true]:after:scale-80",
      "data-[focus-visible=true]:after:scale-80",
```

packages/components/slider/src/use-slider.ts
```diff
@@ -0,0 +1,188 @@
...
+          ? numberFormatter.format(state.values[0])
+          : numberFormatter.formatRange(state.values[0], state.values[state.values.length - 1]),
+    };
+  };
+
+  const getTrackProps: PropGetter = (props = {}) => {
+    return {
+      ref: trackRef,
+      className: slots.track({class: classNames?.track}),
+      "data-thumb-count": fillOffset !== undefined ? state.values.length : undefined,
```
jguddas, line 143:
```suggestion
      "data-thumb-count": fillOffset === undefined ? state.values.length : undefined,
```

packages/core/theme/src/components/slider.ts
```diff
@@ -0,0 +1,134 @@
...
+        base: "opacity-disabled",
+        thumb: "cursor-default",
+      },
+    },
+    disableAnimation: {
+      true: {
+        thumb: "data-[dragging=true]:after:scale-100",
+      },
+      false: {
+        thumb: "motion-reduce:after:transition-all motion-reduce:after:transition-none",
```
jguddas, line 119:
```suggestion
        thumb: "after:transition-all motion-reduce:after:transition-none",
```

packages/core/theme/src/components/slider.ts
jrgarciadev: I think when navigating with the keyboard the interaction/animations should be as simple as possible

Dawsoncodes:
When will this be merged?

jrgarciadev:
https://github.com/nextui-org/nextui/pull/1686#issuecomment-1738317340

@Dawsoncodes tests and small refinements are still missing

mauriciabad:
I found a bug. If the slider is set to "range" and you move both handles to the maximum right, then you can't move them anymore.

jguddas:
> I found a bug. If the slider is set to "range" and you move both handles to the maximum right, then you can't drag them anymore. (the only option is to click on the line, but that feature is not intuitive because there's no visual hint that the line is clickable, so many people would think clicking does nothing and get stuck).

@mauriciabad do you have a proposal for how to fix this, I'm out of ideas?

mauriciabad:
@jrgarciadev Sure!

I would do the same as Material UI does:
https://mui.com/material-ui/react-slider/#range-slider

Notice that if you drag a handle over the other one, they "flip", so the minimum becomes the maximum and the maximum becomes the minimum. 

This also adds a nice feature, and makes more sense to me.

Check out also the [Minimum distance](https://mui.com/material-ui/react-slider/#minimum-distance) feature, it's a nice have.

mauriciabad:
Oh, and to code it, I'd handle the mouse events on the line and not on the handle.

Notice, that in Material UI, the handles don't have any drag or click events. The click-and-drag event is on the background line, and the handles position is updated accordingly. Contrary to your implementation, if you put both handles very close, and you click over one of them, it's possible that you "grab" the other one if the point you clicked is closer to the other handle.

THE-SIMPLE-MARK:
When will this be merged?

mauriciabad:
The error I mentioned is still there. I'll try to contribute to fix it.

PR #1688, opened 2023-09-24
refactor(root): styles refactor

jrgarciadev:
Closes #1685 

## 📝 Description

- Spacing units changed to from `px` to `rem` this improves the mobile components' sizes
- Tabs/Tab new prop added `shouldSelectOnPressUp` which is enabled by default `true`, this prop defines whether the tabs selection should occur on press up instead of press down.
- Chip font size changed to `text-tiny` on `sm` size.

### DIFF
--- a/.changeset/fifty-snails-glow.md
+++ b/.changeset/fifty-snails-glow.md
@@ -1,0 +1,11 @@
+---
+"@nextui-org/listbox": patch
+"@nextui-org/tabs": patch
+"@nextui-org/theme": patch
+---
+
+Styles Changes
+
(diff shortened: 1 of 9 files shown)


jguddas (changes requested):
> Spacing units changed to from px to rem this improves the mobile components' sizes

Can you elaborate, this does break font size scaling a bit and makes it behave just like zooming.

jguddas:
Can you raise 2 new PRs for?
- Tabs/Tab new prop added shouldSelectOnPressUp which is enabled by default true, this prop defines whether the tabs selection should occur on press up instead of press down.
- Chip font size changed to text-tiny on sm size.

PR #1753, opened 2023-10-07
Fix/slider tooltip android position

jrgarciadev:
### Base
https://github.com/nextui-org/nextui/pull/1751

## 📝 Description

Always update the Slider tooltip position when the thumb value changes

## 📝 Additional Information

https://github.com/nextui-org/nextui/assets/30373425/9accaaf7-043d-4b31-a949-a099c4bbffbb


packages/components/slider/src/use-slider-thumb.ts
```diff
@@ -101,16 +109,19 @@ export function useSliderThumb(props: UseSliderThumbProps) {
...
 
     const content =
       renderOutput && typeof renderOutput === "function" ? renderOutput(`${value}`) : value;
 
     return {
       ...tooltipProps,
-      content,
-      isOpen: isHovered || isDragging,
-      placement: isVertical ? "right" : "top",
+      placement: tooltipProps?.placement ?? isVertical ? "right" : "top",
```
jguddas, line 121: Missing curly braces?
```suggestion
      placement: tooltipProps?.placement ?? (isVertical ? "right" : "top"),
```

packages/components/slider/src/use-slider-thumb.ts
```diff
@@ -101,16 +109,19 @@ export function useSliderThumb(props: UseSliderThumbProps) {
...
+      content: tooltipProps?.content ?? content,
```
jguddas, line 122: Just move the `content: content,` befor the `...tooltipProps`

packages/components/slider/src/use-slider-thumb.ts
```diff
@@ -101,16 +109,19 @@ export function useSliderThumb(props: UseSliderThumbProps) {
...
+      updatePositionDeps: [isDragging, isHovered, value],
+      isOpen: tooltipProps?.isOpen ?? (isHovered || isDragging),
     } as TooltipProps;
```
jguddas, line 125: Can you remove this?
```suggestion
    } as TooltipProps;
```

packages/components/tooltip/src/use-tooltip.ts
```diff
@@ -170,6 +180,12 @@ export function useTooltip(originalProps: UseTooltipProps) {
     containerPadding,
   });
 
+  useLayoutEffect(() => {
+    if (!updatePositionDeps.length) return;
+    // force update position when deps change
+    updatePosition();
+  }, updatePositionDeps);
```
jguddas, line 187: This looks dangerous, are you sure this isn't going to have some edge cases where we are calling this a bunch of times and are gonna get performance issues?

packages/core/theme/src/components/popover.ts
```diff
@@ -36,20 +36,20 @@ const popover = tv({
       // visibility
       "data-[arrow=true]:before:block",
       // top
-      "data-[placement=top]:before:-bottom-1",
+      "data-[placement=top]:before:-bottom-[calc(theme(spacing.1)_-_1.5px)]",
```
jguddas, line 39: Why the `-1.5px`?

packages/components/slider/src/use-slider-thumb.ts
jrgarciadev: @jguddas curly braces are automatically added/removed by eslint/prettier, in this is not needed because the first validation is not a ternary

packages/components/slider/src/use-slider-thumb.ts
jrgarciadev: I added this way to be more explicit, It took me some time to fix a bug because of the `...tooltipProps`

packages/components/slider/src/use-slider-thumb.ts
jrgarciadev: Nop, we need explicit returns

CleanShot 2023-10-07 at 16 22 58

packages/components/tooltip/src/use-tooltip.ts
jrgarciadev: I don't think so, it all depends on the dependencies it will not affect production implementation because the dependencies are empty by default and for the Slider/Tooltip implementation we only call it whenever the thumb value changes

packages/core/theme/src/components/popover.ts
jrgarciadev: This makes the arrow look better, try removing the 1.5 by 1px

PR #1778, opened 2023-10-16
fix(slider): Place last dragged slider handle over the other handle

mauriciabad:
## 📝 Description

Fixes https://github.com/nextui-org/nextui/pull/1686#issuecomment-1742695941

There is a bug in the slider that if both handles are placed on the right side, then they can't be draged on to another position, leaving the user stuck. There's still a way to move the handles, using the kayboard or clicking on the track, but the user may not know that those inputs work. 

## 🚀 New behavior

Now, the last grabbed handle has the data attribute `data-focused`. Now, the user can't get stuch at an end. 

No

## 📝 Additional Information

With this fix, it's still not possible to "swap" the handles by draging one over the other, wich would be a nice feature, but not important.

>  I came up with the solution looking at [the implementation](https://github.com/adobe/react-spectrum/blob/db43a95681e4e0f8d45d5fc6746ac3c1694bc331/packages/%40react-spectrum/slider/src/SliderThumb.tsx#L55) of the [react-spectrum's slider](https://react-spectrum.adobe.com/react-spectrum/Slider.html), that uses React Aria too.

### DIFF
--- a/packages/components/slider/__tests__/drag.ts
+++ b/packages/components/slider/__tests__/drag.ts
@@ -1,0 +1,109 @@
+// ####################################################
+// ##                COPYED FROM HERE                ##
+// ##                WITH TYPES ADDED                ##
+// ## https://testing-library.com/docs/example-drag/ ##
+// ####################################################
+
+import { fireEvent } from '@testing-library/dom'
+
+// https://stackoverflow.com/a/53946549/1179377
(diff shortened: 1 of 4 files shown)


jrgarciadev (approved):
Thank you! @mauriciabad stunning job!

PR #1800, opened 2023-10-23
fix(button): show only spinner if button isLoading & isIconOnly

qWici:
## 📝 Description

The button with only icon doesn't look very good during loading, in my opinion. That's why I decided to suggest displaying only the spinner when loading is in progress.

before dark
before light

## 🚀 New behavior

after dark
after light

No

### DIFF
--- a/packages/components/button/src/button.tsx
+++ b/packages/components/button/src/button.tsx
@@ -21,7 +21,19 @@ const Button = forwardRef<"button", ButtonProps>((props, ref) => {
     disableRipple,
     getButtonProps,
     getRippleProps,
+    isIconOnly,
   } = useButton({...props, ref});
+
+  if (isIconOnly && isLoading) {
+    return (
+      <Component ref={domRef} className={styles} {...getButtonProps()}>
+        {startContent}
+        {isLoading && spinner}
+        {endContent}
+        {!disableRipple && <Ripple {...getRippleProps()} />}
+      </Component>
+    );
+  }
 
   return (
     <Component ref={domRef} className={styles} {...getButtonProps()}>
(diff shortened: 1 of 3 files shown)


packages/components/button/src/button.tsx
```diff
@@ -21,8 +21,20 @@ const Button = forwardRef<"button", ButtonProps>((props, ref) => {
...
+    return (
+      <Component ref={domRef} className={styles} {...getButtonProps()}>
+        {startContent}
+        {isLoading && spinner}
+        {endContent}
+        {!disableRipple && <Ripple {...getRippleProps()} />}
+      </Component>
+    );
+  }
+
```
jrgarciadev, line 37: Hey @qWici thanks!, instead of repeating the same logic as the return block try doing this:

```suggestion
 - if (isIconOnly && isLoading) {
 -   return (
 -     <Component ref={domRef} className={styles} {...getButtonProps()}>
 -       {startContent}
 -       {isLoading && spinner}
 -       {endContent}
 -       {!disableRipple && <Ripple {...getRippleProps()} />}
 -     </Component>
 -   );
-  }

  return (
    <Component ref={domRef} className={styles} {...getButtonProps()}>
      {startContent}
      {isLoading && spinnerPlacement === "start" && spinner}
      {isLoading && isIconOnly ? null : children}
      {isLoading && spinnerPlacement === "end" && spinner}
      {endContent}
      {!disableRipple && <Ripple {...getRippleProps()} />}
    </Component>
  );

```

qWici: Great, thank you. I corrected this

jrgarciadev (approved):
Thank you! @qWici

PR #1960, opened 2023-11-12
fix(docs): wrong tooltip classNames type

nanwy:
## 📝 Description

In version 2.2.x, the classNames prop for tooltips is not of type Record<"base"｜"arrow", string>, it should be of type Record<"base"｜"content", string>.
image


jrgarciadev (approved):
Thank you! @nanwy

PR #1976, opened 2023-11-14
chore: remove `tsup`, use `tsc` instead

remorses:
I wanted to open a PR to enable `declarationMap` to make cmd+click navigate to the source code instead of the `.d.ts` files but i noticed that [tsup doesn't support it](https://github.com/egoist/tsup/issues/488)

Tsup has a lot of other problems and unnecessary bloat

So i opened a PR to use `tsc` instead of `tsup`

## Description
- Replaced all build scripts with `tsc` 
- Removed `dev` scripts given that in development the packages `"main"` is set to the `index.ts` files for all packages, making it unnecessary
- Removed `build:fast` scripts for the same reason
- Removed `tsup.config.ts`
- Added `src` to npm files
- Added "use client" to all packages `index.ts` files given that tsc doesn't have a banner option

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -63,3 +63,5 @@ packages/**/.turbo
 
 # content layer
 .contentlayer
+
+cjs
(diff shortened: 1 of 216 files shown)


wingkwong:
Closing due to inactivity.

PR #2050, opened 2023-11-26
fix(hooks): incorrect attribute names in clearOverflow

### LINKED ISSUE #2010: [Bug] use-data-scroll-overflow isEnabled prop
### NextUI Version

2.2.9

### Describe the bug

i was exploring your source code and i saw you have given `topBottom` and `leftRight` directions in [clearOverflow](https://github.com/nextui-org/nextui/blob/088af8b6dfd0ee3932c77e9050a4b30e61ff633f/packages/hooks/use-data-scroll-overflow/src/index.ts#L129C28-L129C28) function but they must be `top-bottom` and `left-right`. because you are removing attributes with `removeAttribute`.

thats why `ScrollShadow` does not remove shadows when scroll is overflowed and u pass `isEnabled={false}`

1. Go to [clearOverflow](https://github.com/nextui-org/nextui/blob/088af8b6dfd0ee3932c77e9050a4b30e61ff633f/packages/hooks/use-data-scroll-overflow/src/index.ts#L129C28-L129C28)
2. change `topBottom` and `leftRight` to `top-bottom` and `left-right`

### Operating System Version

windows 11

### Browser

Chrome

wingkwong:
Closes #2010

## 📝 Description

change `topBottom` and `leftRight` to `top-bottom` and `left-right`

Using `data-top-bottom-scroll="true"` as an example. 

Before triggering `clearOverflow`, we can see `data-top-bottom-scroll` which is expected.

image

After triggering `clearOverflow`, `data-top-bottom-scroll` is still here, which is not expected.

image

## 🚀 New behavior

Before triggering `clearOverflow`, we can see `data-top-bottom-scroll` which is expected.

image

After triggering `clearOverflow`, with the fix in this PR, `data-top-bottom-scroll="true"` is removed as expected.

image

No

### DIFF
--- a/.changeset/fuzzy-panthers-hear.md
+++ b/.changeset/fuzzy-panthers-hear.md
@@ -1,0 +1,5 @@
+---
+"@nextui-org/use-data-scroll-overflow": patch
+---
+
+fixed incorrect attribute names in clearOverflow (#2010)
--- a/packages/hooks/use-data-scroll-overflow/src/index.ts
+++ b/packages/hooks/use-data-scroll-overflow/src/index.ts
@@ -126,7 +126,7 @@ export function useDataScrollOverflow(props: UseDataScrollOverflowProps = {}) {
     };
 
     const clearOverflow = () => {
-      ["top", "bottom", "topBottom", "left", "right", "leftRight"].forEach((attr) => {
+      ["top", "bottom", "top-bottom", "left", "right", "left-right"].forEach((attr) => {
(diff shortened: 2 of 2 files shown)


jrgarciadev:
Hey @wingkwong could you upload some evidence (Video, Screenshot)?

wingkwong:
@jrgarciadev Just added some screenshots and description in Current behavior & New behavior section.

PR #2084, opened 2023-12-02
doc: mention the conflict between `endContent` and `isCLearable`

zrll12:
## 📝 Description

`endContent` overrides the clear button in `isClearable`

No


wingkwong:
next minor release would support both at the same time.

Issue #2098, opened 2023-12-05
[DOCS] -  Wrong import of Button component in nextjs docs


harshmangalam:
### NextUI Version

2.2.9

### Describe the bug

Imported Button component from `@nextui-org/button` instead of `@nextui-org/react` in nextjs installation docs

### Steps to Reproduce the Bug or Issue

https://nextui.org/docs/frameworks/nextjs

### Expected behavior

```
import {Button} from '@nextui-org/react'
```

### Operating System Version

macOS

### Browser

Chrome

alexbmills5:
![image](https://github.com/nextui-org/nextui/assets/128627893/6ec4be3c-b6bd-4a14-b2c6-a04ccb5c6605)

PR #2099, opened 2023-12-05
fix(docs):  import Button component in nextjs installation docs

harshmangalam:
#2098


jrgarciadev:
"Hi @harshmangalam, the button for the App directory import is correct. It needs to be imported from @nextui-org/button to avoid the need for the use client; directive at the top of the file, as it is already included internally.

PR #2156, opened 2023-12-21
docs(modal): fix radius property

robinsmith-source:
## 📝 Description

> This fix corrects the property type of the custom-styled modal. The radius `2xl`, which is not even listed in the API documentation for the modal, is replaced with `lg`.

> The behavior should be changed since `2xl` is not even implemented. I believe the modal in the example is displayed with a radius of `lg`.

## 🚀 New behavior

> Modal code snipped fits to displayed modal in the example.

No


jrgarciadev (approved):
Thanks! @robinsmith-source

PR #2183, opened 2023-12-30
docs(tooltip): fix delay property

huanfe1:
## 📝 Description

delay and closeDelay seem to be written backwards

https://github.com/nextui-org/nextui/blob/921746f284668a303997d1f3b2fd4b825a2bb360/packages/components/tooltip/src/use-tooltip.ts#L38-L47

change of position

## 🚀 New behavior

no


huanfe1:
fixed #2338

PR #2185, opened 2023-12-31
docs(autocomplete): fix missing defaultItems property

robinsmith-source:
## 📝 Description

> This adds the missing `defaultItems` property to the Controlled Autocomplete for the typescript snippet.

> The example doens't work correctly when copy & pasted. --> Could mislead the user.

## 🚀 New behavior

> The example works now.

No.


jrgarciadev (approved):
Thank you! @robinsmith-source

PR #2186, opened 2023-12-31
feat(autocomplete): automatically focus first non-disabled item

juliesaia:
## 📝 Description

Automatically focus the first selectable item in `<Autocomplete />`

Autocomplete doesn't really work as an autocomplete because you have to hover/click/press an arrow key to focus and element.

## 🚀 New behavior

This PR makes the first non-disabled item automatically focus any time the item list changes.

Not really, honestly feels like more of a bug fix but I still called it a feature

## 📝 Additional Information
1ed594b4-13cd-4dc5-95d9-7597a32f39a8

## Summary by CodeRabbit

- **New Features**
	- Enhanced autocomplete functionality to automatically focus on the first non-disabled item when rendered.
  
- **Bug Fixes**
	- Improved handling of key events to maintain expected behavior during user interactions.


jsonMartin:
Nice, I'd really like this feature implemented! I wrote my own logic to handle this in an app I'm working on, but it's cumbersome and it would be really great to have this working out of the box. 

Any updates on getting this merged?

gumgumalan:
How do I implement this locally? Thanks.

jsonMartin:
> How do I implement this locally? Thanks.

Hi @gumgumalan , since this feature isn't baked in, I personally created a hook to watch for every time the Autocomplete is opened and then manually invoke `focus` on the DOM element to achieve this. Here's an example:

```typescript
// AutoFocus on the NextUI search input when the modal opens
useEffect(() => {
  if (!isOpen) return

  setTimeout(() => {
    const element: HTMLElement | null = document.querySelector(*ADD DOM REF HERE*)
    if (!element) console.warn('Not available for autofocus')
    else element?.focus()
  }, 10)
}, [isOpen, label])
```

daniel-esserweb:
I think this could be and awesome feature

wingkwong (approved):

wingkwong (approved):

PR #2237, opened 2024-01-12
fix: dup // fix on a link for edit this page on github

Kinfe123:
## 📝 Description
 
Screenshot from 2024-01-12 11-12-08
there was duplicate url // on the route caused from ```CONTENT_PATH ``` 
> 

## 🚀 New behavior

>  I just fixed the // to have a / that comes from the ```CONTENT_PATH``` instead of adding a new one

<! It does not break any changes at all

## 📝 Additional Information

there are a couple of ways to implement this kind of approach .. lemme know what you think


jrgarciadev (approved):
Thank you! @Kinfe123

PR #2253, opened 2024-01-16
docs: improve router docs to have Tanstack Router instructions

FelipeEmos:
## 📝 Description

Adds documentation on how to setup NextUI with [Tanstack Router](https://tanstack.com/), the awesome router lib that got stable recently.

It's from Tanstack ( same creators of React Query ) and it has typesafety, good serialization, good integration with React Query and other cool stuff, so a lot of people are wanting to make the switch and it would be great to have a section for it in the NextUI lib 🚀 

I have created a repo to test the setup myself in case anyone wants to check it out
https://github.com/FelipeEmos/nextui-tanstack-router

No docs for setting up Tanstack Router

## 🚀 New behavior

Add documentation instructions to setup Tanstack Router

No

### DIFF
--- a/apps/docs/content/docs/guide/routing.mdx
+++ b/apps/docs/content/docs/guide/routing.mdx
@@ -164,6 +164,81 @@ component (e.g. `BrowserRouter`) so that it has access to React Router's interna
 element should also be defined inside `NextUIProvider` so that links inside the rendered routes have access
 to the router.
 
+### Tanstack Router
+
+First setup your `router` using one of the many possible configurations described in the [Tanstack Router docs](https://tanstack.com/router/v1/docs/overview).
+
+
+Now let's use the router you've created to setup the NextUIProvider
+
+```jsx {14}
+// main.tsx or main.jsx
+import {NextUIProvider} from '@nextui-org/react';
+import { RouterProvider } from '@tanstack/react-router'
+
+// Import the router you've already configured
+import { router } from "./{your router path}"
+
+function navigate(path: string) {
+  void router.navigate({ to: path });
+}
+
+ReactDOM.createRoot(document.getElementById('root')).render(
+  <React.StrictMode>
+    <NextUIProvider navigate={navigate}>
+      <RouterProvider router={router}>
+    </NextUIProvider>
+  </React.StrictMode>,
+)
+```
+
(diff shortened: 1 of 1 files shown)


KoJem9Ka:
The unsolved part of the puzzle remains [preloading](https://tanstack.com/router/latest/docs/framework/react/guide/navigation#link-preloading). I couldn't find a better solution and tried to achieve it by creating an original link via closure:
```typescript
import { type AnyRoute, Link, type LinkProps, type RegisteredRouter, type RoutePaths } from '@tanstack/react-router';
import { forwardRef } from 'react';

/**
 * A function for passing TanstackRouter Link together with type-safe props to the "as" props of another component.
 * @example Problem:
 * <Button as={Link} to="/123">Home</Button>
 *                       ^ type is "any" and without checking
 * @example Solution:
 * <Button as={link({ to: '/home' })}>Home</Button>
 *                         ^ strict type for TanstackRouter path and other params
 */
export const link = <
  TRouteTree extends AnyRoute = RegisteredRouter['routeTree'],
  TFrom extends RoutePaths<TRouteTree> | string = string,
  TTo extends string = string,
  TMaskFrom extends RoutePaths<TRouteTree> | string = TFrom,
  TMaskTo extends string = '',
>(props0: LinkProps<TRouteTree, TFrom, TTo, TMaskFrom, TMaskTo>) => {
  return forwardRef<typeof Link, typeof props0>((props1, ref) => {
    return <Link ref={ref} {...props1} {...props0 as any} />;
  });
};
```
This worked at some point, but after updates to NextUI and Tanstack Router versions, problems appeared when using this trick, for example inside Dropdown, i.e. in DropdownItem.
```tsx
<DropdownItem as={link({ to: '/page' })}>Page</DropdownItem>
```
Although preload is triggered, clicking on such link completely reloads the page to navigate to the next route as if it were a regular "a" with "href".

Initially the problem was that outputting the component's prop type in the "as" was not enough, due to the complex generics of the Link component, so I tried to solve this by creating a wrapper-closure by piping the original generic.

Now, in latest versions: NextUI 2.3.6 & Tanstack Router 1.29.2, I can't even see the Link props hints after specifying it in the "as" prop of NextUI components.

wingkwong:
Closing - no longer applicable since it has been added (in canary branch). You may supplement the content in a new PR if necessary.

PR #2274, opened 2024-01-21
fix(components): number input label #2268 fixed

### LINKED ISSUE #2268: [BUG] - Number input label
### NextUI Version

2.2.9

### Describe the bug

When using a number input and with a 0 for the initial value, the label (default or labelPlacement='inside') does not animate to the correct position. Even when the user is setting the value to 0, the label does not alter its state unless a number other than 0 is inputted. 

_No response_

1. Create a number input
2. Add a text label with the placement in default or inside
3. Set the value to 0 or type in 0

### Expected behavior

As a user, the expected behavior is so that if the user does input a 0, or the set value is 0 such as when the application has saved a previous input and that input is 0, the label's state is in the correct position and not in the default position. 

### Screenshots or Videos

What it looks like:
bug1

What it should look like (replace 1 with 0)
bug2

### Operating System Version

Windows 10

### Browser

Edge

Prakash7895:
Closes #2268 

## 📝 Description

> When using a number input and with a 0 for the initial value, the label (default or labelPlacement='inside') does not animate to the correct position. Even when the user is setting the value to 0, the label does not alter its state unless a number other than 0 is inputted.

> When `defaultValue` or `value` prop set as `0`, label do not move to correct position.

## 🚀 New behavior

> When `defaultValue` or `value` prop set as `0`, label moves to correct position.

No


djpfs (approved):

packages/components/input/src/use-input.ts
```diff
@@ -287,7 +287,9 @@ export function useInput<T extends HTMLInputElement | HTMLTextAreaElement = HTML
         "data-filled-within": dataAttr(isFilledWithin),
         "data-has-start-content": dataAttr(hasStartContent),
         "data-has-end-content": dataAttr(!!endContent),
-        className: slots.input({class: clsx(classNames?.input, !!inputValue ? "is-filled" : "")}),
+        className: slots.input({
+          class: clsx(classNames?.input, !isEmpty(inputValue) ? "is-filled" : ""),
```
jrgarciadev, line 291:
```suggestion
          class: clsx(classNames?.input, isFilled ? "is-filled" : ""),
```

packages/components/input/src/use-input.ts
```diff
@@ -330,7 +332,7 @@ export function useInput<T extends HTMLInputElement | HTMLTextAreaElement = HTML
         "data-focus-visible": dataAttr(isFocusVisible),
         "data-focus": dataAttr(isFocused),
         className: slots.inputWrapper({
-          class: clsx(classNames?.inputWrapper, !!inputValue ? "is-filled" : ""),
+          class: clsx(classNames?.inputWrapper, !isEmpty(inputValue) ? "is-filled" : ""),
```
jrgarciadev, line 335:
```suggestion
         class: clsx(classNames?.input, isFilled ? "is-filled" : ""),
```

Prakash7895: "classNames?.inputWrapper" should not be "classNames?.input".

wingkwong (approved):

jrgarciadev:
@Prakash7895 could you please fix the conflicts?

PR #2325, opened 2024-02-08
fix(spinner): pass classNames.label to slot

ali-idrizi:
## 📝 Description

Fixes the issue with `classNames.label` not being passed to spinner's label slot

The `<span>` rendered when `label` is defined cannot be styled using `classNames.label`

## 🚀 New behavior

The classes defined in `classNames.label` are passed down to the slot, which allows styling the element

No


jrgarciadev (approved):
Thank you! @ali-idrizi 🙏🏻

PR #2369, opened 2024-02-17
Added note for installing React.

awhitford:
## 📝 Description

If you are trying to add NextUI to a brand new Astro project, you might forget to install React first.
As a result, this PR adds that step -- with a link to the Astro documentation, similar to the Tailwind advice.

The current documentation assumes the Astro project already has React installed, but a fresh Astro project will not.

## 🚀 New behavior

This revised documentation is simply more explicit about integrating React in Astro before installing NextUI.

No.


jrgarciadev (approved):
Thank you! @awhitford

PR #2390, opened 2024-02-20
feat: add `updatePositionDeps` prop to popover component

u3u:
## 📝 Description

Currently, the Popover component cannot actively update its position, which makes it unusable in some scenarios, such as when the scroll bar position changes in a nested scrolling container. 

## 🚀 New behavior

Like the Tooltip component, an `updatePositionDeps` property has been added to allow external dependencies to force update the position of Popover.

No


jrgarciadev (approved):
Thank you! @u3u

PR #2437, opened 2024-03-03
fix(utils): include enterKeyHint in dom-props

### LINKED ISSUE #2432: [BUG] - `enterKeyHint` does not get passed through `input` element when using `Input` component
### NextUI Version

2.2.9

### Describe the bug

[`enterKeyHint`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/enterKeyHint) doesn't get passed through to the `input` HTML element which means the keyboard (tested only on iOS) doesn't show the correct button (e.g. `enterKeyHint="search"` should show a blue "Search" button in iOS, but instead shows the default grey "return" button).

_No response_

1. Create an `<Input />` with `enterKeyHint` prop set to one of the available values (e.g. "search", "go", etc.)
2. Render the component and check the attributes on the `input` HTML element

### Expected behavior

As I user, I expected to see the `enterKeyHint` attribute to be set on an `input` element created via an `Input` component with `enterKeyHint` prop set.

### Screenshots or Videos

_No response_

### Operating System Version

macOS

### Browser

Chrome

wingkwong:
Closes #2432

## 📝 Description

- currently `enterKeyHint` in `otherProps` got filtered out since it is not considered as a dom prop.
- e.g. `<Input {...args} enterKeyHint="search" />`, `enterKeyHint` will be ignored.

It shows `return`, which is not expected.

image

## 🚀 New behavior

It shows `search`, which is not expected.

image

### DIFF
--- a/.changeset/metal-peas-act.md
+++ b/.changeset/metal-peas-act.md
@@ -1,0 +1,5 @@
+---
+"@nextui-org/react-rsc-utils": patch
+---
+
+include enterKeyHint in dom-props (#2432)
--- a/packages/utilities/react-rsc-utils/src/dom-props.ts
+++ b/packages/utilities/react-rsc-utils/src/dom-props.ts
@@ -47,6 +47,7 @@ export const DOMPropNames = new Set([
   "draggable",
   "dropzone",
   "encType",
+  "enterKeyHint",
   "for",
   "form",
   "formAction",


jrgarciadev:
@wingkwong please add the changeset

wingkwong:
@jrgarciadev just added.

jrgarciadev (approved):

PR #2444, opened 2024-03-04
fix: isReadOnly in Autocomplete MDX

alphaxek:
Closes https://github.com/nextui-org/nextui/issues/2420

## 📝 Description

- revised the isReadOnly logic in Autocomplete
- created isReadOnly example in storybook

when isReadOnly is set, users

- cannot edit the input
- can open the listbox
- can select the menu option

## 🚀 New behavior
when isReadOnly is set, users

- cannot edit the input
- can open the listbox
- cannot select the menu option

No

## 📝 Additional Information
When someone prefer to not open the listbox, then they should use isDisabled instead.


wingkwong (changes requested):
I've made the [PR](https://github.com/nextui-org/nextui/pull/2429) 3 days ago. You don't need to copy my changes and add your own. You can simply supplement the doc part in this PR. In this case, a changeset is not required.

apps/docs/content/docs/components/autocomplete.mdx
```diff
@@ -68,6 +68,13 @@ the end of the label and the autocomplete will be required.
 
 <CodeDemo title="Required" highlightedLines="8" files={autocompleteContent.required} />
 
+### Read Only
+
+If you pass the `isReadOnly` property to the autocomplete, the listbox will open to display
```
jrgarciadev, line 73:
```suggestion
If you pass the `isReadOnly` property to the Autocomplete, the Listbox will open to display 
```

jrgarciadev (changes requested):
Hey @alphaxek please update this PR with the latest `main` changes

alphaxek:
> I've made the [PR](https://github.com/nextui-org/nextui/pull/2429) 3 days ago. You don't need to copy my changes and add your own. You can simply supplement the doc part in this PR. In this case, a changeset is not required.

Hi @wingkwong , I apologize I actually was trying to add my changes to your PR, but since it was my first time doing dual PR, I kind of messed it up, I didn't think it was going to create a whole new PR, I didn't intent to, if you want I can take it back, let me know, thanks

jrgarciadev (approved):
Thanks! @alphaxek

PR #2451, opened 2024-03-05
fix(components): getMenuTriggerProps in use-dropdown.ts returns origi…

### LINKED ISSUE #2448: [BUG] - getMenuTriggerProps in use-dropdown.ts returns originalProps as Object
### NextUI Version

2.1.17

### Describe the bug

In line 130 of [use-dropdown.ts](https://github.com/nextui-org/nextui/blob/9b27da544e950cdb2bccd2d1a57ead772fd10dc1/packages/components/dropdown/src/use-dropdown.ts) it seems the curly brackets include the originalProps Object, causing it to appear in HTML as originalProps=[Object Object]. Did you mean:

...mergeProps(otherMenuTriggerProps, {isDisabled: props.isDisabled}, originalProps),

or

...mergeProps(otherMenuTriggerProps, {isDisabled: props.isDisabled, ...originalProps}), 

or is that correct behaviour? 

_No response_

Use the nextUI Dropdown in version 2.1.17.

### Expected behavior

No property originalprops="[object Object]" in HTML

### Screenshots or Videos

_No response_

### Operating System Version

macOS

### Browser

Chrome

kuri-sun:
…nalProps as Object

Closes #2448

## 📝 Description

As per request from @wingkwong for the review of [this PR](https://github.com/nextui-org/nextui/pull/2450), since my last change was affecting this matter.

`isDisabled` does not disable dropdown.

## 🚀 New behavior

- To fix this, we intentionally use `props.isDisabled` to take the `isDisabled` prop from the Dropdown component in effect as well.
- No more appear in HTML as `originalProps=[Object Object]`.

https://github.com/nextui-org/nextui/assets/62743644/cbbc7e18-0e24-488e-83c7-c8dfd968a0a4

No.

## 📝 Additional Information

No.

### DIFF
--- a/packages/components/dropdown/src/use-dropdown.ts
+++ b/packages/components/dropdown/src/use-dropdown.ts
@@ -127,7 +127,7 @@ export function useDropdown(props: UseDropdownProps) {
     const {onKeyDown, onPress, onPressStart, ...otherMenuTriggerProps} = menuTriggerProps;
 
     return {
-      ...mergeProps(otherMenuTriggerProps, {isDisabled}, originalProps),
+      ...mergeProps(otherMenuTriggerProps, {isDisabled: props.isDisabled}, originalProps),
       ref: mergeRefs(_ref, triggerRef),
     };
   };
--- a/.changeset/lemon-elephants-shout.md
+++ b/.changeset/lemon-elephants-shout.md
@@ -1,0 +1,5 @@
+---
+"@nextui-org/dropdown": major
+---
+
+a small support fix for getMenuTriggerProps in use-dropdown.ts returns originalProps as Object


packages/components/dropdown/src/use-dropdown.ts
```diff
@@ -127,7 +127,7 @@ export function useDropdown(props: UseDropdownProps) {
     const {onKeyDown, onPress, onPressStart, ...otherMenuTriggerProps} = menuTriggerProps;
 
     return {
-      ...mergeProps(otherMenuTriggerProps, {isDisabled}, originalProps),
+      ...mergeProps(otherMenuTriggerProps, {isDisabled: props.isDisabled}, originalProps),
```
jrgarciadev, line 130: @kuri-sun the `isDisabled` is already being extracted from the `props` please check this line https://github.com/nextui-org/nextui/pull/2451/files#diff-88cdde9752c6237b962186b11532b521dd8ed7fd98b684e90b050993124cf2c9R53

kuri-sun: @jrgarciadev and @wingkwong 
When the `isDisabled` does not get passed to the Dropdown component, it always becomes `isDisabled=false`(https://github.com/nextui-org/nextui/blob/9b27da544e950cdb2bccd2d1a57ead772fd10dc1/packages/components/dropdown/src/use-dropdown.ts#L53). 
So I was worrying about this situation: the trigger Button component was disabled however the still is enabled.

1.  `isDisabled` (Current)

https://github.com/nextui-org/nextui/assets/62743644/58d7a3ea-f9f5-4dc7-b053-d87a2c90b8dd

3. `props.isDisabled` (This one)

https://github.com/nextui-org/nextui/assets/62743644/a763a604-d2a3-45e2-91f0-872d11c6b209



By setting `{isDisabled: props.isDisabled}`, it becomes `isDisabled: undefined`, so that coming the Button's `isDisabled` can override that, if there is any `isDisabled` on the Button component.

However, if we can ignore the Button component's `isDisabled` in this case, I would agree with going with it as it is! 😄 

Sorry for the long explanation... I hope this makes sense! Thank you! ❤️

wingkwong: Okay. I finally got what you mean. I can see why we need `undefined` here now. It wasn't that obvious though. @jrgarciadev Should we cover this case as well?

jrgarciadev: Hey guys, I fixed it here https://github.com/nextui-org/nextui/pull/2452/files

PR #2491, opened 2024-03-10
fix(table): hide empty content if isLoading is true (#2470)

### LINKED ISSUE #2470: [BUG] - Table: specifying both emptyContent and loadingContent shows both of them while content is loading
### NextUI Version

2.2.9

### Describe the bug

If I'm loading data dynamically, and use both `loadingContent` and `emptyContent`, NextUI seems to show both of them while `isLoading` is `true` and then either keep `emptyContent` if I get nothing back, or both of them disappear if the API returns actual content.

image

Tangential to this, you'll also notice that when there's no `emptyContent` defined and `items` is empty (`[]`), the loading element will be rendered on top of the columns. I think it's missing a minimum `height` in its definition.

https://stackblitz.com/edit/nextui-table-bug

1. Use Table
2. Ensure isLoading is true
3. Ensure emptyContent is non-empty
4. Ensure items is an empty array ([])

### Expected behavior

loadingContent and emptyContent should be mutually exclusive. While loadingContent is being rendered, emptyContent should not be used, and vice-versa.

### Screenshots or Videos

_No response_

### Operating System Version

Windows

### Browser

Edge

brunokc:
Closes #2470

## 📝 Description

This change ensures that `emptyContent` is hidden if `isLoading` is `true`

Currently, if `loadingContent` and `emptyContent` are both provided, and `isLoading` is true, NextUI will render both of them on top of each other.

## 🚀 New behavior

This change suppresses the rendering of `emptyContent` if `isLoading` is true.

No.

## 📝 Additional Information

Note that there's an existing issue with `loadingContent` where if `emptyContent` is **not** provided, NextUI will render `loadingContent` on top of the columns (instead of inside the table). This change doesn't fix that, but rather make `isLoading`/`loadingContent` play nice with a provided `emptyContent`. A separate fix is still needed for when `emptyContent` is not provided.

### DIFF
--- a/packages/components/table/src/table-body.tsx
+++ b/packages/components/table/src/table-body.tsx
@@ -105,7 +105,7 @@ const TableBody = forwardRef<"tbody", TableBodyProps>((props, ref) => {
           colSpan={collection.columnCount}
           role="gridcell"
         >
-          {bodyProps.emptyContent}
+          {!isLoading && bodyProps.emptyContent}
         </td>
       </tr>
     );


brunokc:
Hi @jrgarciadev @tianenpang @wingkwong,

This change seems to be enough for `isLoading` and `emptyContent` to play nice in my scenario, where both are supplied. However, I had to change my original fix so that `loadingContent` renders in the right place (inside the table instead of on top of the columns), but that only works if `emptyContent` is provided. 

I'm not sure this is acceptable. I intended to fix the "`loadingContent` without `emptyContent`" rendering issue but I don't know enough about the framework to propose a fix. I'd be happy to follow your guidance and try to fix it, if you have any suggestions.

jrgarciadev (approved):
Thank you! @brunokc

alphaxek:
> Hi @jrgarciadev @tianenpang @wingkwong,
> 
> This change seems to be enough for `isLoading` and `emptyContent` to play nice in my scenario, where both are supplied. However, I had to change my original fix so that `loadingContent` renders in the right place (inside the table instead of on top of the columns), but that only works if `emptyContent` is provided.
> 
> I'm not sure this is acceptable. I intended to fix the "`loadingContent` without `emptyContent`" rendering issue but I don't know enough about the framework to propose a fix. I'd be happy to follow your guidance and try to fix it, if you have any suggestions.

I will raise PR for this

PR #2589, opened 2024-03-26
fix(hooks): handle numeric selectedKeys in Select 

### LINKED ISSUE #2587: [BUG] - Select controlled mode not support number key
### NextUI Version

2.2.10

[https://stackblitz.com/\~/github.com/Zeiion/react-nextui](https://stackblitz.com/\~/github.com/Zeiion/react-nextui)

1. npm install & npm run dev
2. the first select is not default selected

### Expected behavior

\`\` is expect to support React.Key (string | number), so the first select should be default selected and works well when selecting new items

### Screenshots or Videos

CleanShot 2024-03-26 at 16 49 12

### Operating System Version

macOS

### Browser

Chrome

### Describe the bug

the `selectedKeys` props for the \`\` is not working when keys are numeric
but when i parse them into strings, it works.

The problem seems to appear in `packages/hooks/use-aria-multiselect/src/use-multiselect-list-state.ts`

```typescript
const {
    collection,
    disabledKeys,
    selectionManager,
    selectionManager: {setSelectedKeys, selectedKeys, selectionMode},
  } = useListState(props);
```

when call `collection.getItem(1)`, it gets `undefined`

sb test to reproduce the bug can be add as

```typescript
const ControlledNumberTemplate = ({color, variant, ...args}: SelectProps<User>) => {
  const [value, setValue] = React.useState<Selection>(new Set([1]));

  const handleSelectionChange = (e: ChangeEvent<HTMLSelectElement>) => {
    setValue(new Set([Number(e.target.value)]));
  };

  return (
    <div className="flex flex-col w-full max-w-xs gap-2">
      <Select
        className="max-w-xs mt-8"
        color={color}
        items={usersData}
        label="Assigned to"
        selectedKeys={value}
        variant={variant}
        onChange={handleSelectionChange}
        {...args}
      >
        {(item) => (
          <SelectItem key={item.id} textValue={item.name}>
            <div className="flex items-center gap-2">
              <Avatar alt={item.name} className="flex-shrink-0" size="sm" src={item.avatar} />
              <div className="flex flex-col">
                <span className="text-small">{item.name}</span>
                <span className="text-tiny text-default-400">{item.email}</span>
              </div>
            </div>
          </SelectItem>
        )}
      </Select>
      <p className="text-default-500">Selected: {value}</p>
    </div>
  );
};
```

wingkwong:
Closes #2587

## 📝 Description

In collection, the key of the map is always string no matter the inputted key is numeric or string. Therefore, handle such case when we get the item from the collection.

image

## 🚀 New behavior

image

No

## Summary by CodeRabbit

- **Bug Fixes**
    - Improved handling of numeric keys in the multi-select component to ensure consistent behavior.

### DIFF
--- a/packages/hooks/use-aria-multiselect/src/use-multiselect-list-state.ts
+++ b/packages/hooks/use-aria-multiselect/src/use-multiselect-list-state.ts
@@ -29,7 +29,7 @@ export function useMultiSelectListState<T extends object>(
     if (selectedKeys.size !== 0) {
       return Array.from(selectedKeys)
         .filter(Boolean)
-        .filter((key) => !collection.getItem(key));
+        .filter((key) => !collection.getItem(key + ""));
     }
 
     return [];
@@ -39,7 +39,7 @@ export function useMultiSelectListState<T extends object>(
     selectedKeys.size !== 0
       ? Array.from(selectedKeys)
           .map((key) => {
-            return collection.getItem(key);
+            return collection.getItem(key + "");
           })
           // Remove undefined values when some keys are not present in the collection
           .filter(Boolean)
--- a/packages/hooks/use-aria-multiselect/src/use-multiselect-list-state.ts
+++ b/packages/hooks/use-aria-multiselect/src/use-multiselect-list-state.ts
@@ -29,7 +29,7 @@ export function useMultiSelectListState<T extends object>(
     if (selectedKeys.size !== 0) {
       return Array.from(selectedKeys)
         .filter(Boolean)
-        .filter((key) => !collection.getItem(key + ""));
+        .filter((key) => !collection.getItem(`${key}`));
     }
 
     return [];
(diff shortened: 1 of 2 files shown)


packages/hooks/use-aria-multiselect/src/use-multiselect-list-state.ts
```diff
@@ -29,7 +29,7 @@ export function useMultiSelectListState<T extends object>(
     if (selectedKeys.size !== 0) {
       return Array.from(selectedKeys)
         .filter(Boolean)
-        .filter((key) => !collection.getItem(key));
+        .filter((key) => !collection.getItem(key + ""));
```
jrgarciadev, line 32:
```suggestion
        .filter((key) => !collection.getItem(`${key}`));
```

packages/hooks/use-aria-multiselect/src/use-multiselect-list-state.ts
```diff
@@ -39,7 +39,7 @@ export function useMultiSelectListState<T extends object>(
     selectedKeys.size !== 0
       ? Array.from(selectedKeys)
           .map((key) => {
-            return collection.getItem(key);
+            return collection.getItem(key + "");
```
jrgarciadev, line 42:
```suggestion
            return collection.getItem(`${key}`);
```

jrgarciadev (approved):

PR #2595, opened 2024-03-27
fix(checkbox): incorrect onChange typing in Checkbox Group

### LINKED ISSUE #2594: [BUG] Checkbox group onchange incorrect type
### NextUI Version

2.2.10

### Describe the bug

I'm trying to create a `CheckboxGroup` with `onChange` event, but cannot really get the type correct. 
For example, if I try to do `const handleOnChange = (values: string[]) => {}` then TypeScript will complain the following. 

```
Type '(values: string[]) => void' is not assignable to type 'FormEventHandler<HTMLDivElement> & ((value: string[]) => void)'.
  Type '(values: string[]) => void' is not assignable to type 'FormEventHandler<HTMLDivElement>'.
    Types of parameters 'values' and 'event' are incompatible.
      Type 'FormEvent<HTMLDivElement>' is missing the following properties from type 'string[]': length, pop, push, concat, and 35 more.ts(2322)
```

https://stackblitz.com/edit/stackblitz-starters-eft3qw?file=app%2Fpage.tsx

See the stackblitz URL. 

### Expected behavior

No type error according to documentation: https://nextui.org/docs/components/checkbox-group#checkbox-group-events

### Screenshots or Videos

image

### Operating System Version

macOS

### Browser

Chrome

wingkwong:
Closes #2594

## 📝 Description

as titled

image

## 🚀 New behavior

image

## Summary by CodeRabbit

- **Bug Fixes**
	- Fixed an issue with incorrect `onChange` typing in Checkbox Group, ensuring it now correctly handles an array of strings as values.

### DIFF
--- a/packages/components/checkbox/src/use-checkbox-group.ts
+++ b/packages/components/checkbox/src/use-checkbox-group.ts
@@ -14,7 +14,7 @@ import {clsx, safeAriaLabel} from "@nextui-org/shared-utils";
 
 import {CheckboxProps} from "./index";
 
-interface Props extends Omit<HTMLNextUIProps<"div">, "onChange"> {
+interface Props extends HTMLNextUIProps<"div"> {
   /**
    * Ref to the DOM node.
    */
@@ -44,13 +44,9 @@ interface Props extends Omit<HTMLNextUIProps<"div">, "onChange"> {
    * React aria onChange event.
    */
   onValueChange?: AriaCheckboxGroupProps["onChange"];
-  /**
-   * Handler that is called when the value changes.
-   */
-  onChange?: (values: string[]) => void;
 }
 
-export type UseCheckboxGroupProps = Props &
+export type UseCheckboxGroupProps = Omit<Props, "onChange"> &
   AriaCheckboxGroupProps &
   Partial<
     Pick<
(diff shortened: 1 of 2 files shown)


packages/components/checkbox/src/use-checkbox-group.ts
```diff
@@ -14,7 +14,7 @@ import {clsx, safeAriaLabel} from "@nextui-org/shared-utils";
 
 import {CheckboxProps} from "./index";
 
-interface Props extends HTMLNextUIProps<"div"> {
+interface Props extends Omit<HTMLNextUIProps<"div">, "onChange"> {
```
ryo-manba, line 17: As indicated by the warning, the issue resides with `FormEventHandler<HTMLDivElement>`. Therefore, I believe the following code should be corrected.
https://github.com/nextui-org/nextui/blob/main/packages/components/checkbox/src/use-checkbox-group.ts#L49

I think it would suffice to simply remove `onChange` from `Props`.
```
export type UseCheckboxGroupProps = Omit<Props, "onChange"> &
```

This would eliminate the need to redefine `onChange`.

wingkwong: Thanks. much cleaner.

ryo-manba (approved):

jrgarciadev (approved):

PR #2597, opened 2024-03-27
fix(radio): isRequired & missing warning message in Form

wingkwong:
Closes #2585

## 📝 Description

- avoid overriding required prop
- merge with domRef to keep `validationMessage`

## ⛳️ Current behavior (updates)

[pr2597-issue-demo.webm](https://github.com/nextui-org/nextui/assets/35857179/283ffac0-0aed-4e81-b163-fb69a5ff723e)

## 🚀 New behavior

[pr2597-demo.webm](https://github.com/nextui-org/nextui/assets/35857179/ccb6b65b-37dc-435b-a082-d3b44b944f01)

## 💣 Is this a breaking change (Yes/No):

No

## Summary by CodeRabbit

- **Bug Fixes**
	- Patched the radio component to ensure all required properties are present and validation messages are correctly displayed.
	- Improved the radio component's handling of input properties for better performance and reliability.

PR #2601, opened 2024-03-28
fix: pnpm-lock & typing issues

wingkwong:
Closes # 

## 📝 Description

Current Problem:

if we remove pnpm-lock.yaml and run pnpm install again, we can see the changes introduced. 

previously we couldn't reproduce some issues reported by users locally but in the build package (e.g. isRequired prop not working in radio group). Tests didn't catch the failed case as well. With the updated pnpm-lock.yaml, the issue can be reproducible & the test fails expectedly.

After updating pnpm-lock.yaml, some typings would fail causing the build to fail. This ticket is to resolve all the build issues. Failing test cases will be handled separately.

## 🚀 New behavior

## 📝 Additional Information

There are two test cases failing

1. input ref value test case: handled in #2613
2. isRequired test case issue: handled in #2597

### DIFF
--- a/packages/components/checkbox/src/checkbox-group.tsx
+++ b/packages/components/checkbox/src/checkbox-group.tsx
@@ -26,7 +26,7 @@ const CheckboxGroup = forwardRef<"div", CheckboxGroupProps>((props, ref) => {
         <CheckboxGroupProvider value={context}>{children}</CheckboxGroupProvider>
       </div>
       {errorMessage ? (
-        <div {...getErrorMessageProps()}>{errorMessage}</div>
+        <div {...getErrorMessageProps()}>{errorMessage as React.ReactNode}</div>
       ) : description ? (
         <div {...getDescriptionProps()}>{description}</div>
       ) : null}
--- a/packages/hooks/use-aria-press/src/index.ts
+++ b/packages/hooks/use-aria-press/src/index.ts
@@ -49,6 +49,8 @@ const LINK_CLICKED = Symbol("linkClicked");
  * Handles press interactions across mouse, touch, keyboard, and screen readers.
  * It normalizes behavior across browsers and platforms, and handles many nuances
  * of dealing with pointer and keyboard events.
+ *
+ * @deprecated use `usePress` instead from `react-aria`
  */
 export function usePress(props: PressHookProps): PressResult {
   let {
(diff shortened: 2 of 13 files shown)


packages/components/checkbox/src/checkbox-group.tsx
```diff
@@ -26,7 +26,7 @@ const CheckboxGroup = forwardRef<"div", CheckboxGroupProps>((props, ref) => {
         <CheckboxGroupProvider value={context}>{children}</CheckboxGroupProvider>
       </div>
       {errorMessage ? (
-        <div {...getErrorMessageProps()}>{errorMessage}</div>
```
jrgarciadev, line 29: @wingkwong why do we need to explicitly set the type? this should come from `@react-types/shared`

wingkwong: `UseCheckboxGroupProps` -> `AriaCheckboxGroupProps` -> `CheckboxGroupProps` -> `HelpTextProps`

```ts
export interface HelpTextProps {
  /** A description for the field. Provides a hint such as specific requirements for what to choose. */
  description?: ReactNode,
  /** An error message for the field. */
  errorMessage?: ReactNode | ((v: ValidationResult) => ReactNode)
}
```

setting explicitly since we don't need to use it as a function

packages/hooks/use-aria-press/src/index.ts
```diff
@@ -709,7 +713,7 @@ export function usePress(props: PressHookProps): PressResult {
       };
 
       let onScroll = (e: Event) => {
-        if (state.isPressed && (e.target as Element).contains(state.target)) {
+        if (state.isPressed && state.target && (e.target as Element).contains(state.target)) {
           cancel({
             currentTarget: state.target,
             shiftKey: false,
```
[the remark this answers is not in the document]
wingkwong: Updated

packages/components/checkbox/src/checkbox-group.tsx
jrgarciadev: @wingkwong  I think we will need to support the function one once @ryo-manba finishes the react-aria migration

ryo-manba:
@wingkwong @jrgarciadev

Due to our use of the caret (`^`) version specifier in package.json, re-installing packages after deleting pnpm-lock.yaml can unexpectedly update to the latest minor versions. This has caused issues, especially with [react-aria updates that brought breaking changes in validation](https://react-spectrum.adobe.com/releases/2023-11-8.html), which we're addressing in https://github.com/nextui-org/nextui/pull/2561.

Therefore, it would be helpful if we could close this PR. Thank you!

wingkwong:
As discussed with @ryo-manba, PR2561 would address the issues. It's better to have those changes from his PR due to some dependencies with latest react-aria changes. Therefore, I'm closing this PR.

PR #2603, opened 2024-03-28
fix: react hook form issue

### LINKED ISSUE #1969: [BUG] - react-hook-form
### NextUI Version

2.2.9

### Describe the bug

`

<Input
          type="text"
          placeholder="0"
          classNames={{
            input: "text-right",
          }}
          radius="none"
          onValueChange={(value: string) => inputChange(value, "amount")}
          size="sm"
          {...register("amount", {
            required: true,
          })}
    />
`

It uses React-hook-form and next ui input, and is initialized when the value is entered with setValue.

_No response_

""

### Expected behavior

""

### Screenshots or Videos

_No response_

### Operating System Version

masOs

### Browser

Chrome

wingkwong:
~~**_build & tests issues will be handled in PR #2561_**~~

Closes #1969

## 📝 Description

handle RHF on NextUI Form component

- [x] input
- [x] checkbox
- [x] select
- [x] autocomplete 

Currently if we use react-hook-form to register NextUI components, it is not working as expected. For example, setting a default value won't reflect in UI. The reason is that react-hook-form changes the value using ref directly while in nextui we handle the state value separately. This PR is to sync the uncontrolled value to the nextui state.

## 🚀 New behavior


jrgarciadev:
@wingkwong could you please fix the conflicts?

PR #2649, opened 2024-04-05
fix(select): disable all keys for isDisabled case

### LINKED ISSUE #2345: [BUG] - Disabled Select Allows Changes Using Blur + Keyboard
### NextUI Version

2.2.9

### Describe the bug

When setting a Select to disabled, the select still allows the user to change the value of the component by typing in the first letter of an item. 

I've highlighted the select on chrome using tab, then pressed 's' to change the value to Shark instead of Cat.
I can also do this by clicking on the label of the element, and then pressing a key.

https://stackblitz.com/edit/stackblitz-starters-umk7zm?file=app%2Fpage.tsx

1. Select the disabled element using tab, or by clicking the select's label
2. Type a character that corresponds to a different select item's value.
3. Notice the disabled input has had its value changed.

### Expected behavior

If the element is set as disabled, keyboard inputs should not allow the value to change.

### Screenshots or Videos

_No response_

### Operating System Version

macOS, Windows

### Browser

Chrome

wingkwong:
Closes #2345

## 📝 Description

> Add a brief description

## 🚀 New behavior

### DIFF
--- a/.changeset/nervous-geckos-visit.md
+++ b/.changeset/nervous-geckos-visit.md
@@ -1,0 +1,5 @@
+---
+"@nextui-org/select": patch
+---
+
+Fixed Disabled Select Allows Changes Using Blur + Keyboard (#2345)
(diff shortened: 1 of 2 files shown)


ShrinidhiUpadhyaya (commented):
This `disabledKeys: new Set([...state.collection.getKeys()].map((k) => k)),` could be changed to  `disabledKeys: new Set([...state.collection.getKeys()]),`. This `map((k) => k)` is redundant.

jrgarciadev (approved):

PR #2709, opened 2024-04-14
fix(modal): input carry over with IMEs in modal forms

ryo-manba:
Closes https://github.com/nextui-org/nextui/issues/2445

## 📝 Description

Fixed an issue where typing with an IME (like Korean or Japanese) and pressing tab would carry the input to the next field. This was resolved using the [isComposing property](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/isComposing).

https://github.com/nextui-org/nextui/assets/76232929/0c7a228f-677a-4f0d-a930-7adc8b3419dd

When entering text in a modal form using Korean, pressing the tab key causes the last character to be entered into the next input field.

## 🚀 New behavior

Pressing the tab key after entering text in Korean no longer transfers the last character to the next input field. 

No.

## 📝 Additional Information

This bug does not occur in Storybook, so the behavior was verified using the [docs](https://nextui.org/docs/components/modal#with-form).
This was tested using both Japanese and 2-set Korean IMEs on browsers including Chrome, Safari, and Firefox.

## Summary by CodeRabbit

- **Bug Fixes**
	- Implemented a fix in modal components to prevent carryover of IME (Input Method Editor) input when switching fields using the Tab key.
	- Enhanced accessibility by handling Tab key press event in the `ModalContent` component.

### DIFF
--- a/.changeset/old-cameras-sip.md
+++ b/.changeset/old-cameras-sip.md
@@ -1,0 +1,5 @@
+---
+"@nextui-org/modal": patch
+---
+
+prevent IME input carryover in form fields when tabbing
--- a/.changeset/old-cameras-sip.md
+++ b/.changeset/old-cameras-sip.md
@@ -2,4 +2,4 @@
 "@nextui-org/modal": patch
 ---
 
-prevent IME input carryover in form fields when tabbing
+Prevent IME input carryover in form fields when tabbing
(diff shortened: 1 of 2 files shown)


jrgarciadev (approved):
Thanks! @ryo-manba

PR #2721, opened 2024-04-15
fix(cmdk): modal - searching with one char shows no results & cmdk esc does not work without recent search results

### LINKED ISSUE #2643: [BUG] - Modal - searching with one char shows no results
### NextUI Version

2.2.10

### Describe the bug

When users search from the search-modal with one single char like the pic, it will fail with no result.

_No response_

Reproduce process is the same as above.

### Expected behavior

It should show that "No results for "s"

Try adding more characters to your search term." or sth like that.

### Screenshots or Videos

https://github.com/nextui-org/nextui/assets/62743644/7410ad4a-ccf0-4266-b3de-f80d1b6ae63e

### Operating System Version

macOS

### Browser

Chrome

kuri-sun:
Closes #2643

## 📝 Description

*make sure that you remove the recent search results.
 - When users search from the search-modal with one single char like the pic, it will fail with no result.
 -  Cmdk ESC does not work when there is no recent search results cache.

one char searching will show no result in the search modal.

## 🚀 New behavior

https://github.com/nextui-org/nextui/assets/62743644/e7907c29-ec8a-4863-939a-ffaa4de8ee0c

No.

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved the display logic for the "No results" message in the command palette to ensure it shows correctly based on the query length.


apps/docs/components/cmdk.tsx
```diff
@@ -266,7 +266,7 @@ export const Cmdk: FC<{}> = () => {
         }
       }
     },
-    [activeItem, items, router],
+    [activeItem, items, router, query],
```
wingkwong, line 269: do we need router and query here? The function seems not depend on these.

wingkwong (approved):

PR #2796, opened 2024-04-19
fix(input): isClearable & add test cases to input

### LINKED ISSUE #2791: [BUG] - Input Clear Button No Longer Working
### NextUI Version

2.3.5

### Describe the bug

When you try to clear the input using the clear button, the value is emptied, but the input still displays the uncleared text.

You can try it directly on the docs page: https://nextui.org/docs/components/input#clear-button

_No response_

You can try it directly on the docs page: https://nextui.org/docs/components/input#clear-button and try to clear the input.

### Expected behavior

I still see the text I tried to clear and I cannot get rid of it.

### Screenshots or Videos

_No response_

### Operating System Version

macOS Sonoma 14.3.1 (23D60)

### Browser

Chrome

wingkwong:
Closes #2791

## 📝 Description

- use `useSafeLayoutEffect` to handle react-hook-form case. consistent with others
- add test case for isClearable
- add react hook form to input storybook (for test case, will be handled in another PR)

with `isClearable`, users fail to clear the value due to the previous change

## 🚀 New behavior

with `isClearable`, users are able to clear the value.

[pr2796-demo.webm](https://github.com/nextui-org/nextui/assets/35857179/39ee8d35-502f-41cb-b51b-d6f28c153f0b)

No

## Summary by CodeRabbit

- **New Features**
    - Enhanced the input component to support clear functionality with the `isClearable` option.
    - Introduced `WithReactHookForm` component for better form handling and validation using `react-hook-form`.

- **Bug Fixes**
    - Fixed an issue where the input value was not properly cleared when using the clear button.

- **Tests**
    - Added tests to ensure the clear functionality triggers the `onClear` callback correctly.

- **Dependencies**
    - Added `react-hook-form` to improve form state management and validation.

### DIFF
--- a/packages/components/input/__tests__/input.test.tsx
+++ b/packages/components/input/__tests__/input.test.tsx
@@ -1,5 +1,6 @@
 import * as React from "react";
-import {render} from "@testing-library/react";
+import {act, render} from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
 
 import {Input} from "../src";
 
@@ -115,5 +116,33 @@ describe("Input", () => {
     container.querySelector("input")?.focus();
 
     expect(ref.current?.value)?.toBe(value);
+  });
+
+  it("should clear the value and onClear is triggered", async () => {
+    const onClear = jest.fn();
+
+    const ref = React.createRef<HTMLInputElement>();
+
+    const {container} = render(
+      <Input
+        ref={ref}
+        isClearable
+        defaultValue="junior@nextui.org"
+        label="test input"
+        onClear={onClear}
+      />,
+    );
+
+    const clearButton = container.querySelector("[data-slot='clear-button']") as HTMLElement;
+
+    expect(clearButton).not.toBeNull();
+
+    await act(async () => {
+      await userEvent.click(clearButton);
+    });
+
+    expect(ref.current?.value)?.toBe("");
+
+    expect(onClear).toHaveBeenCalledTimes(1);
   });
 });
--- a/packages/components/input/__tests__/input.test.tsx
+++ b/packages/components/input/__tests__/input.test.tsx
@@ -1,5 +1,5 @@
 import * as React from "react";
-import {act, render} from "@testing-library/react";
+import {render} from "@testing-library/react";
 import userEvent from "@testing-library/user-event";
 
 import {Input} from "../src";
@@ -123,7 +123,7 @@ describe("Input", () => {
 
     const ref = React.createRef<HTMLInputElement>();
 
-    const {container} = render(
+    const {getByRole} = render(
       <Input
         ref={ref}
         isClearable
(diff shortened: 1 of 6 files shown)


packages/components/input/__tests__/input.test.tsx
```diff
@@ -116,4 +117,32 @@ describe("Input", () => {
...
+      <Input
+        ref={ref}
+        isClearable
+        defaultValue="junior@nextui.org"
+        label="test input"
+        onClear={onClear}
+      />,
+    );
+
+    const clearButton = container.querySelector("[data-slot='clear-button']") as HTMLElement;
```
ryo-manba, line 136: It seems better to use `getByRole` instead of `querySelector`, considering the [priority of queries].(https://testing-library.com/docs/queries/about/#priority)

```suggestion
    const {getByRole} = render(
      <Input
        ref={ref}
        isClearable
        defaultValue="junior@nextui.org"
        label="test input"
        onClear={onClear}
      />,
    );

    const clearButton = getByRole("button");
```

packages/components/input/__tests__/input.test.tsx
```diff
@@ -116,4 +117,32 @@ describe("Input", () => {
...
+
+    expect(clearButton).not.toBeNull();
+
+    await act(async () => {
+      await userEvent.click(clearButton);
+    });
```
ryo-manba, line 142: As recommended in the documentation:
> This exists to ease the transition to version 14 and writing simple tests. Using the methods on the instances returned by userEvent.setup() is recommended.

https://testing-library.com/docs/user-event/setup/#direct-apis

```suggestion
    const user = userEvent.setup();

    await user.click(clearButton);
```

packages/components/input/stories/input.stories.tsx
```diff
@@ -474,6 +475,38 @@ const CustomWithHooksTemplate = (args: InputProps) => {
...
+      withDefaultValue: "wkw",
+      withoutDefaultValue: "",
+      requiredField: "",
+    },
+  });
+
+  const onSubmit = (data: any) => {
+    // eslint-disable-next-line no-console
+    console.log(data);
+    alert("Submitted value: " + JSON.stringify(data));
```
ryo-manba, line 494: I don't need console.log because I can check it with alert. 
What do you think?

ryo-manba (commented):
Looks good overall! I've left some comments on a few points.

packages/components/input/stories/input.stories.tsx
wingkwong: I followed Junior's example done in checkbox story.

jrgarciadev (approved):

D4nGuerrero:
Updating the input value for React Hook Form actually introduces breaking changes. Previously, it would function with RHF's ASYNC default values because, as you mentioned, RHF uses ref.value. However, the value is now overwritten due to setInputValue being called first, which sets it to initial ref.value. As a result, async values come back, and RHF updates ref.value. However, NextUI's input state is already set so changing ref.value  only reflects changes temporarily (and for input only), NextUI will use its own input state value.

Visual example:
https://github.com/nextui-org/nextui/assets/20586092/81531e3b-0694-4134-a203-05ac91c99a9b

Workarounds:
Setting the values yourself after async response, or a quicker way is to conditionally render Input/Select since they are not in the DOM yet this won't activate:
```
 useSafeLayoutEffect(() => {
    if (!domRef.current) return;

    setInputValue(domRef.current.value);
  }, [domRef.current]);
}

```

PR #2801, opened 2024-04-19
fix(docs): Date input - Time zone example code is not showing js (code mode)

withden:
## 📝 Description
In Date time (Time zones example), there is no need for ts and js separately for code.

This PR request 

Currently, Date time (Time zones example - docs), in `js` code there Is nothing (Sample app code)

## 🚀 New behavior

It will be the same for ts and js code

No

## Summary by CodeRabbit

- **New Features**
	- Enhanced date input components to support converting absolute time to local time zones.


wingkwong (changes requested):
- please also import `parseAbsoluteToLocal`.
- if possible, please group the similar changes in one PR

wingkwong (approved):

PR #2815, opened 2024-04-20
fix: pickChildren breaks when components are wrapped in HOC

sapkra:
+ updated react types, fixed alphabetical sorting in package.json

## 📝 Description

In my project I'm wrapping every component with an HOC, which leads to the problem that pickChildren won't detect the right component type. This change will use the displayName instead which can be passed down by the HOC.

- compares an entire component object

## 🚀 New behavior

- compares just the component name

No, I don't think so. Already tested a few cases.

## Summary by CodeRabbit

- **Updates**
	- Updated React and React DOM type definitions to improve compatibility and developer experience.
- **Refactor**
	- Enhanced child component selection in utilities to allow filtering by display name, improving component interaction flexibility.


wingkwong (commented):
can you please share a failing example for us to test?

wingkwong:
Closing - inactivity

PR #2822, opened 2024-04-20
docs(button): add missing type prop and its default

Sajgoniarz:
Closes # 

## 📝 Description

Filled documentation for type property of Button.

Current docs are not stating the default, nor possible values for button's type property.

## 🚀 New behavior

Default type and it's values can be found in docs.

No

## Summary by CodeRabbit

- **New Features**
	- Added a new property `type` for buttons, allowing specification as `button`, `submit`, or `reset`.


wingkwong:
Thanks for the contribution. However, this section is for the props specific to NextUI Button. What you added is just a general HTML attribute, which by default, we extend for each component. Therefore, they won't be shown in such section.

Sajgoniarz:
Helo @wingkwong and thanks for response.
Indeed, I added general HTML attribute, but what I care the most about in this PR is to document it's default value that you have overridden from default HTML's one, so from `submit` to `button`.
I think that overriding props in components which can inherit them from native HTML elements is a good thing to have documented, since it can lead to confusion.

PR #2937, opened 2024-05-03
fix(select): onSelectionChange can handle number

ryo-manba:
Closes https://github.com/nextui-org/nextui/issues/2926

## 📝 Description

Fixed an issue where specifying a `number` as the `key` for `SelectItem` resulted in `onSelectionChange` treating it as a `string`.

When a `number` is specified as the `key` for `SelectItem`, a `string` is passed to `onSelectionChange`.

## 🚀 New behavior

Now, when a `number` is specified as the `key` for `SelectItem`, `onSelectionChange` will receive it as a `number`.

No.

## 📝 Additional Information

This issue stems from React converting the key to a string.

> Note that if you do specify a custom key on each Item using the key prop, React will convert the provided key to a string. As a result, all the collection component's key-related event handlers and props (e.g. onSelectionChange, selectedKey(s)) will expect and return strings.

See: https://react-spectrum.adobe.com/react-stately/collections.html#unique-keys

## Summary by CodeRabbit

- **New Features**
  - Enhanced item selection structure in components for improved user experience.
  - Updated documentation examples for better clarity and usability.

- **Bug Fixes**
  - Corrected rendering logic in the Select component to ensure accurate display of selected items and end content.

- **Documentation**
  - Aligned highlighted code section line numbers with current source code for accurate documentation examples.

- **Refactor**
  - Streamlined key access in multi-select list state hook for enhanced performance and simplicity.

### DIFF
--- a/apps/docs/content/components/select/colors.ts
+++ b/apps/docs/content/components/select/colors.ts
@@ -1,29 +1,17 @@
 const data = `export const animals = [
-  {label: "Cat", value: "cat", description: "The second most popular pet in the world"},
-  {label: "Dog", value: "dog", description: "The most popular pet in the world"},
-  {label: "Elephant", value: "elephant", description: "The largest land animal"},
-  {label: "Lion", value: "lion", description: "The king of the jungle"},
-  {label: "Tiger", value: "tiger", description: "The largest cat species"},
-  {label: "Giraffe", value: "giraffe", description: "The tallest land animal"},
-  {
-    label: "Dolphin",
-    value: "dolphin",
-    description: "A widely distributed and diverse group of aquatic mammals",
-  },
-  {label: "Penguin", value: "penguin", description: "A group of aquatic flightless birds"},
-  {label: "Zebra", value: "zebra", description: "A several species of African equids"},
-  {
-    label: "Shark",
-    value: "shark",
-    description: "A group of elasmobranch fish characterized by a cartilaginous skeleton",
-  },
-  {
-    label: "Whale",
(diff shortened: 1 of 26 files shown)


apps/docs/content/components/select/colors.ts
```diff
@@ -51,7 +39,7 @@ export default function App() {
           className="max-w-xs"
         >
           {animals.map((animal) => (
-            <SelectItem key={animal.value} value={animal.value}>
+            <SelectItem key={animal.key}>
```
ryo-manba, line 42: There is no need to give `value` to `SelectItem`, so it is removed.
See: https://nextui.org/docs/components/select#selectitem-props

apps/docs/content/components/select/description.ts
```diff
@@ -1,29 +1,17 @@
...
+  {key: "lion", label: "Lion"},
+  {key: "tiger", label: "Tiger"},
+  {key: "giraffe", label: "Giraffe"},
+  {key: "dolphin", label: "Dolphin"},
+  {key: "penguin", label: "Penguin"},
+  {key: "zebra", label: "Zebra"},
+  {key: "shark", label: "Shark"},
+  {key: "whale", label: "Whale"},
+  {key: "otter", label: "Otter"},
+  {key: "crocodile", label: "Crocodile"}
```
jrgarciadev, line 14: @ryo-manba I believe we need the description key for this example `description`

jrgarciadev (approved):

PR #2942, opened 2024-05-04
Update dependency array on setPage

jesuzon:
Changes:
Add the onChange function to the dependency array of the setPage callback to ensure it always reflects the latest state.

Impact:
This fix ensures that the pagination component accurately reflects the current state when triggering onChange.

Closes # 

## 📝 Description

This fixes an issue where search parameters were not updated correctly when changing pages in the pagination component. The root cause was the missing onChange dependency in the setPage callback function, which prevented the correct state from being passed on page changes.

Currently, state is memoized during onChange after page changes, meaning that writing onChange handlers do not always contain current state. This was causing issues with hooks such as useSearchParams, where the search params would not be fully up-to-date

## 🚀 New behavior

Adding onChange to the callback dependencies ensures state is refreshed.

Should not break anything.

## Summary by CodeRabbit

- **Refactor**
	- Improved the responsiveness of the pagination feature by ensuring changes are effectively tracked and updated.

### DIFF
--- a/packages/hooks/use-pagination/src/index.ts
+++ b/packages/hooks/use-pagination/src/index.ts
@@ -82,7 +82,7 @@ export function usePagination(props: UsePaginationProps) {
         onChangeActivePage(pageNumber);
       }
     },
-    [total, activePage],
+    [total, activePage, onChange],
   );
 
   const next = () => (isRTL ? setPage(activePage - 1) : setPage(activePage + 1));


jesuzon:
Fix to this open bug issue: https://github.com/nextui-org/nextui/issues/2417

wingkwong (commented):
why not onChangeActivePage?

jesuzon:
> why not onChangeActivePage?

I supposed they are equivalent in this context, one calls the other anyway - I have tested and that works as well.

wingkwong:
@jesuzon `setPage` doesn't depend on `onChange`. I could set the active page without triggering onChange.

jesuzon:
> @jesuzon `setPage` doesn't depend on `onChange`. I could set the active page without triggering onChange.

Changing the active page the intended documented way (using setPage rather than changing the internal state directly using setActivePage which you technically can’t do directly I think) triggers an onChange by calling onChangeActivePage: 

`const onChangeActivePage = (newPage: number) => {
    setActivePage(newPage);
    onChange && onChange(newPage);
  };
`

Do you want me to make a different commit adding onChangeActivePage instead of onChange? Either way, it should fix the issue I tagged above

jesuzon:
Any updates on this merge?

wingkwong:
> Do you want me to make a different commit adding onChangeActivePage instead of onChange? Either way, it should fix the issue I tagged above

Yes please.

wingkwong:
Closing - no longer applicable

PR #2994, opened 2024-05-13
fix: show `errorMessage` for inputs without having to pass `isInvalid`

remorses:
Closes https://github.com/nextui-org/nextui/issues/2835

If user provides an errorMessage to an input, the input is invalid, there is no need to ask the user to also pass the `isInvalid` prop

This change was introduced here: https://github.com/nextui-org/nextui/commit/f724509c83ce1586f08ac1042ccacb7ce54cbe83

This was a breaking change introduced 2 weeks ago

## Summary by CodeRabbit

- **Bug Fixes**
	- Adjusted the conditional rendering logic across various components to improve the visibility of error messages and descriptions. Error messages will now display whenever present, and descriptions will only show when relevant and the component is in an invalid state. This change affects:
		- Checkbox groups
		- Input fields
		- Text areas
		- Radio groups
		- Select dropdowns


ryo-manba:
This PR is closed. Please see the following comment for details.
https://github.com/nextui-org/nextui/pull/2994

PR #3000, opened 2024-05-14
fix(autocomplete): ignore pointer events when the clear button is hidden

chirokas:
Closes # 

## 📝 Description

- `opacity: 0` causes the elements to be very transparent but users can still interact with them.

- On mobile, this gives a bit of annoying results where after clicking the button, the keyboard is shown again because the input is focused again. See [RAC#6783](https://github.com/adobe/react-spectrum/pull/6783)

https://github.com/nextui-org/nextui/assets/157580465/56767b0a-0315-434b-b8fd-453cbc5147c2

## 🚀 New behavior

https://github.com/nextui-org/nextui/assets/157580465/2a6a16f6-5873-4561-8fa8-3088ccd6efb8

## Summary by CodeRabbit

- **New Features**
	- Enhanced AutoComplete component with improved visibility and interaction for the Clear Button.
	- Refined focus management for mobile users to ensure a seamless experience when using the Clear Button.

- **Bug Fixes**
	- Resolved keyboard issue that affected the reopening of the AutoComplete component on mobile devices.
	- Improved pointer event handling to prevent unintended interactions when the component is not visible.


wingkwong (commented):
I'm wondering if we need to keep both at the same time.

chirokas:
Visibility is an animatable property. However, the result looks terrible and gives a jarring effect. This is because it's a [discrete animation](https://www.w3.org/TR/web-animations/#animating-visibility).

A solution to this problem is to specify a transition on both `visibility` and `opacity`.

jrgarciadev:
@chirokas what about `pointer-events-none` ?

wingkwong (approved):

PR #3013, opened 2024-05-16
fix: make VisuallyHidden's element type as span when it's inside phrasing element

kosmotema:
## 📝 Description

The Checkbox has a `VisuallyHidden` component, that wraps the input. By default, it renders the `div` wrapper, but `div` is not allowed as a child of the `label` element (since `label` [accepts only phrasing content](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label#technical_summary))

The Checkbox currently renders this content (with a child of the disallowed `div` type):

```html
<label>
  <div style="...">
    <input />
  </div>
  ...
</label>
```

## 🚀 New behavior

Now the Checkbox renders `span` instead of `div` for `VisuallyHidden`, which is the allowed child of the `label` element:

```html
<label>
  <span style="...">
    <input />
  </span>
  ...
</label>
```

No

## Summary by CodeRabbit

- **Enhancements**
  - Improved accessibility by updating the `VisuallyHidden` component to use `elementType="span"` across Checkbox, Radio, Select, and Switch components.

### DIFF
--- a/.changeset/fluffy-seals-complain.md
+++ b/.changeset/fluffy-seals-complain.md
@@ -1,5 +1,0 @@
----
-"@nextui-org/checkbox": patch
----
-
-make VisuallyHidden's element type as span
--- a/.changeset/fluffy-seals-complain.md
+++ b/.changeset/fluffy-seals-complain.md
@@ -1,0 +1,5 @@
+---
+"@nextui-org/checkbox": patch
+---
+
+make VisuallyHidden's element type as span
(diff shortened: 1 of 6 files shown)


kosmotema:
> Can you also take a look at these files to see if you can apply the same change.

Yep, also applied this change to the Switch, Radio and Select components
The Select component still has a `div` as the child of the `Component` (which is `button` by default), but I think it's better to change this in a separate PR

Other components (like table parts or a slider thumb) have default `Component`s that accept `div` content, but the user can customize the element, so I don't know if we should also add `elementType="span"` for `VisuallyHidden` components there

wingkwong (approved):

jrgarciadev (approved):
Thanks! @kosmotema

jrgarciadev:
@kosmotema could you please fix the conflicts?

kosmotema:
@jrgarciadev Fixed

PR #3049, opened 2024-05-22
fix(hooks): optimize useScrollPosition with useCallback and useRef

Gaic4o:
## 📝 Description

Refactored the useScrollPosition hook to improve performance and stability by using useCallback for the handler function and useRef for throttleTimeout.

The current implementation of the useScrollPosition hook:
- Uses a local variable for throttleTimeout, which may cause issues with maintaining state between renders, leading to potential memory leaks and redundant timeout calls.
- Does not use useCallback for the handler function, which can lead to unnecessary re-renders as the handler function reference changes on each render.
- Event listeners might not be properly cleaned up, potentially causing multiple listeners to accumulate over time, which can degrade performance.

## 🚀 New behavior

- Wrapped the handler function with useCallback to ensure it does not cause unnecessary re-renders and always uses the latest values.
- Used useRef for throttleTimeout to maintain its value between renders and avoid resetting the timeout on every render.
- Added cleanup for throttleTimeout in useEffect's return function to prevent potential memory leaks.
- Ensured that event listeners are properly added and removed to avoid redundant event handlers.

No

## 📝 Additional Information

These changes enhance the performance and stability of the useScrollPosition hook without introducing breaking changes.

## Summary by CodeRabbit

- **Refactor**
  - Enhanced the `useScrollPosition` hook for better performance and reliability by utilizing `useCallback` and `useRef`.

### DIFF
--- a/.changeset/lucky-cobras-jog.md
+++ b/.changeset/lucky-cobras-jog.md
@@ -1,5 +1,5 @@
 ---
-"@nextui-org/use-scroll-position": major
+"@nextui-org/use-scroll-position": patch
 ---
 
 WHAT: Refactored the useScrollPosition hook to improve performance and stability by using useCallback for the handler function and useRef for throttleTimeout.
--- a/.changeset/lucky-cobras-jog.md
+++ b/.changeset/lucky-cobras-jog.md
@@ -1,0 +1,5 @@
+---
+"@nextui-org/use-scroll-position": major
+---
+
+WHAT: Refactored the useScrollPosition hook to improve performance and stability by using useCallback for the handler function and useRef for throttleTimeout.
--- a/packages/hooks/use-scroll-position/src/index.ts
+++ b/packages/hooks/use-scroll-position/src/index.ts
@@ -77,7 +77,7 @@ export const useScrollPosition = (props: UseScrollPositionOptions): ScrollValue 
         clearTimeout(throttleTimeout.current);
       }
     };
-  }, [elementRef?.current, delay, isEnabled]);
+  }, [elementRef?.current, delay, handler, isEnabled]);
 
   return position.current;
 };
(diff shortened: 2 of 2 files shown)


.changeset/lucky-cobras-jog.md
```diff
@@ -0,0 +1,5 @@
+---
+"@nextui-org/use-scroll-position": major
```
jrgarciadev, line 2:
```suggestion
"@nextui-org/use-scroll-position": patch
```

packages/hooks/use-scroll-position/src/index.ts
```diff
@@ -41,26 +41,28 @@ export const useScrollPosition = (props: UseScrollPositionOptions): ScrollValue
...
     if (!isEnabled) return;
 
     const handleScroll = () => {
       if (delay) {
-        if (throttleTimeout === null) {
-          throttleTimeout = setTimeout(handler, delay);
+        if (throttleTimeout.current === null) {
+          throttleTimeout.current = setTimeout(() => {
+            handler();
+          }, delay);
```
jrgarciadev, line 65:
```suggestion
        if (throttleTimeout.current === null) {
          throttleTimeout.current = setTimeout(handler, delay);
```

jrgarciadev (approved):
Thank you! @Gaic4o

packages/hooks/use-scroll-position/src/index.ts
```diff
@@ -71,7 +71,12 @@ export const useScrollPosition = (props: UseScrollPositionOptions): ScrollValue
...
     target.addEventListener("scroll", handleScroll);
 
-    return () => target.removeEventListener("scroll", handleScroll);
+    return () => {
+      target.removeEventListener("scroll", handleScroll);
+      if (throttleTimeout.current) {
+        clearTimeout(throttleTimeout.current);
+      }
+    };
   }, [elementRef?.current, delay, isEnabled]);
```
jrgarciadev, line 80:
```suggestion
  }, [elementRef?.current, delay, handler, isEnabled]);
```

PR #3064, opened 2024-05-23
fix(hooks): resolve type error in onKeyDown event handler

Gaic4o:
## 📝 Description

- Fixed a type error in the onKeyDown event handler for the menu component.

- The current implementation uses `@ts-ignore` to suppress type errors in the onKeyDown event handler. This may result in reduced type safety.

## 🚀 New behavior

- Replaced `@ts-ignore` with the appropriate type annotation for the onKeyDown event handler. The handler now ensures type safety by checking if listProps.onKeyDown is defined before calling it.

- No


wingkwong (changes requested):
please add the changeset. see [here](https://github.com/changesets/changesets/blob/main/docs/adding-a-changeset.md)

Gaic4o:
> please add the changeset. see [here](https://github.com/changesets/changesets/blob/main/docs/adding-a-changeset.md)

@wingkwong  I have modified it based on the information you provided. thank you

PR #3076, opened 2024-05-25
fix(date-range-picker): fixed input overflow

### LINKED ISSUE #3037: [BUG] - DATE_RANGE_PICKER_OVERFLOW
### NextUI Version

2.3.6

### Describe the bug

When the component is on mobile dimension it overflow to the sides. 

https://nextui.org/docs/components/date-range-picker

1. Go to https://nextui.org/docs/components/date-range-picker
2. Put your screen in mobile dimension (320x806) 100%
3. Resize the window to see the overflow

### Expected behavior

As a user or developer i expect, if there's no place for the full date i expect to see something like
7/05/2024, 12:4...

Cut the date text when there's no space for show the date. I a user i will know there's no much space to show the full date so i will click the button. Also helps a lot because the side button can't be pressed easily when this bug occurs.

### Screenshots or Videos

https://github.com/nextui-org/nextui/assets/58609880/6a13e513-ea66-417e-9dda-3bddac6a5195

### Operating System Version

macOS

### Browser

Chrome

ShrinidhiUpadhyaya:
Closes #3037 

## 📝 Description
When the component was on mobile screen, its contents were overflowing.

image

image

## 🚀 New behavior

image

image

## 📝 Additional Information
If the PR gets accepted please use my GitHub email-id (shrinidhiupadhyaya1195@gmail.com) instead of my other email-id for the Co-authored-by: message.

## Summary by CodeRabbit

- **New Features**
  - Added an optional `classNames` property for enhanced customization of date range picker components.

- **Style**
  - Improved layout stability in the date range picker by adding `flexShrink: 0`.
  - Enhanced date input component by applying `overflow-hidden` to prevent content overflow.

- **Bug Fixes**
  - Fixed input overflow issue in the date range picker component.


packages/components/date-picker/src/use-date-range-picker.ts
```diff
@@ -290,6 +290,7 @@ export function useDateRangePicker<T extends DateValue>({
       style: {
         ...props.style,
         maxWidth: "fit-content",
+        flexShrink: 0,
```
wingkwong, line 293: do we have to set it here instead of `packages/core/theme/src/components/date-input.ts`?

wingkwong (commented):
In this case, we can't scroll to see the value.

image

wingkwong (commented):
since you've modified code in packages, please include a changeset.

packages/components/date-picker/src/use-date-range-picker.ts
ShrinidhiUpadhyaya: I have set it here as there was no slot just for `start-input`.

ShrinidhiUpadhyaya: Regarding the scrolling, could we wrap the `start-input` and `end-input` inside a `div`? So that we could achieve something like this

image

wingkwong: > I have set it here as there was no slot just for start-input.

just a few lines below

```tsx
className: dateInputSlots.input({
  class: clsx(classNames?.input, props?.className),
}),
```

> Regarding the scrolling, could we wrap the start-input and end-input inside a div? So that we could achieve something like this

I think it's fine. We should only scroll the input. The scrollbar can look better tho.

ShrinidhiUpadhyaya: > 
> just a few lines below
> 
> ```tsx
> className: dateInputSlots.input({
>   class: clsx(classNames?.input, props?.className),
> }),
> ```

I guess we might not need `flexShrink: 0` as we are adding a scrollbar.


> 
> I think it's fine. We should only scroll the input.

If we dont add a `div` wrapper then the complete `input-wrapper` will be scrolled and by default the calendar button is not visible, we need to scroll to press the calendar button.

image

image

> 
> The scrollbar can look better tho.

I have used the default scrollbar like the components `Table` or  `ScrollShadow`.

Is their a custom scrollbar already present or do you want to me implement one?

Apart from the default scrollbar, the only scrollbar i found being used in the project is the below one? Do you want to me make use of this?
image

wingkwong: use the default scrollbar first. we may refactor later.

ShrinidhiUpadhyaya: I guess we might need to increase the height of the component as just applying scrollbar hides the input as shown below:-
image

Do you want me to increase the height? If yes could you tell me the values to which i need to increase?
Currently the `innerWrapper` height is set `h-6` and the size of the `inputWrapper` is `h-8` `h-10` `h-12` for size `sm` `md` & `lg` respectively.

wingkwong: i think not. the height shouldn't be changed.

ShrinidhiUpadhyaya: If the height is not changed then the text will not be visible.

wingkwong: the scrollbar shouldn't look like this. may double check with other pages. I rmb it was quite thin.

PR #3169, opened 2024-06-03
feat(hooks): use-theme hook

wingkwong:
Closes # 

## 📝 Description

- to replace the outdated `use-dark-mode`

## 🚀 New behavior

`useTheme(ThemeProps.LIGHT)`: 
image

`useTheme(ThemeProps.DARK)`: 
image

`useTheme(ThemeProps.SYSTEM)`: 

image

## Summary by CodeRabbit

- **New Features**
  - Introduced the `useTheme` hook for managing and switching between light and dark themes.
  - Enhanced theme management capabilities for improved user experience.

- **Documentation**
  - Updated documentation to reflect the transition from the `use-dark-mode` hook to the `@nextui-org/use-theme` hook.
  - Clarified installation instructions and example usage for the new hook, including updated theme switching methods.
  - Added a note indicating that the "dark mode" documentation has been recently updated.


packages/hooks/use-theme/src/index.ts
```diff
@@ -0,0 +1,49 @@
...
+        localStorage.setItem(ThemeProps.KEY, newTheme);
+        document.documentElement.classList.remove(theme);
+        document.documentElement.classList.add(newTheme);
+        setThemeState(newTheme);
+      } catch {}
+    },
+    [theme],
+  );
+
+  +useEffect(() => setTheme(theme), [theme, setTheme]);
```
jrgarciadev, line 46: @wingkwong why do we need a `+` symbol before the `useEffect` ?

wingkwong: Typo. It's not necessary.

AnYiEE:
Hello, if I understand the code correctly, the behavior of this `use-theme` hook is not consistent with `[use-dark-mode](https://github.com/donavon/use-dark-mode)` and [next-themes](https://github.com/pacocoursey/next-themes).

In short, it only has two states: "dark" and "light", lacking the "system" state.

Specifically, the main differences lie in dynamic response to system theme changes and calculation of initial theme. The current hook does not respond to system theme changes, meaning we cannot achieve the effect of "responding to system theme changes if the currently selected theme is `ThemeProps.SYSTEM`; otherwise forcibly applying `passed-in value (defaultTheme)`/dark/light themes" like next-themes does. If no passed-in value, the default theme for the current hook ends up being `ThemeProps.LIGHT`, rather than `ThemeProps.SYSTEM`.

You can experience the effects I mentioned on [my website](https://izakaya.cc). Before making a selection, the third item (system theme) in the theme menu at the navber right should be selected. Try changing the system theme and observe the changes. After selecting the first or second item in the theme menu, change the system theme again and observe the changes.

If there is any misunderstanding on my part, please point it out. Thank you.

wingkwong:
@AnYiEE 

> Hello, if I understand the code correctly, the behavior of this use-theme hook is not consistent with [use-dark-mode](https://github.com/donavon/use-dark-mode) and [next-themes](https://github.com/pacocoursey/next-themes).

Yes. The behaviours could be different because we are not trying to replace them but to customise more for our lib. We do take them as a reference though.

> In short, it only has two states: "dark" and "light", lacking the "system" state.

I'm not sure what your "system" state mean. From our definition, the system theme includes light or dark only. You can see how our components look like in light / dark mode [storybook](https://storybook.nextui.org/?path=/story/components-select--default&globals=locale:bg-BG).

In real life case, users may customise their own themes. As mentioned in the README doc, apart from system themes (dark & light), you can set other themes (e.g. purple-dark) using `setTheme`. However, this has to be defined in `tailwind.config.js` (see [example](https://nextui.org/docs/customization/create-theme)). Therefore, your statement `it only has two states: "dark" and "light"` is not correct.

AnYiEE:
> to customise more for our lib.

If I understand correctly, the [doc](https://nextui.org/docs/customization/dark-mode#using-use-dark-mode-hook) will also be modified. The doc has a wide impact, and the behavior of this hook is different from the previously recommended third-party implementations, so it may not be a backward-compatible change?

> I'm not sure what your "system" state mean.

I know that users can define a custom theme named "purple-dark," but what I mean by the system theme is not a theme named "system" similar to "purple-dark," but rather an "unselected" state.

> From our definition, the system theme includes light or dark only.

The current default return value of the hook is `ThemeProps.LIGHT`, which works fine when the system's color mode is light or when a custom theme named "purple-dark" has been defined by the user.

But if the system's color mode is dark, and `defaultTheme` is not passed, nor "purple-dark", then the default return value of the hook is still `ThemeProps.LIGHT`, that's right?

The `setTheme` method of this hook can only select existing, explicit themes (such as: dark, light, purple-dark, etc.), instead of adopting the system's color mode when no theme is specified as previously behavior by third-party implementations.

Thank you for taking the time to reply to me. If my understanding is incorrect, please point it out again.

wingkwong:
> If I understand correctly, the [doc](https://nextui.org/docs/customization/dark-mode#using-use-dark-mode-hook) will also be modified. The doc has a wide impact, and the behavior of this hook is different from the previously recommended third-party implementations, so it may not be a backward-compatible change?

`use-dark-mode` hasn't been maintained for a while and we provide a new alternative. I think both hook are designed for different purposes (our hook is more for our library). You need to do some changes if you migrate to our new hook. That's why we schedule this PR to v2.5.0.

> But if the system's color mode is dark, and defaultTheme is not passed, nor "purple-dark", then the default return value of the hook is still ThemeProps.LIGHT, that's right?

Correct. By default, it is using light theme.

> The setTheme method of this hook can only select existing, explicit themes (such as: dark, light, purple-dark, etc.), instead of adopting the system's color mode when no theme is specified as previously behavior by third-party implementations.

Correct. In NextUI, the styles of the components are designed to be customised by tailwindcss or tailwind-variants, not by other third-part implementations.

AnYiEE:
> By default, it is using light theme.

This is exactly the point that puzzles me. Why does it default to `ThemeProps.LIGHT` even if no default value is provided and no custom theme is defined, instead of the color mode of the user current operating system?

wingkwong:
because this is the base theme we design for that.

> the color mode of the user current operating system

can you share more how it affects the components? you may ping me at [discord](https://discordapp.com/users/495466997304262658) for discussion.

AnYiEE:
> can you share more how it affects the components?

I just set up a minimal new environment locally using Vite, with only the following key lines of code:

```ts
import { Button } from "@nextui-org/button" // v2.4.6
import { useTheme } from './use-theme/dist' // From the PR built
import { useEffect } from 'react'

function App() {
  const { theme, setTheme } = useTheme()

  useEffect(() => {
    console.log(theme)
  }, [theme])

  return (
    <Button onPress={() => {
      setTheme(theme === 'light' ? 'dark' : 'light')
    }}>Button</Button>
  )
}
```

Start the development server, when the system color mode is dark (can also be simulated through the console), enter the page, you can see that the page background and the button are still in light. The class of HTML tags and the value in local storage are "light". At this time, you can only switch color modes by clicking on buttons.

> the base theme we design for that

This is a fact, but it does not mean that we cannot prioritize the system's color mode as the default.

Just like the NextUI official website, regardless of whether my system color mode is light or not, it always defaults to dark unless manually switched. Unlike the official website, standalone components used externally should not have this default preference.

That's what I think. Thank you again for taking the time to reply to me.

wingkwong:
Can you add me at discord for further discussion? I think I may need to dig into this part.

packages/hooks/use-theme/src/index.ts
```diff
@@ -0,0 +1,81 @@
...
+  | customTheme;
+
+/**
+ * React hook to switch between themes
+ *
+ * @param defaultTheme the default theme name (e.g. light, dark, purple-dark and etc)
+ * @returns An object containing the current theme and theme manipulation functions
+ */
+export function useTheme(defaultTheme?: Theme) {
+  const [theme, setThemeState] = useState<Theme>(() => {
```
ryo-manba, line 30: It would be more effective to initialize the state with either the value from localStorage or the `defaultTheme`. Additionally, it makes sense to create an `applyTheme` function to handle theme changes when they occur. You can refer to the following implementation.
https://github.com/pacocoursey/next-themes/blob/main/next-themes/src/index.tsx#L42-L49

```suggestion
export function useTheme(defaultTheme: Theme = ThemeProps.SYSTEM) {
  const [theme, setThemeState] = useState<Theme>(() => localStorage.getItem(ThemeProps.KEY) ?? defaultTheme);
```

packages/hooks/use-theme/src/index.ts
```diff
@@ -0,0 +1,81 @@
...
+    // if it is using system theme, check `prefers-color-scheme` value
+    // return light theme if not specified
+    if (defaultTheme === ThemeProps.SYSTEM) {
+      return window.matchMedia && window.matchMedia("(prefers-color-scheme: dark)").matches
+        ? ThemeProps.DARK
+        : ThemeProps.LIGHT;
+    }
+
+    // return default theme with light theme as default one
+    return defaultTheme ?? ThemeProps.LIGHT;
```
ryo-manba, line 49: It would be more convenient for users if the system theme is applied when the theme is not found in localStorage or no defaultTheme is provided. Next.js docs follows a similar approach.
https://nextjs.org/docs

packages/hooks/use-theme/package.json
```diff
@@ -0,0 +1,52 @@
...
+  "clean-package": "../../../clean-package.config.json",
+  "tsup": {
+    "clean": true,
+    "target": "es2019",
+    "format": [
+      "cjs",
+      "esm"
+    ]
+  }
+}
```
ryo-manba, line 52:
```suggestion
}

```

ryo-manba (changes requested):
Nice, I’ve added some comments!

packages/hooks/use-theme/src/index.ts
AnYiEE: If modified as you said, when `defaultTheme` is not passed, `defaultTheme` will default to `system`. At this point, the result is returned in line 43, and this line only needs to `return defaultTheme`, no additional `??` needed.

packages/hooks/use-theme/src/index.ts
AnYiEE: I think there is no need to modify line 30, the version before modification is safer and more complete, just need to change line 29.

packages/hooks/use-theme/src/index.ts
```diff
@@ -0,0 +1,87 @@
...
+              ? ThemeProps.DARK
+              : ThemeProps.LIGHT
+            : newTheme;
+
+        localStorage.setItem(ThemeProps.KEY, targetTheme);
+
+        document.documentElement.classList.remove(theme);
+        document.documentElement.classList.add(targetTheme);
+        setThemeState(newTheme);
+      } catch {}
```
ryo-manba, line 62: Why is the `try-catch` block necessary?

packages/hooks/use-theme/src/index.ts
```diff
@@ -0,0 +1,87 @@
...
+    },
+    [theme, MEDIA],
```
ryo-manba, line 64: Since `MEDIA` is constant, there's no need to include in the dependency array.

packages/hooks/use-theme/src/index.ts
AnYiEE: I think it is because `localStorage` and its methods are not always necessarily available.

packages/hooks/use-theme/src/index.ts
```diff
@@ -0,0 +1,85 @@
...
+  const setTheme = useCallback(
+    (newTheme: Theme) => {
+      const targetTheme =
+        newTheme === ThemeProps.SYSTEM
+          ? window.matchMedia?.(MEDIA).matches
+            ? ThemeProps.DARK
+            : ThemeProps.LIGHT
+          : newTheme;
+
+      localStorage.setItem(ThemeProps.KEY, targetTheme);
```
ryo-manba, line 56: This only allows choosing between light or dark themes. It would be nice if the "system" option could also be saved in localStorage, like in the [Next.js docs](https://nextjs.org/docs).
image

ryo-manba (approved):
LGTM! Nice job.

PR #3174, opened 2024-06-04
fix(input): input display with hidden type

### LINKED ISSUE #3170: [BUG] - <Input type='hidden' /> does not hide the entire component
### NextUI Version

2.4.0

### Describe the bug

Adding the `type='hidden'` attribute on the `Input` component does not hide it completely - it only hides the underlying HTML `<input>`. 

_No response_

`<Input type='hidden' />`

### Expected behavior

Ideally the entire element should be hidden while still being able to retrieve the value in a form submission, etc.

### Screenshots or Videos

The wrapper still being visible:
image

### Operating System Version

Windows

### Browser

Edge

wingkwong:
Closes #3170

## 📝 Description

- handle the display of the input for hidden type

a readonly input is shown

## 🚀 New behavior

input is not shown if `type` is `hidden`

No

## Summary by CodeRabbit

- **Bug Fixes**
  - Resolved an issue where `Input` components with a hidden type were incorrectly displayed.

- **Tests**
  - Added a new test case to verify the correct visibility behavior of `Input` components with a hidden type.

- **Refactor**
  - Updated base CSS classes for `Input` components to conditionally include a hidden class based on data attributes.

### DIFF
--- a/packages/components/input/src/use-input.ts
+++ b/packages/components/input/src/use-input.ts
@@ -253,7 +253,6 @@ export function useInput<T extends HTMLInputElement | HTMLTextAreaElement = HTML
         labelPlacement,
         isClearable,
         disableAnimation,
-        isHiddenType,
       }),
     [
       objectToDeps(variantProps),
--- a/packages/components/input/src/use-input.ts
+++ b/packages/components/input/src/use-input.ts
@@ -312,6 +312,7 @@ export function useInput<T extends HTMLInputElement | HTMLTextAreaElement = HTML
       isFilledWithin,
       hasPlaceholder,
       focusWithinProps,
+      isHiddenType,
       originalProps.isReadOnly,
       originalProps.isRequired,
       originalProps.isDisabled,
--- a/packages/components/input/src/use-input.ts
+++ b/packages/components/input/src/use-input.ts
@@ -148,9 +148,7 @@ export function useInput<T extends HTMLInputElement | HTMLTextAreaElement = HTML
   const isHiddenType = type === "hidden";
   const isMultiline = originalProps.isMultiline;
 
-  const baseStyles = clsx(classNames?.base, className, isFilled ? "is-filled" : "", {
-    hidden: isHiddenType,
-  });
+  const baseStyles = clsx(classNames?.base, className, isFilled ? "is-filled" : "");
 
   const handleClear = useCallback(() => {
     setInputValue("");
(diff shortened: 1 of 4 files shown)


packages/components/input/src/use-input.ts
```diff
@@ -288,6 +292,7 @@ export function useInput<T extends HTMLInputElement | HTMLTextAreaElement = HTML
         "data-has-value": dataAttr(!isPlaceholderShown),
         ...focusWithinProps,
         ...props,
+        hidden: isHiddenType,
```
jrgarciadev, line 295:
```suggestion
        "data-hidden": dataAttr(isHiddenType),
```

packages/components/input/src/use-input.ts
```diff
@@ -145,9 +145,13 @@ export function useInput<T extends HTMLInputElement | HTMLTextAreaElement = HTML
   const isFilledByDefault = ["date", "time", "month", "week", "range"].includes(type!);
   const isFilled = !isEmpty(inputValue) || isFilledByDefault;
   const isFilledWithin = isFilled || isFocusWithin;
-  const baseStyles = clsx(classNames?.base, className, isFilled ? "is-filled" : "");
+  const isHiddenType = type === "hidden";
   const isMultiline = originalProps.isMultiline;
 
+  const baseStyles = clsx(classNames?.base, className, isFilled ? "is-filled" : "", {
+    hidden: isHiddenType,
```
jrgarciadev, line 152: Move this to the input styles in the `theme` package, use the `data-hidden` attr

jrgarciadev: The reason is because tailwindcss only compiles the styles inside the `theme` package not inside the `react` or `input` one

packages/core/theme/src/components/input.ts
```diff
@@ -236,6 +236,11 @@ const input = tv({
         clearButton: ["transition-opacity", "motion-reduce:transition-none"],
       },
     },
+    isHiddenType: {
+      true: {
+        base: "hidden",
+      },
+    },
```
jrgarciadev, line 243: @wingkwong adding this as a variant means that the Input component will receive `isHiddenType` as prop, instead of adding a variant, please use the `data-hidden` prop.

e.g.

`base: "data-[hidden=true]:hidden"`

jrgarciadev (approved):

Issue #3225, opened 2024-06-09
[BUG] - Autocomplete's keyboard shortcuts behave weirdly


takuyahara:
### NextUI Version

0.0.0-canary-20240607005641

### Describe the bug

Autocomplete's keyboard shortcuts behave weirdly as follows. They don't happen in examples on  [document](https://nextui.org/docs/components/autocomplete).
1. When hitting tab key next autocomplete's textbox is not focused and popover opens but none of keyboard shortcut (ex. arrow up/down, enter, esc, etc...) works.
2. When autocomplete is the last focusable component, hitting tab key results popover keeps opened.

But when textbox is clicked, hitting tab key focuses to next autocomplete's textbox. This is an expected behavior.

### Your Example Website or App

https://stackblitz.com/~/github.com/takuyahara/nextui-autocomplete-issue

### Steps to Reproduce the Bug or Issue

1. Go to the above-mentioned stackblitz page.
2. Open page in new tab to see an isolated version.
3. Hit tab key and arrow up/down keys.

### Expected behavior

1. Hitting tab key focuses to next autocomplete's textbox with popover automatically opened and keyboard shortcut (ex. arrow up/down, enter, esc, etc...) works.
2. When autocomplete is the last focusable component, hitting tab key closes popover.

### Screenshots or Videos

https://github.com/nextui-org/nextui/assets/46240835/7434f41d-b1fe-4934-9a39-d9f2dbf4d18d

### Operating System Version

macOS

### Browser

Brave v1.66.118 (Chromium v125.0.6422.147)

takuyahara:
Another issues found.

### Describe the bug

1. Clicking autocomplete doesn't focus on textbox on first click so input is not accepted. It has to be clicked twice.
2. When typing in textbox and delete key is hit, focus is automatically blurred.
3. When popover is opened and outside of the browser is clicked, popover closes. And when inside of the page is clicked popover reopens without keyboard support.

### Expected behavior

1. Clicking autocomplete focuses on textbox on first click.
2. Hitting delete key in textbox doesn't blur.
3. When inside of the page is clicked again, textbox is focused and popover reopens without keyboard support.

### Screenshots or Videos

https://github.com/nextui-org/nextui/assets/46240835/7db1b539-ccf8-4f2b-84eb-f9d01c495b03

wingkwong:
This is what I see. Just use tab and arrow down key.

[...]

Issue #3203, opened 2024-06-06
[BUG] - Autocomplete don't focus on input


iPagar:
### NextUI Version

2.4.1

### Describe the bug

When I click on autocomplete it should focus on input

### Your Example Website or App

https://codesandbox.io/p/sandbox/objective-nash-lff3pt

### Steps to Reproduce the Bug or Issue

1. Copy code from https://nextui.org/docs/components/autocomplete#usage
2. Click on autocomplete
3. Input is not focused 

### Expected behavior

Input has to be focused as it shown at https://nextui.org/docs/components/autocomplete#usage

### Operating System Version

macOS

### Browser

Chrome

wingkwong:
seems it looks fine in local storybook but not in app.

MarcoRodasW:
Same issue for me, i have autocompletes everywhere in the app 🥲

MarcoRodasW:
@wingkwong Any temporary solution until it is fixed?

daniel-esserweb:
This still not fixed in v2.4.2

MarcoRodasW:
@daniel-esserweb Noup, I switched to version 2.2.9 until they fix it

daniel-esserweb:
> @daniel-esserweb Noup, I switched to version 2.2.9 until they fix it

I dont understand why, in the documentation is working since 2.4.0 i think, but in the application hasn't worked yet

MarcoRodasW:
@daniel-esserweb Me neither, I tried different versions and only 2.29 and below worked for me.

cnfw:
I do this temporary workaround to get around the autofocus issue, but it would only work for mouse clicks - keyboard users still need to be considered.

```tsx
<Autocomplete
    onClick={(event) =>
        // @ts-expect-error the input doesn't automatically focus
        event.target.focus()
    }

>
// ...
</Autocomplete>
```

kuri-sun:
@iPagar Thanks for the report! 
Speaking of which, I also found a related bug for Autocomplete component so I will put it down here as well. 
When an user try to remove chars from the input, the input will freeze 😞  (In the video, it's really hard to see but I was pressing the key multiple times.)
These issues seem like happening only on local env(dev env) & sandbox env. 
I will also try to take a look at these! Thanks again for reporting this issue! 👍 

https://github.com/nextui-org/nextui/assets/62743644/04672809-c1e2-4ab6-8bd4-8d950f9872a7

cedoromal:
[...]

Issue #3124, opened 2024-05-29
[BUG] - Autocomplete modal open action makes the input to loose focus


c0d3rm0n:
### NextUI Version

2.4.0

### Describe the bug

Autocomplete inputs loose focus when its modal (options list) opens. 
This happens by click or using Tab... Even if I click again in the input area to get focus, as soon as I type something that makes my lookup return [ ] as a result of options, the modal closes, and when I delete part of the string so the options returned include my search, when the modal re-opens, the input looses focus once more.
I have this problem since v2.3. It works just fine in v2.2.10.

### Steps to Reproduce the Bug or Issue

1. Click in NextUI Autocomplete input
2. The input will not have focus
3. Click again to get focus and type something (let's use animals as example: bird)
4. It returns 1 option
5. Type another 'd' to make it return [ ] from its lookup
6. The modal closes
7. Delete the last 'd'
8. The options returned will be ['bird'], This makes the modal to re-open and the input looses focus

### Expected behavior

The user should be able to click and type in Autocomplete while maintaining focus on the input.

### Operating System Version

Ubuntu 22.04.4 LTS

### Browser

Firefox

c0d3rm0n:
Previous issue: [2849](https://github.com/nextui-org/nextui/issues/2849).

wingkwong:
Can you provide a minimal reproducible example or a video for demonstration? also for "modal", are you referring the nextui modal or the autocomplete combobox?

c0d3rm0n:
> Can you provide a minimal reproducible example or a video for demonstration? also for "modal", are you referring the nextui modal or the autocomplete combobox?

I'm talking about the Autocomplete box where the options are rendered.

wingkwong:
still please provide some code and a video for demonstration. I was lost in step 6.

MinChanSike:
Same issue,
**Steps to Reproduce the Bug or Issue**

1. Click in NextUI Autocomplete input
2. The input will not have focus
3. Click on input to get focus
4. Type "Cat" then press `Arrow Down` to select the item
5. Press `Back`, so only "Ca" is remain on input then model become active, loose focus

Please take a look in screen capture 
https://github.com/nextui-org/nextui/assets/2868221/0c3d7a74-ca40-4713-9979-0f447d1e5eaa

wingkwong:
[...]

PR #3311, opened 2024-06-24
fix(popover): dialog focus behaviour 

wingkwong:
- Closes: #3225
- Closes: #3124
- Closes: #3203

## 📝 Description

The focus on autocomplete works only in the following cases:
- production environment
- dev environment without React strictMode

This PR is to avoid autocomplete to focus on dialog before focusing on input so that the focus logic in autocomplete hook also works with React strictMode.

- open autocomplete
- focus on dialog (logic from react-aria side)
- focus on input (logic from nextui autocomplete hook)

## 🚀 New behavior

- open autocomplete
- focus on input (logic from nextui autocomplete hook)

Demo (with strictMode: true)

[pr3311-demo.webm](https://github.com/nextui-org/nextui/assets/35857179/6ecf39b4-f441-4835-aabf-c8a6c8a45797)

[pr3311-demo-2.webm](https://github.com/nextui-org/nextui/assets/35857179/fb93040f-7f66-4b42-8bf8-9352cd8c6ea0)

No

## Summary by CodeRabbit

- **New Features**
	- Added `disableDialogFocus` option to the autocomplete and popover components, allowing users to keep focus on the input when the popover is open.
- **Bug Fixes**
	- Resolved focus-related issues in popover and autocomplete components (#3225, #3124, #3203).

### DIFF
--- a/.changeset/plenty-suns-juggle.md
+++ b/.changeset/plenty-suns-juggle.md
@@ -1,0 +1,6 @@
+---
+"@nextui-org/autocomplete": patch
+"@nextui-org/popover": patch
+---
+
+add `disableDialogFocus` to free-solo-popover (#3225, #3124, #3203)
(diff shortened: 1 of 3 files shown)


wingkwong:
The tests related to focus have been there. The issue doesn't exist in production since it's related to strict mode (when you set strict mode, it only applies to dev). In our local storybook / doc, we haven't set strict mode so it looks fine. If tests with strictMode on are required, there is [another PR](https://github.com/nextui-org/nextui/pull/3352) doing that.

PR #3315, opened 2024-06-24
fix(navbar): shouldHideOnScroll fixed

pratyush3124:
Closes # https://github.com/nextui-org/nextui/issues/3139

## 📝 Description

The `useScrollPosition` hook responsible for reporting new scroll position when scrolled was faulty.

`throttleTimeout.current` not being null is causing `handler()` to not execute at all.

I am clearing the timeout and calling `handler()` after delay in such case.

- nav not hiding on scroll

## 🚀 New behavior

- nav hides on scroll

https://github.com/nextui-org/nextui/assets/44525862/eb7af9ed-1790-42f0-b026-2d563ecc4202

No

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected the `useScrollPosition` hook to properly clear the throttle timeout when setting a new timeout.

### DIFF
--- a/.changeset/happy-cameras-whisper.md
+++ b/.changeset/happy-cameras-whisper.md
@@ -1,0 +1,5 @@
+---
+"@nextui-org/use-scroll-position": minor
+---
+
+clearing throttleTimeout when it is supposed to be null in useScrollPosition
--- a/.changeset/happy-cameras-whisper.md
+++ b/.changeset/happy-cameras-whisper.md
@@ -1,5 +1,5 @@
 ---
-"@nextui-org/use-scroll-position": minor
+"@nextui-org/use-scroll-position": patch
 ---
 
-clearing throttleTimeout when it is supposed to be null in useScrollPosition
+clearing throttleTimeout when it is supposed to be null in useScrollPosition (#3139)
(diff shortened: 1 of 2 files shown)


.changeset/happy-cameras-whisper.md
```diff
@@ -0,0 +1,5 @@
+---
+"@nextui-org/use-scroll-position": minor
```
wingkwong, line 2: please change to `patch` instead.

```
"@nextui-org/use-scroll-position": patch
```

.changeset/happy-cameras-whisper.md
```diff
@@ -0,0 +1,5 @@
...
+---
+
+clearing throttleTimeout when it is supposed to be null in useScrollPosition
```
wingkwong, line 5: please include the issue number as well

```
clearing throttleTimeout when it is supposed to be null in useScrollPosition (#3139)
```

wingkwong (changes requested):
can be simplified as below

```tsx
const handleScroll = () => {
  if (delay) {
    if (throttleTimeout.current) {
      clearTimeout(throttleTimeout.current);
    }

    throttleTimeout.current = setTimeout(handler, delay);
  } else {
    handler();
  }
};
```

wingkwong (approved):

jrgarciadev (approved):

Issue #3353, opened 2024-06-28
[BUG] - Autocomplete component return the Label instead the Value


GVALFER:
### NextUI Version

Latest

### Describe the bug

On <Autocomplete> component, the value received is the Label instead the value.
I guess its a BUG or not? It should log the value, not the label

### Steps to Reproduce the Bug or Issue

```javascript
const animals = [
    { label: "Big Cat", value: "cat"},
    { label: "Big Dog", value: "dog"}
];

const handleSubmit = (values) => {
    const name = values.get("animal");
    console.log(name); // Big Cat
}

<form action={handleSubmit}>     
   <Autocomplete
      name="animal"
      label="Favorite Animal" 
      defaultItems={animals}
      defaultSelectedKey="cat"
  >
      {(item) => <AutocompleteItem key={item.value}>{item.label}</AutocompleteItem>}
  </Autocomplete>
  
  <Button type="submit">Submit</Button>
</form>
```
### Expected behavior

As normal, it should be returned the value and not the label

### Operating System Version

macOS

### Browser

Edge

devconstrib:
Yo tengo exactamente el mismo problema con los Autocomplete creo que no se presta intención porque esta en español :(

devconstrib:
https://github.com/nextui-org/nextui/issues/3343#issue-2378476447
 Te dejo mi caso es similar el unico detalle esque yo saco los datos de una api al final es casi lo mismo

MohammadHarish52:
"defaultSelectedkey" might have the value of "cat" hence it is returning the label to solve this you might work on the handle submit handler fn `const animals = [
    { label: "Big Cat", value: "cat"},
    { label: "Big Dog", value: "dog"}
];

const handleSubmit = (values) => {
    const selectedLabel = values.get("animal");
    const selectedAnimal = animals.find(animal => animal.label === selectedLabel);
    const name = selectedAnimal ? selectedAnimal.value : null;
    console.log(name); // Should log "cat" if "Big Cat" is selected
}

<form action={handleSubmit}>     
   <Autocomplete
      name="animal"
      label="Favorite Animal" 
      defaultItems={animals}
      defaultSelectedKey="cat"
  >
      {(item) => <AutocompleteItem key={item.value}>{item.label}</AutocompleteItem>}
  </Autocomplete>
  
  <Button type="submit">Submit</Button>
</form>
`

this might help

MohammadHarish52:
[...]

PR #3375, opened 2024-06-30
fix(autocomplete): return value instead of label in form submission

wingkwong:
- Closes: #3353
- Closes: #3343

## 📝 Description

> Add a brief description

## 🚀 New behavior

### DIFF
--- a/packages/components/autocomplete/src/use-autocomplete.ts
+++ b/packages/components/autocomplete/src/use-autocomplete.ts
@@ -415,8 +415,11 @@ export function useAutocomplete<T extends object>(originalProps: UseAutocomplete
       }),
     } as ButtonProps);
 
-  const getInputProps = () =>
-    ({
+  const getInputProps = () => {
+    // TODO(wkw): add comments
+    delete inputProps["name"];
+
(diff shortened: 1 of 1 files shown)


ehsanrezaeee:
is there any progress on this issue? please merge so i can use the autocomplete component

ehsanrezaeee:
Thanks. really appreciate it 🙌👌

ryo-manba:
@wingkwong @jrgarciadev 
I don't think this behavior is a bug.
Autocomplete works as an input, not a select. So, it makes sense to send the label (the displayed text) on form submission.

For example, if the key is a sent as the value, the [Custom value](https://nextui.org/docs/components/autocomplete#custom-value) feature will send the key instead of the user's input. This would not be useful.

ryo-manba:
As mentioned in [this comment](https://github.com/nextui-org/nextui/pull/3375#issuecomment-2483421983), autocomplete is not a select, so the label is submitted.
If you want to submit a value other the label, please use a [Select](https://nextui.org/docs/components/select) or handle it as a controlled component.

PR #3444, opened 2024-07-11
fix(use-image): cached image flashing

### LINKED ISSUE #3437: [Avatar] Fix flashing when image is already cached
### NextUI Version

2.4.2

### Describe the bug

Avatars flicker and animate when they should be rendered instantly when cached, Radix shows a fix here ->

https://github.com/radix-ui/primitives/pull/3008

_No response_

Using Avatar should render immediately if the image is cached with no loading animation or flickerAvatars flicker and animate when they should be rendered instantly when cached, Radix shows a fix here ->

https://github.com/radix-ui/primitives/pull/3008

### Expected behavior

Using Avatar should render immediately if the image is cached with no loading animation or flickerAvatars flicker and animate when they should be rendered instantly when cached, Radix shows a fix here ->

https://github.com/radix-ui/primitives/pull/3008

### Screenshots or Videos

_No response_

### Operating System Version

macOS

### Browser

Chrome

rkkautsar:
Closes #3437  

## 📝 Description

As mentioned in #3437, avatars may flicker even if its already cached. The fix was to check the image loading status immediately after creating the image to see if it's instantly loaded (when the image is cached), and use it to initialize the state. This will affect Avatar, AvatarGroup, and Image components.

Also added test for use-image too since it doesn't exist before.

| Before | After |
|--------|--------|
| nextui-flashing-before | nextui-flashing-after |

Image flickers due to use-image initializes with `pending` and only updated the status in `useLayoutEffect`.

## 🚀 New behavior

Initialize image loading status with loaded if possible.

## Summary by CodeRabbit

- **Bug Fixes**
	- Resolved issue where cached images were flashing due to improper initialization of the image's loaded state.

- **Tests**
	- Added comprehensive tests for the `useImage` hook covering scenarios like missing source, loading, errors, and cached images.

- **Chores**
	- Updated `use-image` package devDependencies to include `@nextui-org/test-utils`.

- **Refactor**
	- Refined image loading logic in the `useImage` function to improve the image initialization and status determination process.

### DIFF
--- a/.changeset/pink-beans-sit.md
+++ b/.changeset/pink-beans-sit.md
@@ -1,0 +1,6 @@
+---
+"@nextui-org/use-image": patch
+"@nextui-org/test-utils": patch
+---
+
+fix cached image flashing due to use-image always returning pending initially. The fix was to check if the image is loaded instantly through HTMLImageElement.complete attribute and use that to initialize the state.
--- a/packages/hooks/use-image/__tests__/use-image.test.tsx
+++ b/packages/hooks/use-image/__tests__/use-image.test.tsx
@@ -1,0 +1,44 @@
+import {renderHook} from "@testing-library/react-hooks";
+import {mocks} from "@nextui-org/test-utils";
+
+import {useImage} from "../src";
+
+describe("use-image hook", () => {
+  let mockImage: {restore: any; simulate: (value: "loaded" | "error") => void};
+
+  beforeEach(() => {
+    mockImage = mocks.image();
+  });
+  afterEach(() => {
+    mockImage.restore();
+  });
+
+  it("can handle missing src", () => {
+    const rendered = renderHook(() => useImage({}));
+
+    expect(rendered.result.current).toEqual("pending");
+  });
+
+  it("can handle loading image", async () => {
+    const rendered = renderHook(() => useImage({src: "/test.png"}));
+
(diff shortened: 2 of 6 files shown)


wingkwong (commented):
I tested it on your PR changes following your video. I still see a few images flashing while most look same as your video. Have you encountered this issue?

rkkautsar:
@wingkwong yes, I think it had to do with the actual image caching behaviour of each browser. I used chrome, and from what I can see the caching seems to be size limited, if I were to guess it's because there is a few level of caching, the one that is instant is memory-cached, while the other one that still flashes is disk-cached.

If I show a big image (like in `<Image/>` component) it can only cache one image at a time, so switching between different image will flash. But switching components that is showing the same image won't flash.

Conversely, for small image like in avatars (and avatar groups), it seems that most of the small images is cached in memory so it always never flash (unless you load a big image and come back, the big image will replace the memory cache I think)

wingkwong:
will revisit this PR after the one in Radix is merged. May check if those comments can be applied here or not.

jrgarciadev:
Hey @rkkautsar thanks for you great contribution, we had to create the PR internally to move things a bit faster https://github.com/nextui-org/nextui/pull/3987 but do not worry we will add you to the contributors list, thanks 🙏🏻

PR #3445, opened 2024-07-11
fix(card): removed the default z-10 of card header

### LINKED ISSUE #3198: [BUG] - CardHeader z-index =10 by default
### NextUI Version

2.4.1

### Describe the bug

It is weird that CardHeader is the only component to my knowledge to have a z-index of 10.
CardBody doesn't and behaves as expected, but not CardHeader as you can see in this screenshot :

image

I cloned nextui and think I found it there : 
image

_No response_

1. Create a sticky header with z-index = 2
2. Create a scrollable column that is supposed to go under the header
3. Add cards with a CardHeader and whatever you want in it to the column
4. Scroll and watch the CardHeader be visible on top of the sticky header

### Expected behavior

As a developer, I expect the CardHeader to behave like CardBody or CardFooter, and not having a different z-index

### Screenshots or Videos

_No response_

### Operating System Version

WSL

### Browser

Chrome

abhinandan-verma:
Closes #3198

## 🚀 New behavior

### DIFF
--- a/.changeset/grumpy-days-refuse.md
+++ b/.changeset/grumpy-days-refuse.md
@@ -1,0 +1,5 @@
+---
+"@nextui-org/theme": patch
+---
+
+removed the default z-10 of card header(#3198)
--- a/packages/core/theme/src/components/card.ts
+++ b/packages/core/theme/src/components/card.ts
@@ -35,7 +35,6 @@ const card = tv({
     header: [
       "flex",
       "p-3",
-      "z-10",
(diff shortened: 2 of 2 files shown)


Cajuteq (approved):

Cajuteq:
Hi @wingkwong,
I see you closed without merge, can we know why ?
CardFooter and CardBody don't set z index, why should CardHeader be any different ?

PR #3458, opened 2024-07-12
Docs: installation.mdx is incorrect

davimarcilio:
The `@nextui-org/theme` module has several `.mjs` files that are not being read by `tailwind.config.ts`, which causes components to render without styling.

Updated the content array in `tailwind.config.ts` to include `.mjs` files:

```js
'./node_modules/@nextui-org/theme/dist/**/*.{js,ts,jsx,tsx,mjs}'
```

## 📝 Description

This pull request updates the tailwind.config.ts file to include .mjs files from the @nextui-org/theme module. This change ensures that TailwindCSS reads the module's .mjs files, which resolves the issue of components rendering without styling.

Currently, the tailwind.config.ts file does not include .mjs files in the content array. As a result, TailwindCSS does not read the styles from the @nextui-org/theme module, causing components to render without the correct styling.

## 🚀 New behavior

With this update, the content array in tailwind.config.ts will include .mjs files. This change allows TailwindCSS to read the .mjs files from the @nextui-org/theme module, ensuring that components are styled correctly.

```js
module.exports = {
  content: [
    './node_modules/@nextui-org/theme/dist/**/*.{js,ts,jsx,tsx,mjs}',
    // other paths
  ],
  // other configurations
};
```

No.

## 📝 Additional Information

This change is backward compatible and does not affect existing functionality. It simply adds support for .mjs files in the TailwindCSS configuration, which is necessary for the proper styling of components from the @nextui-org/theme module.

## Summary by CodeRabbit

- **Documentation**
  - Updated the installation guide to include support for `.mjs` file extension in the theme module configuration.

### DIFF
--- a/apps/docs/content/docs/guide/installation.mdx
+++ b/apps/docs/content/docs/guide/installation.mdx
@@ -151,7 +151,7 @@ module.exports = {
   content: [
     // ...
     // make sure it's pointing to the ROOT node_module
-    "./node_modules/@nextui-org/theme/dist/**/*.{js,ts,jsx,tsx}",
+    "./node_modules/@nextui-org/theme/dist/**/*.{js,ts,jsx,tsx,mjs}",
   ],
   theme: {
     extend: {},


wingkwong (commented):
> The `@nextui-org/theme` module has several `.mjs` files that are not being read by `tailwind.config.ts`, which causes components to render without styling.

I don't think that is required as it should be already covered. Can you share two stackblitz links to  prove your statement:

- one showing the components to render without styling (without this PR change)
- one showing the components to render with styling (with this PR change)

PR #3490, opened 2024-07-17
fix(button): rescaling in disable animation

### LINKED ISSUE #3489: [BUG] - `disableAnimation` prop not working in Button component
### NextUI Version

@nextui-org/button@2.0.31

### Describe the bug

The button's `disableAnimation` prop should disable the rescaling animation on press. However, we currently need to add a `data-[pressed=true]:scale-100` class override as well in order to achieve that.

- I traced it to [this line](https://github.com/nextui-org/nextui/blob/canary/packages/core/theme/src/components/button.ts#L40) in the button component's source code, which seems to be getting applied even when we pass `disableAnimation`, likely due to [this change](https://github.com/nextui-org/nextui/pull/2929/files#diff-ae075221d59003e0b1c8add05fbb594132e2df8bd2d6b987f48aec5188166486L89)
- I believe that change accidentally broke the expected behavior of the `disableAnimation` prop, and would request that you re-disable the scaling behavior on press when `disableAnimation` is false.

_No response_

You can see it not working in [Storybook](https://storybook.nextui.org/?path=/story/components-button--disable-ripple&args=disableAnimation:!true). It seems like it removes the smooth transition but keeps the button resizing on press effect.

### Expected behavior

I would expect the button not to resize on press when `disableAnimation` is passed in.

### Screenshots or Videos

_No response_

### Operating System Version

MacOS

### Browser

Chrome

abhinandan-verma:
Closes #3489 

## 📝 Description

FIxed the button rescaling in disableAnimation

https://github.com/user-attachments/assets/1ad1ed52-1cb5-4740-9dec-418525b8217d

## 🚀 New behavior

No rescaling in disableAnimation.

## Summary by CodeRabbit

- **Bug Fixes**
  - Resolved an issue with button rescaling when animations are disabled, ensuring consistent button appearance.


wingkwong (changes requested):
this change isn't fixing anything. by default, there is already `data-[pressed=true]:scale-[0.97]` in `base` so this PR actually does nothing.

abhinandan-verma:
@wingkwong  without this change, rescaling is there even in disableAnimation

wingkwong:
@abhinandan-verma the issue is about disabling the rescaling. with / without this change, rescaling is still there.

abhinandan-verma:
@wingkwong  sir the effect is visible only if disableAnimation is true.

wingkwong:
@abhinandan-verma i don't think you understand the issue and my comment. I'll take over it.

PR #3568, opened 2024-07-28
fix(select): prevent select from closing instantly when clicking on label of select

### LINKED ISSUE #3558: [BUG] - Select opens and closes instantly when clicking on the label
### NextUI Version

~~2.4.6~~ 2.4.8

### Describe the bug

Just upgraded to the latest version and now my Selects are not behaving properly. When I click on the label of a Select it used to Open and allow user to select an option. Now it opens and instantly closes. I see it is also happening on your Docs website.

https://nextui.org/docs/components/select

1. Go to https://nextui.org/docs/components/select
2. Click on the label of any of the Selects
3. Notice how it opens and closes instantly

### Expected behavior

When you click on the Label it should remain open.

### Screenshots or Videos

https://github.com/user-attachments/assets/c4dac053-244c-4ea4-bbf7-4c8dc29cfa2e

### Operating System Version

MacOS

### Browser

Chrome

### LINKED ISSUE #3683: [Select Label] - Unexpected behavior when selecting selector tag
### NextUI Version

2.2.5

### Describe the bug

When an option is selected from the dropdown, attempting to select a different option by clicking the label causes an unexpected behavior in the list of options.

[label.webm](https://github.com/user-attachments/assets/1906e6e4-65ae-4abe-922a-a0dcecad80b9)

_No response_

1. Select an option from the list
2. click on the label (Not on the selected option), otherwise on the label
3. see unexpected behavior

### Expected behavior

It is expected that when I click on the label the options to select will appear again

### Screenshots or Videos

[label.webm](https://github.com/user-attachments/assets/8e140ad7-d21d-4af9-bed3-5e6628a2bf7e)

### Operating System Version

windows

### Browser

Chrome

sohan01fw:
Closes #3558
Closes #3683

📝 Description

Fixed select component unexpectedly close while click on label of select component

⛳️ Current behavior (updates)

Below recording shows the current behavior of component.

https://github.com/user-attachments/assets/7bc1411b-e456-41d3-92d7-1032ca0c6e45

🚀 New behavior

Below recording shows the fix behavior of component

https://github.com/user-attachments/assets/73bb0136-5905-4536-8fb0-d40a2f84bb92

💣 Is this a breaking change (Yes/No):No

📝 Additional Information

## Summary by CodeRabbit

- **New Features**
	- Improved user interaction by preventing default actions on label clicks within the select component.
	- Simplified interaction model for the select component by adjusting class assignments based on its filled state.

- **Bug Fixes**
	- Resolved a toggle issue in the select component that caused unexpected behavior when clicking on the label, enhancing the user experience.

### DIFF
--- a/.changeset/clean-dryers-notice.md
+++ b/.changeset/clean-dryers-notice.md
@@ -2,4 +2,4 @@
 "@nextui-org/select": patch
 ---
 
-Fix toggle issue while clicking on label of select component unexpectedly open and close the model instantly.The removal of code block containing group-data-[filled=true]:pointer-events-auto prevent from any unwanted user interactions and resolve the flickering issue.
+Fixed toggle issue while clicking on label of select component unexpectedly open and close the model instantly. The removal of code block containing `group-data-[filled=true]:pointer-events-auto` prevent from any unwanted user interactions and resolve the flickering issue (#3558)
--- a/.changeset/clean-dryers-notice.md
+++ b/.changeset/clean-dryers-notice.md
@@ -1,0 +1,5 @@
+---
+"@nextui-org/select": patch
+---
+
+Fix toggle issue while clicking on label of select component unexpectedly open and close the model instantly.The removal of code block containing group-data-[filled=true]:pointer-events-auto prevent from any unwanted user interactions and resolve the flickering issue.
--- a/packages/components/select/src/use-select.ts
+++ b/packages/components/select/src/use-select.ts
@@ -469,9 +469,6 @@ export function useSelect<T extends object>(originalProps: UseSelectProps<T>) {
       }),
       ...labelProps,
       ...props,
-      onClick: (e: React.MouseEvent<HTMLLabelElement>) => {
-        e.preventDefault();
-      },
     }),
     [slots, classNames?.label, labelProps],
   );
--- a/packages/components/select/src/use-select.ts
+++ b/packages/components/select/src/use-select.ts
@@ -469,6 +469,9 @@ export function useSelect<T extends object>(originalProps: UseSelectProps<T>) {
       }),
       ...labelProps,
       ...props,
+      onClick: (e: React.MouseEvent<HTMLLabelElement>) => {
+        e.preventDefault();
+      },
     }),
     [slots, classNames?.label, labelProps],
   );
--- a/packages/core/theme/src/components/select.ts
+++ b/packages/core/theme/src/components/select.ts
@@ -535,13 +535,6 @@ const select = tv({
         trigger: "h-16 min-h-16 py-2.5 gap-0",
       },
     },
-    //  labelPlacement=[inside, outside]
-    {
-      labelPlacement: ["inside", "outside"],
-      class: {
-        label: ["group-data-[filled=false]:pointer-events-auto"],
-      },
-    },
     {
       labelPlacement: "outside",
       isMultiline: false,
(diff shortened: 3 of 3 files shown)


wingkwong (changes requested):
In your solution, if I add a onClick to the prop, it will get replaced by yours. Also I don't think this is the correct solution even though it seems working in your demo. Actually I've already pointed out the root cause in the issue. In `ariaShouldCloseOnInteractOutside`, we are using `domRef` which is the `input`, the ref should be trigger ref which covers the label. I don't why `domRef` is used here, probably there are other cases you may need to dig.

sohan01fw:
lol, ya i thought that this simple couldn't be the solution. Ok, i will go with other solution as u mention

abhinandan-verma:
@wingkwong  I have tried changing the domRef to `triggerRef`, but it leads to another problem that on clicking the select the popover opens bit it doesn't closes on clicking again, which should happen

Actually the the button is itself being passed on as element and trigger both on clicking the label, that is why  the state is getting closed.

sohan01fw:
@abhinandan-verma bro it worked with that simple code so i think let's raise pr and iam also getting error during build time in my local so to fix and check both i did raise pr. Now it makes clear tho to solve it by figuring out the further things in deep.

sohan01fw:
@wingkwong i tried to solve it but it throws test error in select component but when i click and see the select component work fine.

Screenshot (49)

wingkwong:
I remember this issue. After opening and clicking selector button, the select is not closed.

sohan01fw:
should i push this changes or?

wingkwong:
this would break the selector button closing behavior. need a solution to cover both case.

sohan01fw:
@wingkwong when i make changes in test code it work perfectly but is it a right approach ?
Screenshot (50)

wingkwong:
Nope. It shouldn't be like that. What's the new solution btw?

sohan01fw:
nah same solution i just try to make changes in test cases to fix out the error

sohan01fw:
@wingkwong I think this is the only way . I tried everything from my side nothing works.
Screenshot (51)

wingkwong:
I don't know why you keep making changes on the test while the solution is not expected

sohan01fw:
no not this one i reset the above test changes  on this . You can see file changes

wingkwong:
I mean adding onClick is not the correct solution, even though all tests are passed and it looks working fine.

sohan01fw:
@wingkwong what u say on this ? i did change the  `label: ["group-data-[filled=true]:pointer-events-auto"]`  to  `label: ["group-data-[filled=false]:pointer-events-auto"]` and it work
Screenshot (52)

wingkwong:
1. push the code so that i could test on the build. 
2. explain why changing that works

sohan01fw:
Changing the data-filled attribute from true to false stops the flickering because it disables the CSS styles associated with the `group-data-[filled=true]:pointer-events-auto class`.

When the data-filled attribute is set to true, the `group-data-[filled=true]:pointer-events-auto` class is applied to the label element, which enables pointer events and allows the label to respond to user interactions.

However, when the data-filled attribute is set to false, the `group-data-[filled=true]:pointer-events-auto` class is removed from the label element, which disables pointer events and prevents the label from responding to user interactions.

wingkwong:
what about just removing the whole block (L538 - L544)?

sohan01fw:
yeah i think it's ok to do that cause it's not a link tho!

sohan01fw:
wait i will push code

wingkwong (commented):
please add changeset

sohan01fw:
ok

wingkwong (approved):

PR #3593, opened 2024-08-02
fix: remove disableAnimation prop from span element instead of img el…

kuri-sun:
Closes #

Im working on this linear issue: https://linear.app/nextui-inc/issue/ENG-1225/react-doesnt-recognize-disableanimation

## 📝 Description

[Cause of the error]
The error says we need to remove `disableAnimation` prop from `<span>` element. 

Screenshot 2024-08-02 at 11 36 31 PM

Currently we are trying to remove this prop from `` element instead of `<span>`. So, I removed those lines of codes.
The cause of this error is we did not extract `disabledAnimation` prop in `useAvatar` function's argument object de-structuring so that we accidentally passed `disableAnimation` to `<span>` element with `otherProps`.

[Solution]
So, to avoid passing this prop to `<span>` element(which means we shall avoid including `disableAnimation` in `otherProps`), we added the object de-structuring as a solution for this.

the error message in console

## 🚀 New behavior

Screenshot 2024-08-02 at 11 47 02 PM

Working as fine.

No

## Summary by CodeRabbit

- **Improvements**
	- Enhanced the handling of animation properties for better clarity and performance.
	- Simplified the logic for managing the animation control within the avatar component, leading to improved maintainability. 
	- Filtered DOM properties to exclude unnecessary complexity, potentially improving overall performance.

### DIFF
--- a/packages/components/avatar/src/use-avatar.ts
+++ b/packages/components/avatar/src/use-avatar.ts
@@ -1,10 +1,10 @@
 import type {AvatarSlots, AvatarVariantProps, SlotsToClasses} from "@nextui-org/theme";
-import type {DOMElement, DOMAttributes, HTMLNextUIProps, PropGetter} from "@nextui-org/system";
+import type {HTMLNextUIProps, PropGetter} from "@nextui-org/system";
 
 import {avatar} from "@nextui-org/theme";
 import {useProviderContext} from "@nextui-org/system";
 import {mergeProps} from "@react-aria/utils";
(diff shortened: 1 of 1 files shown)


wingkwong (commented):
1. can you share how to reproduce the issue?
2. `const disableAnimation = originalProps.disableAnimation ?? globalContext?.disableAnimation ?? false;` happens in other components. Is it necessary to change them as well?

wingkwong:
handled in #3946

PR #3607, opened 2024-08-04
fix(hooks): ensure exposed functions wrap by `useCallback`

AnYiEE:
Closes #

## 📝 Description

In this commit, I made improvements to our custom React hooks by wrapping the exposed functions with `useCallback`. This change aims to enhance performance.

N/A.

## 🚀 New behavior

N/A.

## 💣 Is this a breaking change

No.

## 📝 Additional Information

By wrapping functions with `useCallback`, we ensure that these functions are not recreated on every component re-render unless their dependencies change. This optimization can significantly reduce unnecessary renders and improve the application's performance, especially when these functions are passed as props to child components.

By ensuring that the functions returned by our hooks are memoized, we provide a more consistent behavior, aligning with React's best practices. This consistency is particularly beneficial when dealing with complex state management or event handlers in components.

With functions wrapped in `useCallback`, developers can more easily reason about when and why a component might re-render, improving code maintainability and debugging ease.

## Summary by CodeRabbit

- **New Features**
	- Enhanced performance of the application by optimizing hooks for clipboard handling, shape updates, and state management.
	- Improved efficiency in React components by memoizing critical functions, reducing unnecessary re-renders.
  
- **Bug Fixes**
	- Resolved issues related to function identity in props, ensuring stable behavior across component renders.

### DIFF
--- a/packages/hooks/use-real-shape/src/index.ts
+++ b/packages/hooks/use-real-shape/src/index.ts
@@ -14,7 +14,7 @@ export function useRealShape<T extends HTMLElement>(ref: RefObject<T | null>) {
     const {width, height} = getRealShape(ref.current);
 
     setState({width, height});
-  }, [ref]);
+  }, []);
 
   useEffect(() => updateShape(), [updateShape]);
 
--- a/packages/hooks/use-real-shape/src/index.ts
+++ b/packages/hooks/use-real-shape/src/index.ts
@@ -1,4 +1,4 @@
-import {RefObject, useState, useEffect} from "react";
+import {RefObject, useCallback, useState, useEffect} from "react";
 import {ShapeType, getRealShape} from "@nextui-org/react-utils";
 
 export type ShapeResult = [ShapeType, () => void];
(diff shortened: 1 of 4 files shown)


packages/hooks/use-real-shape/src/index.ts
```diff
@@ -8,15 +8,15 @@ export function useRealShape<T extends HTMLElement>(ref: RefObject<T | null>) {
...
   });
-  const updateShape = () => {
+  const updateShape = useCallback(() => {
     if (!ref?.current) return;
 
     const {width, height} = getRealShape(ref.current);
 
     setState({width, height});
-  };
+  }, [ref]);
```
wingkwong, line 17: i think we don't need to add `ref` to dependency array here.

AnYiEE: Yes, it looks like what you said from the code. It's just that my local ESLint rule "react-hooks/exhaustive-deps" thinks that "ref" needs to be one of the dependencies, and adding it doesn't have any impact, so I kept it.

wingkwong: let's remove it to keep it consistent with other code and align with the project lint.

wingkwong (approved):

PR #3643, opened 2024-08-12
fix(input): prevent clearable when input is readonly

ryxxn:
## 📝 Description

This PR addresses the issue where the clear button (`isClearable`) should not be displayed when the input field is read-only (`isReadOnly`). The `useInput` hook has been modified to automatically set `isClearable` to `false` when `isReadOnly` is `true`.

image

Currently, the clear button can still be shown even if the input field is read-only. This is not the desired behavior because users should not be able to clear a read-only input field.

## 🚀 New behavior

With this PR, when the input field is set to read-only (`isReadOnly`), the clear button (`isClearable`) will automatically be hidden, ensuring that read-only fields cannot be cleared by the user.

No

## 📝 Additional Information

The `useInput` hook has been updated to check the `isReadOnly` property and conditionally set the `isClearable` property. This change ensures consistency and improves the user experience by preventing unintended actions on read-only fields.

## Summary by CodeRabbit

- **New Features**
	- Improved usability of the input component by ensuring the clear button is disabled when the input is read-only.
  
- **Bug Fixes**
	- Resolved an issue where the clear button was incorrectly enabled on read-only input fields.

- **Tests**
	- Added new test cases to validate the behavior of the clear button when the input is in read-only mode.

### DIFF
--- a/packages/components/input/__tests__/input.test.tsx
+++ b/packages/components/input/__tests__/input.test.tsx
@@ -180,6 +180,26 @@ describe("Input", () => {
 
     expect(inputs[1]).toBeVisible();
   });
+
+  it("should not show clear button when isReadOnly is true", () => {
+    const {queryByRole} = render(
+      <Input isClearable isReadOnly defaultValue='This is "isReadOnly"' label="test input" />,
+    );
+
+    const clearButton = queryByRole("button");
+
+    expect(clearButton).toBeNull();
+  });
+
+  it("should show clear button when isReadOnly is false", () => {
+    const {getByRole} = render(
+      <Input isClearable defaultValue='This is "isReadOnly"' label="test input" />,
+    );
+
+    const clearButton = getByRole("button");
+
+    expect(clearButton).not.toBeNull();
+  });
 });
 
 describe("Input with React Hook Form", () => {
--- a/packages/components/input/__tests__/input.test.tsx
+++ b/packages/components/input/__tests__/input.test.tsx
@@ -181,24 +181,31 @@ describe("Input", () => {
     expect(inputs[1]).toBeVisible();
   });
 
-  it("should not show clear button when isReadOnly is true", () => {
-    const {queryByRole} = render(
-      <Input isClearable isReadOnly defaultValue='This is "isReadOnly"' label="test input" />,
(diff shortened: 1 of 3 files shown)


wingkwong (changes requested):
1. please add changeset since you've modified file under packages
2. please include a test

ryxxn:
> 1. please add changeset since you've modified file under packages
> 2. please include a test

Thank you for the feedback!
I've added a changeset and included a test

ryxxn:
> 1. please add changeset since you've modified file under packages
> 2. please include a test

Thank you for the feedback!

I've added a changeset and included test.

wingkwong (commented):
> Currently, the clear button can still be shown even if the input field is read-only. This is not the desired behavior because users should not be able to clear a read-only input field.

I agree users should not able to click it but i'm thinking if we should make it hidden or not. If you look at `isDisabled` + `isClearable`, the clear button is here but not able to click.

image

packages/components/input/__tests__/input.test.tsx
```diff
@@ -180,6 +180,26 @@ describe("Input", () => {
...
+    );
+
+    const clearButton = queryByRole("button");
+
+    expect(clearButton).toBeNull();
+  });
+
+  it("should show clear button when isReadOnly is false", () => {
+    const {getByRole} = render(
+      <Input isClearable defaultValue='This is "isReadOnly"' label="test input" />,
```
wingkwong, line 196: The statement `This is "isReadOnly"` seems not true in this case

ryxxn: Thank you for your feedback!
I have written new test code.

ryxxn:
> > Currently, the clear button can still be shown even if the input field is read-only. This is not the desired behavior because users should not be able to clear a read-only input field.
> 
> I agree users should not able to click it but i'm thinking if we should make it hidden or not. If you look at `isDisabled` + `isClearable`, the clear button is here but not able to click.
> 
> image

To maintain consistency with the behavior when isDisabled, I have disabled the click action for the clear button when isReadOnly is true.

wingkwong (approved):

PR #3694, opened 2024-08-29
fix(date-input): modifying validation error containing date

### LINKED ISSUE #3693: [BUG] - DateRangePicker error message not using NextUIProvider Locale
### NextUI Version

2.4.2

### Describe the bug

Set the locale in the `NextUIProvider`
```tsx
  <NextUIProvider locale='en-GB'>{children}</NextUIProvider>
```

_No response_

Set the locale in the `NextUIProvider`
```tsx
  <NextUIProvider locale='en-GB'>{children}</NextUIProvider>
```

Consume the `DateRangePicker` component

```tsx
...
// from the docs
const min = today(getLocalTimeZone());
...
<DateRangePicker
  label="Start date - End date"
  minValue={min}
/>
```

### Expected behavior

The error message should respect the provided's locale format.
ATM I receive `Value must be 8/29/2024 or later.` I expect instead `Value must be 29/08/2024 or later.`

### Screenshots or Videos

image

### Operating System Version

macOs

### Browser

Chrome

macci001:
Closes #3693

#### 📝 Description

The PR fixes the date mentioned in the validation error messages of date-picker.

#### ⛳️ Current behavior (updates)

Currently the date in the validation message of date-picker in based on the browser and not on the locale mentioned.
The reason behind this is from `react-aria` as it uses the browser for the validation.

Below shows the current behaviour:
* MinValue Validation failed
Screenshot 2024-08-29 at 11 55 33 PM
* MinValue Validation failed(with `hideTimeZone` as true)
Screenshot 2024-08-29 at 11 55 14 PM
* MaxValue Validation failed
Screenshot 2024-08-29 at 11 56 42 PM
* MaxValue Validation failed(with `hideTimeZone` as true)
Screenshot 2024-08-29 at 11 57 03 PM

#### 🚀 New behavior
Below shows the new behaviour:

* MinValue Validation failed
Screenshot 2024-08-29 at 11 48 36 PM
* MinValue Validation failed(with `hideTimeZone` as true)
Screenshot 2024-08-29 at 11 48 58 PM
* MaxValue Validation failed
Screenshot 2024-08-29 at 11 51 39 PM
* MaxValue Validation failed(with `hideTimeZone` as true)
Screenshot 2024-08-29 at 11 52 01 PM

## Summary by CodeRabbit

- **New Features**
	- Enhanced date-picker component with improved error message formatting based on locale settings.
	- Added specific error messages for date input constraints related to minimum and maximum values.

- **Bug Fixes**
	- Resolved issues with validation error handling for date inputs, ensuring accurate feedback for users.

- **Tests**
	- Added new test cases to verify locale-based error messaging for minimum and maximum date values in the DateInput component.


macci001:
@coderabbitai review

wingkwong (changes requested):
please also include test case

macci001:
Hello @wingkwong ,
I have added the test here, maybe you can have a look at this again?

wingkwong:
will do

wingkwong (changes requested):
With Chinese locale, it changes the message to english in your PR which is not expected.

image

The links I used 
- existing prod storybook: https://storybook.nextui.org/?path=/story/components-dateinput--min-date-value&globals=locale:zh-CN
- your PR storybook: https://nextui-storybook-v2-git-fork-macci001-macci00-a17be3-nextui-org.vercel.app/?path=/story/components-dateinput--min-date-value&globals=locale:zh-CN

macci001:
Applied the suggestions in: fadddd6d9c17defe23f1e285c20e7a7a0ddd9154
Screenshot 2024-09-09 at 4 43 23 PM
// cc @wingkwong

wingkwong:
@macci001 can you explain a bit why we need to include `packages/components/date-input/intl/messages.ts`? supposedly this should be handled in RA side instead of us hardcoding it (?)

macci001:
Browser is responsible for the content of the validation messages and RA has no role to play in it. So in-order to cutomize error message based on the locale provided, we need to handle it by our self.
For ref: https://github.com/adobe/react-spectrum/issues/6227#issuecomment-2066892432
// cc @wingkwong

wingkwong:
Got it. Lemme take a look later

packages/components/date-input/src/use-date-input.ts
```diff
@@ -184,6 +186,48 @@ export function useDateInput<T extends DateValue>(originalProps: UseDateInputPro
     isInvalid: ariaIsInvalid,
   } = useAriaDateField({...originalProps, label, validationBehavior, inputRef}, state, domRef);
 
+  const stringFormatter = useLocalizedStringFormatter(intlMessages) as any;
```
wingkwong, line 189: i think `as any` is not required.

packages/components/date-input/src/use-date-input.ts
```diff
@@ -184,6 +186,48 @@ export function useDateInput<T extends DateValue>(originalProps: UseDateInputPro
...
+
+  if (props.minValue != undefined && validationDetails.rangeUnderflow) {
+    const indexInValidationErrors: number =
+      Number(validationDetails.badInput) +
+      Number(validationDetails.customError) +
+      Number(validationDetails.patternMismatch);
```
wingkwong, line 195: do we need to cater `rangeUnderflow` as well?

wingkwong (changes requested):

packages/components/date-input/src/use-date-input.ts
macci001: I don't think we need `rangeUnderflow` here as only `badInput`, `customError` and `patternMatch` errors appears before `rangeUnderflow` error in validationErrors array. Hence, if some or all of them exist they will shift the index of `rangeUnderflow` in validationErrors. 

But in case of `rangeOverflow` as we have `badInput`, `customError` , `patternMatch`  and `rangeUnderflow` errors before it in validationErrors array, we need to consider all of them.

macci001: So I assumed that the errors in the `validationField` will be populated according to the occurrence of the property in [ValidityState](https://developer.mozilla.org/en-US/docs/Web/API/ValidityState).
But seems like I made a wrong assumption, as Errors related to `Bad Input` are populated in validationError after `rangeUnderflow` and `rangeOverflow` by RA .([ref](https://github.com/adobe/react-spectrum/blob/main/packages/%40react-stately/datepicker/src/utils.ts#L51-L61)).

Also, while scrapping through RA code, I understood that if ValidationError has `rangeUnderflow`/`rangeOverflow` then it can only have `badInput` error. 
Now, as we might not consider the order of errors in ValidityState as order in ValidityError, it might be bit tricky to find out which error in ValidationError list.

It is clear that if exists in ValidationError, the `rangeOverflow` and `rangeUnderflow` will be its first element. So should we edit the first value in the list?

The reason why the `rangeOverflow` and `rangeUnderflow` will always be first element:
* ValidationErrors can either be obtained from controlledError, serverError, clientError, builtinValidation, currentValidity [(ref)](https://github.com/adobe/react-spectrum/blob/main/packages/%40react-stately/form/src/useFormValidationState.ts#L145-L147) 
* controlledError, serverError, clientError and currentValidity => All of these does not have true for the ValidityState `rangeUnderflow`, `rangeOverflow` and `BadInput`. 
* `rangeOverflow` / `rangeUnderflow` happens to be in the validation errors list if it is obtained from builtinValidation.
* The order in which the errors are populated for builtinValidation is [this](https://github.com/adobe/react-spectrum/blob/main/packages/%40react-stately/datepicker/src/utils.ts#L51-L61).
* Hence, the `rangeOverflow` / `rangeUnderflow` appears to be the first element if they do exist.

macci001: Updated the following changes in 7838f48cbab983cc2c9b0c91356af5165e72f237
* Added support for badInput message with i18n.
* Modifying the validationError according exactly how they were populated by RA

ryo-manba (commented):
I think it's hard to manage error messages by country in NextUI. Validating and maintaining each language can be costly. Since [React Aria is talking to the globalization team](https://github.com/adobe/react-spectrum/issues/6342#issuecomment-2101712598), it might be best to wait for their updates.

In the meantime, you could use [React Aria hooks](https://react-spectrum.adobe.com/react-aria/useLocale.html) to show custom error messages.
Thanks for the PR!

devongovett:
Hi, we consulted our globalization team, and decided not to change anything in React Aria. These validation error messages come from the browser, so they will be in the user's browser language setting (e.g. `navigator.langauge`). Unfortunately we don't have any control over that, so it cannot respect the `IntlProvider` setting. We decided that it was not worth the bundle size cost of including our own localized strings for every possible form error message (there are a lot). In most cases for real users, the browser locale and app-level language setting will match so we don't think this will be a common enough issue to justify the cost. If it is a real problem in your specific application then you can provide your own custom validation messages.

ryo-manba:
Thanks @devongovett , for the clarification and your team's work on this.
I’ll go ahead and close this PR now.

PR #3735, opened 2024-09-10
fix(table): Add an isKeyboardNavigationDisabled prop to disable the navigation

### LINKED ISSUE #3681: Cannot perform inline editing of input/textarea if in table cell
### NextUI Version

2.4.5

### Describe the bug

I have a scenario where I have a Table component and inside the table I have an input/textarea field, if i try to edit the text using arrow keys the navigation takes me to next cell of table.
I have tried using `e.preventDefault();` but it doesn't work. how do I disable the arrow key navigation in table and allow user to navigate the text.

_No response_

Create a table and inside table cell add input field or textarea.

Following is my code:

```
const handleKeyDown = (e: React.KeyboardEvent<HTMLTextAreaElement>) => {
   const { key } = e;
   if (key === 'ArrowUp' || key === 'ArrowDown' || key === 'ArrowLeft' || key === 'ArrowRight') {
       e.stopPropagation(); // Prevent the event from bubbling up to the table
        // Optionally, you can prevent default behavior to stop scrolling, if any
       e.preventDefault();   
}};
```

```
<TableCell className="px-0.5">
    <div>
       <Textarea
            variant="bordered"
            maxRows={1}
            minRows={1}
            placeholder="Name"
            disableAnimation
            onChange={(e) => onDetailChange(e.target.value, index, 'name')}
           defaultValue={employee.name}
            classNames={{
                input: "resize-y",
            }}
            onKeyDown={(e) => handleKeyDown(e)}
        />
    </div>
</TableCell>
```

### Expected behavior

It should let me perform inline editing and not navigate to next cell.

### Screenshots or Videos

_No response_

### Operating System Version

macOS

### Browser

Chrome

macci001:
Closes #3681 

#### 📝 Description

> PR adds isKeyboardNavigationDisabled prop to disable navigation on keypress

#### ⛳️ Current behavior (updates)

* Currently, whenever any arrow-key keypress is triggered it navigates the focus to other cell/row.
* This maybe an issue when the table cell contains a component which requires this keys for specific purpose.
* Example for above scenario could be: table cell contains input component, it might need arrow keys for editing. But it is not possible because whenever the keypress triggers navigation

https://github.com/user-attachments/assets/61ae353d-cdca-4cbf-b12d-1146da371704

#### 🚀 New behavior

* The PR adds an `isKeyboardNavigationDisabled` prop to disable the navigation through keyboard.

https://github.com/user-attachments/assets/c05c084a-8caf-4959-aee8-102818220a0a

#### Additional info
* updated the docs
Screenshot 2024-09-10 at 4 34 44 PM

## Summary by CodeRabbit

- **New Features**
  - Introduced `isKeyboardNavigationDisabled` prop for the Table component, allowing users to disable keyboard navigation when interacting with components that require arrow key inputs.
  
- **Documentation**
  - Enhanced Table component documentation with details about the new `isKeyboardNavigationDisabled` prop, updated props and events tables, and added usage examples for improved clarity and usability.

### DIFF
--- a/.changeset/fluffy-icons-refuse.md
+++ b/.changeset/fluffy-icons-refuse.md
@@ -1,0 +1,5 @@
+---
+"@nextui-org/table": patch
+---
+
+Currently, whenever any arrow-key keypress is triggered it navigates the focus to other cell/row. This creates an issue when the table cell contains a component which requires this keys for specific purpose (eg. if a table cell contains input component, it might need arrow keys for editing. But it is not possible because whenever the keypress triggers navigation). The PR adds an `isKeyboardNavigationDisabled` prop to disable the navigation.
--- a/apps/docs/content/docs/components/table.mdx
+++ b/apps/docs/content/docs/components/table.mdx
@@ -457,7 +457,7 @@ You can customize the `Table` component by passing custom Tailwind CSS classes t
 | disableAnimation              | `boolean`                                                                                                                                                     | Whether to disable the table and checkbox animations.                                                        | `false`     |
 | checkboxesProps               | [CheckboxProps](/docs/components/checkbox/#checkbox-props)                                                                                                    | Props to be passed to the checkboxes.                                                                        | -           |
 | classNames                    | `Record<"base" ｜ "table" ｜ "thead" ｜ "tbody" ｜ "tfoot" ｜ "emptyWrapper" ｜ "loadingWrapper" ｜ "wrapper" ｜ "tr" ｜ "th" ｜ "td" ｜ "sortIcon", string>` | Allows to set custom class names for the dropdown item slots.                                                | -           |
-
+| isKeyboardNavigationDisabled  | `boolean`                                                                                                                                                     | Whether to disable keyboard navigations or not.                                                              | `false`     |
 ### Table Events
 
 | Attribute         | Type                                  | Description                                                         |
--- a/packages/components/table/__tests__/table.test.tsx
+++ b/packages/components/table/__tests__/table.test.tsx
@@ -1,8 +1,10 @@
 import * as React from "react";
-import {act, render} from "@testing-library/react";
+import {act, render, fireEvent} from "@testing-library/react";
 import userEvent, {UserEvent} from "@testing-library/user-event";
 
+
 import {Table, TableHeader, TableCell, TableColumn, TableBody, TableRow} from "../src";
+import {keyCodes} from "../../../utilities/test-utils/src";
 
 const columns = [
   {name: "Foo", key: "foo"},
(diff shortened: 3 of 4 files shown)


AnYiEE:
I think a better solution is to automatically disable keyboard navigation temporarily when the focus is on the input, so that can move the cursor with the keyboard and use the Tab key to navigate to other elements of the table after leaving the input.

macci001:
Thanks for the suggestion @AnYiEE. But I do not think it would be a great idea to write in a logic that checks if component is input component then stop the keyboard navigation.
One work-around which I thought could be helpful is rather than adding a prop `isKeyboardNavigationDisabled` to the table, we can add this prop to any of the cell. If any cell with `isKeyboardNavigationDisabled` as `true`, keyboard navigation will be stopped when then cell is in focus.

any thoughts here @wingkwong, should we go forward with adding the prop to the `Table` or it would be helpful to add the functionality to the `TableCell` instead.

soyeon9211:
@macci001 Is it not merge yet? I need this option :(

macci001:
@wingkwong can I please get in some review here?

ryo-manba (commented):
Ideally, I think it would be best to add an edit mode to the table. Since the React Spectrum team is [working on this](https://github.com/adobe/react-spectrum/issues/2328), we can either wait for them to fix it or implement it ourselves.

macci001:
Reflecting on your [comment](https://github.com/nextui-org/nextui/pull/3735#pullrequestreview-2354564738) @ryo-manba
I can think of following design for the edit mode: 

* In any of the cases, one when the cell has focusable children and other when cell do not have focusable children, always entire cell is focused. That is, focusMode set to `cell` instead of `child`.
* Whenever there is `Enter` keypress on a focused cell:
   *  If cell does have some focusable children:
      * Set the focus on the first/last focusable children (based on RTL)
      * Add the keyboard navigation on `shift + leftarrowkey` and `shift + rightarrowkey` for navigating to prev/next focusable children of the cell.
   * If cell does not have any focusable children -> do nothing i.e. edit mode is not relevant in this case. 
*  `Esc/Tab` keypress to exit the edit mode.


Note 1: For navigation between focusable children of the cell, `getFocusableTreeWalker` will be used.

Note 2: For now, the `useTableCell` internally uses `useGridCell`'s default value for focusMode i.e.  `child`. This means, that  whenever the cell will be in focus, the focus will shift to the focusableChild if it exist, else entire cell would be in focus. In order  to implement the edit mode, we would need the react-aria team to add a prop by which we can set the focusMode to `cell`.

ryo-manba:
@macci001 
Thank you for your detailed insights. I'll discuss it with the team and see how we can proceed.

jrgarciadev (approved):

jrgarciadev:
Hey @macci001 please fix the conflicts

jrgarciadev (approved):

PR #3774, opened 2024-09-17
fix(input): ensure clear button is not focusable when disabled

ryxxn:
## 📝 Description

This PR fixes an issue where the `clear button` inside the `Input` component could still receive focus when the input is disabled (`isDisabled`). The button is now properly excluded from the tab order by setting `tabIndex` to `-1` when the input is disabled.

https://github.com/user-attachments/assets/cc929087-c1ee-4887-a0b0-6888a2ca3ed1

Currently, the `clear button` in the `Input` component can still receive focus via the Tab key when the input is disabled. This behavior can cause accessibility issues and an inconsistent user experience.

## 🚀 New behavior

With this fix, the `clear button` will no longer be focusable when the input is disabled. The `tabIndex` for the `clear button` is set to `-1` when `isDisabled` is true, preventing it from being part of the tab order.

## 📝 Additional Information

This fix enhances accessibility and user experience by ensuring that the `clear button` behaves correctly when the input is disabled.

## Summary by CodeRabbit

- **New Features**
	- Enhanced accessibility for the clear button in the input component by preventing focus when the input is disabled.
	- Updated the clear button from a `<span>` to a `<button>` element for improved semantic meaning and usability.
- **Bug Fixes**
	- Adjusted the `tabIndex` property for the clear button to ensure compliance with keyboard navigation standards.
- **Tests**
	- Added new test cases to verify the correct behavior of the clear button's state when the input is disabled and ensure it is not focusable.

### DIFF
--- a/packages/components/input/__tests__/input.test.tsx
+++ b/packages/components/input/__tests__/input.test.tsx
@@ -38,9 +38,9 @@ describe("Input", () => {
   });
 
   it("should disable the clear button when isDisabled", () => {
-    const {container} = render(<Input isClearable isDisabled label="test input" />);
+    const {getByRole} = render(<Input isClearable isDisabled label="test input" />);
 
-    const clearButton = container.querySelector("button");
+    const clearButton = getByRole("button");
 
     expect(clearButton).toBeDisabled();
   });
@@ -139,7 +139,7 @@ describe("Input", () => {
 
     const ref = React.createRef<HTMLInputElement>();
 
-    const {container} = render(
+    const {getByRole} = render(
       <Input
         ref={ref}
         isClearable
@@ -149,7 +149,7 @@ describe("Input", () => {
       />,
     );
 
-    const clearButton = container.querySelector("button")!;
+    const clearButton = getByRole("button")!;
 
     expect(clearButton).not.toBeNull();
 
@@ -194,7 +194,7 @@ describe("Input", () => {
 
     const ref = React.createRef<HTMLInputElement>();
 
-    const {container} = render(
+    const {getByRole} = render(
       <Input
         ref={ref}
         isClearable
@@ -205,7 +205,7 @@ describe("Input", () => {
       />,
     );
 
-    const clearButton = container.querySelector("button")!;
+    const clearButton = getByRole("button")!;
 
     expect(clearButton).not.toBeNull();
 
--- a/packages/components/input/__tests__/input.test.tsx
+++ b/packages/components/input/__tests__/input.test.tsx
@@ -37,6 +37,14 @@ describe("Input", () => {
     expect(container.querySelector("input")).toHaveAttribute("disabled");
   });
 
+  it("should set tabIndex to -1 when isDisabled is true", () => {
+    const {getByRole} = render(<Input isClearable isDisabled label="test input" />);
+
+    const clearButton = getByRole("button");
+
+    expect(clearButton).toHaveAttribute("tabIndex", "-1");
+  });
+
   it("should have required attribute when isRequired with native validationBehavior", () => {
     const {container} = render(<Input isRequired label="test input" validationBehavior="native" />);
 
--- a/packages/components/input/__tests__/input.test.tsx
+++ b/packages/components/input/__tests__/input.test.tsx
@@ -45,6 +45,14 @@ describe("Input", () => {
     expect(clearButton).toBeDisabled();
   });
 
+  it("should not allow clear button to be focusable", () => {
+    const {getByRole} = render(<Input isClearable label="test input" />);
+
+    const clearButton = getByRole("button");
+
+    expect(clearButton).toHaveAttribute("tabIndex", "-1");
+  });
+
   it("should have required attribute when isRequired with native validationBehavior", () => {
     const {container} = render(<Input isRequired label="test input" validationBehavior="native" />);
 
--- a/packages/components/input/__tests__/input.test.tsx
+++ b/packages/components/input/__tests__/input.test.tsx
@@ -37,12 +37,12 @@ describe("Input", () => {
     expect(container.querySelector("input")).toHaveAttribute("disabled");
   });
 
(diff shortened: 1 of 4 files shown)


packages/components/input/src/use-input.ts
```diff
@@ -511,7 +511,7 @@ export function useInput<T extends HTMLInputElement | HTMLTextAreaElement = HTML
       return {
         ...props,
         role: "button",
-        tabIndex: 0,
+        tabIndex: originalProps?.isDisabled ? -1 : 0,
```
ryo-manba, line 514: It seems like you can use the button element and pass the `disabled` attribute, which would remove the need for setting `tabindex` and `role: "button"`.
https://github.com/nextui-org/nextui/blob/bed4a59b8a89240cf9706266a31d7e554a627e87/packages/components/input/src/input.tsx#L39

packages/components/input/__tests__/input.test.tsx
```diff
@@ -37,6 +37,14 @@ describe("Input", () => {
     expect(container.querySelector("input")).toHaveAttribute("disabled");
   });
 
+  it("should set tabIndex to -1 when isDisabled is true", () => {
```
ryo-manba, line 40: The test should verify that clicking the button does not clear the input.

ryo-manba (changes requested):

packages/components/input/__tests__/input.test.tsx
ryxxn: image

The clear button is already non-functional when isDisabled is true.

packages/components/input/src/use-input.ts
ryxxn: Thank you for your comment! 

I agree with using the button element.
Do you think it would be better to proceed with that change?

ryo-manba: Yes, I think that would be better.

packages/components/input/__tests__/input.test.tsx
ryo-manba: That’s true, but if you’re adding tests, I would appreciate it if you could verify that the button is disabled. 
I think specifying the `tabindex` will be resolved by marking it up as a `button`.

ryxxn:
I have updated the code to replace the span element with a button element for the clear button. 

This change removes the need to manually handle tabIndex and role, while improving accessibility.

packages/components/input/__tests__/input.test.tsx
```diff
@@ -37,12 +37,12 @@ describe("Input", () => {
     expect(container.querySelector("input")).toHaveAttribute("disabled");
   });
 
-  it("should set tabIndex to -1 when isDisabled is true", () => {
-    const {getByRole} = render(<Input isClearable isDisabled label="test input" />);
+  it("should disable the clear button when isDisabled", () => {
+    const {container} = render(<Input isClearable isDisabled label="test input" />);
 
-    const clearButton = getByRole("button");
+    const clearButton = container.querySelector("button");
```
ryo-manba, line 43: Can you use `getByRole` instead?

packages/components/input/__tests__/input.test.tsx
```diff
@@ -149,7 +149,7 @@ describe("Input", () => {
       />,
     );
 
-    const clearButton = getByRole("button");
+    const clearButton = container.querySelector("button")!;
```
ryo-manba, line 152: It would be preferable to use `getByRole` wherever possible.
ref: https://testing-library.com/docs/queries/about/#priority

ryxxn: Thanks for the suggestion!
I've updated the tests to use getByRole("button") as recommended.

packages/components/input/src/use-input.ts
```diff
@@ -510,8 +510,8 @@ export function useInput<T extends HTMLInputElement | HTMLTextAreaElement = HTML
     (props = {}) => {
       return {
         ...props,
-        role: "button",
-        tabIndex: 0,
+        type: "button",
```
ryo-manba, line 513:
```suggestion
        type: "button",
        tabIndex: -1,
```

It seems better for keyboard users if the clear button is not focusable via the Tab key.

see: https://github.com/mui/material-ui/issues/30815#issuecomment-2296573576

ryxxn: Thank you for the suggestion!
I've updated the code to set tabIndex to -1 as recommended.

ryo-manba (approved):
Good! Thanks for the PR!

wingkwong (changes requested):
clear button is not focusable when the input is not disabled. please add a test case to cover this base case as well.

Your PR: https://nextui-storybook-v2-git-fork-ryxxn-fix-input-6107e9-nextui-org.vercel.app/iframe.html?args=&id=components-input--clearable&viewMode=story

Production: https://storybook.nextui.org/iframe.html?globals=locale:bg-BG&args=&id=components-input--clearable&viewMode=story

ryxxn:
@wingkwong 

I previously received feedback suggesting that setting `tabIndex="-1"` for the clear button to make it non-focusable at all times would improve web accessibility. 

https://github.com/nextui-org/nextui/pull/3774/files/b3bd46ab9434449c5db01f1f9daf0d1dd893ed9c#r1774284969



Could we discuss which approach aligns better with the project's accessibility guidelines? Should the clear button always be non-focusable for accessibility, or should it remain focusable when the input is not disabled?

I can adjust the implementation based on the preferred approach.

wingkwong:
I see. Lemme discuss with Ryo internally first and get back to you afterwards. Thanks.

wingkwong:
confirmed the clear button will not be focusable no matter it is disabled or not.

ref: 
- https://react-spectrum.adobe.com/react-aria/SearchField.html
- https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/search

PR #3789, opened 2024-09-22
fix(date-picker): month and year picker scroller

### LINKED ISSUE #3785: [BUG] - Date Picker Showing Wrong Year
### NextUI Version

2.4.6

### Describe the bug

image

Speaks for itself really... 1987 is in the scroller, but 1986 in the focus. Which is it going to pick? Not clear...

A few scrolls later and this even more confusing collection of conflicts appears:

image

This was my code (using react-hook-form):

```
  const dateOfBirth = watch('dateOfBirth');

  const setDateValue = useCallback(
    (value: CalendarDate) => {
      const isoString = value
        .toDate(getLocalTimeZone())
        .toISOString()
        .split('T')[0];
      setValue('dateOfBirth', isoString);
    },
    [setValue]
  );
```

```
          <DatePicker
            name={'dateOfBirth'}
            aria-label={'Date of Birth'}
            label={'Date of Birth'}
            value={parseDate(dateOfBirth)}
            showMonthAndYearPickers={true}
            onChange={setDateValue}
          />
```

_No response_

Use the date picker in a component. Could it be related to using a controlled value?

### Expected behavior

I would expect the highlighted, focused and selected values to all be in agreement.

### Screenshots or Videos

_No response_

### Operating System Version

- Windows 10 Home V.22H2

### Browser

Chrome

buchananwill:
Closes #3785

## 📝 Description

> Fixing the month/year UI sync bug.

> The month and year were incorrectly updating due to a UI rectangle intersection error. This has been removed and the data is updated directly from the value attribute.

## 🚀 New behavior

No.

## Summary by CodeRabbit

- **New Features**
	- Introduced the `AccordionTree` component for hierarchical data representation.
	- Added an aria label for the month/year picker toggle button to enhance accessibility.
	- Enhanced keyboard interaction capabilities in the calendar picker component.
	- Added a new custom hook for managing key press states and scroll events.
- **Bug Fixes**
	- Improved user interface and data synchronization for the month/year picker component.
- **Refactor**
	- Simplified scrolling behavior and focused date management in the calendar picker component.
- **Chores**
	- Updated versions for `@nextui-org/accordion` and `@nextui-org/calendar` packages.
	- Added a new test case for the month and year picker functionality in the `DatePicker` component.


wingkwong (changes requested):
1. please include a before/after video and include steps on how to reproduce the issue.
2. please include the issue link if possible.
3. please remove accordion code. it's not related to this PR.
4. please include a test case.

wingkwong (changes requested):
just keep one changeset file where no accordion is mentioned

buchananwill:
> just keep one changeset file where no accordion is mentioned

Is that correct now?

wingkwong (changes requested):

buchananwill:
Please suggest an alternative way to mark this component so the test can
find it.

wingkwong:
In the test, you can simply get the element by `container.querySelector("[data-slot='header']")`

ryo-manba:
Please include the content of this comment in the PR description.
https://github.com/nextui-org/nextui/pull/3789#issuecomment-2367949681

packages/components/date-picker/__tests__/date-picker.test.tsx
```diff
@@ -600,6 +601,48 @@ describe("DatePicker", () => {
       expect(onHeaderExpandedChangeSpy).not.toHaveBeenCalled();
     });
 
+    it("should focus the month and year when pressed in the picker", () => {
```
ryo-manba, line 604: I’m unclear about what you want to verify with this test. I think the issue is that the focus indicator misaligns when the up and down arrow keys are held down. We should verify that it is synchronized properly.

ryo-manba (changes requested):
This implementation prevents switching values using the up and down arrow keys, so please fix that.

https://github.com/user-attachments/assets/07928b41-a308-42ec-92da-b4a92072835c

packages/components/date-picker/__tests__/date-picker.test.tsx
buchananwill: Sorry, forgot to check the keyboard inputs as well. I've gone round the UI one more time, testing:

- mouse
- keyboard individual keystrokes
- held down keys

The current behaviour seems to be:
- mouse: blue focus and grey highlighter always in sync throughout interactions
- individual keystrokes: same
- HELD KEYS: the blue focus indicator always syncs to the focused date, but _sometimes_ it manages to travel beyond the grey highlighter, but will immediately return on the next **individual** keystroke.

This is from my perspective a big improvement, because the toggle on/off focus mutation is gone, and the only time the UI desyncs is _(sometimes)_ when you hold down a key.

wingkwong:
@buchananwill please click re-request review once it's ready for review.

buchananwill:
I had one more idea, which I've added. The PickerItems now also have a keyUp listener, which just calls the scrollIntoView function. This seems to be a bullet proof way of ensuring that if you've been holding down any of the directional keys, then whatever was the last key to fire (and thus has focus) is scrolled to the desired position.

I have to confess my weaker experience with writing imperative DOM events, so please if anyone with that could ponder whether this broad-brush fix could have any unexpected side-effects?

I also changed the home/end values of the keyDown event to 1 and 12, since months are 1-indexed, not zero indexed. These buttons don't work for the year picker, since 1 and 12 aren't valid years within the available range. Couldn't see an obvious way to pass in the max/min years that would be dynamically set (since it could vary with locale etc.).

Hopefully this wraps up the functionality of this component & hook?

packages/components/calendar/src/use-calendar-picker.ts
```diff
@@ -221,11 +177,30 @@ export function useCalendarPicker(props: CalendarPickerProps) {
 
       const nextItem = map.get(nextValue);
 
+      scrollTo(nextValue, list);
```
ryo-manba, line 180: Two things need fixing in the Home/End key behavior:
1. The year isn't changing.
2. It's jumping straight to the start/end instead of moving smoothly as it did before.

packages/components/date-picker/__tests__/date-picker.test.tsx
```diff
@@ -600,6 +601,48 @@ describe("DatePicker", () => {
...
+
+      expect(monthBefore).toHaveAttribute("data-value", "3");
+      expect(yearBefore).toHaveAttribute("data-value", "2023");
+      expect(monthBefore).toHaveAttribute("tabindex", "-1");
+      expect(yearBefore).toHaveAttribute("tabindex", "-1");
+      triggerPress(monthBefore);
+      triggerPress(yearBefore);
+      expect(valueChangeSpy).not.toHaveBeenCalled();
+      expect(monthBefore).toHaveAttribute("tabindex", "0");
+      expect(yearBefore).toHaveAttribute("tabindex", "0");
```
ryo-manba, line 643: Please also check the following two points:

1. Ensure that the value in the header is synchronized.
2. Verify that pressing the Enter key triggers the `valueChangeSpy`.

ryo-manba (changes requested):

ryo-manba:
The Home/End keys seem to already function in the [original implementation](https://nextui.org/docs/components/date-picker#with-month-and-year-pickers). Where did you find the information about the month's index?

> I also changed the home/end values of the keyDown event to 1 and 12, since months are 1-indexed, not zero indexed. These buttons don't work for the year picker, since 1 and 12 aren't valid years within the available range.

buchananwill:
I added two additional hooks to cover the different behaviours needed for the different navigation keys:

1. Home/End use a deferred focus that only triggers when the scroll ends. Otherwise, the focus event overrides the scrolling, and the element immediately snaps into view.
2. Home/End also use a repeated key blocker, to prevent the value from receiving more than one update from these keys, since 
    a. repetition of this event is meaningless
    b. it was playing havoc with the other scrolling/focus events/behaviours.

If you have existing hooks with the required API already in the codebase, feel free to swap them in. Equally, feel free to migrate them to a different (generic) hook folder if you want to centralize them for other scenarios to use.

I can't find any more bugs in this component.

On a side note, I prefer the simpler version where home/end cause the instant scrolling. I get that this is a minor inconsistency in the style of the UI, but the two extra control flow checks to defer the focus update _and_ block repeat key strokes seems overly convoluted. I wouldn't be surprised to find it has negative performance implications or introduces other bugs on different systems from my own...

Let me know what you think! :-)

ryo-manba:
> On a side note, I prefer the simpler version where home/end cause the instant scrolling. I get that this is a minor inconsistency in the style of the UI, but the two extra control flow checks to defer the focus update and block repeat key strokes seems overly convoluted. I wouldn't be surprised to find it has negative performance implications or introduces other bugs on different systems from my own...

This seems to be a separate issue from the current one. If you'd like to address it, please open a new issue for further discussion.

packages/components/calendar/src/use-scroll-end-callback.ts
```diff
@@ -0,0 +1,69 @@
+import {useCallback, useEffect, useRef} from "react";
+
+function useScrollEndCallback(debounce: number) {
```
ryo-manba, line 3: The value is not being synced during mouse scroll. Please fix it.

packages/components/date-picker/__tests__/date-picker.test.tsx
ryo-manba: Please add tests.

ryo-manba (changes requested):

packages/components/calendar/src/use-scroll-end-callback.ts
buchananwill: Done.

buchananwill: Could you please specify the tests you need? Given that original component had tests that failed to catch this bug, it's a somewhat open-ended task. My current codebase uses with a different UI library, so it's difficult to justify spending any more time on the NextUI repo at the moment, but if you can ring-fence exactly what needs doing I will make time for that.

luka-on-gh:
Any updates on this PR?

Issue #1403, opened 2023-08-15
[BUG] - Using Dropdown component in Next.JS throws a runtime exception


iyxan23:
### NextUI Version

^2.0.15

### Describe the bug

Using the Dropdown component as stated in the docs causes a Runtime Exception when the trigger is clicked. The error itself does not provide any meaningful information for debugging: `Error: Unknown element <[object Object]> in collection.`, so I believe this is an internal bug. Source: https://nextui.org/docs/components/dropdown

### Steps to Reproduce the Bug or Issue

1. Install NextUI as described in the docs
2. Copy the code from https://nextui.org/docs/components/dropdown#usage
3. Clicks the dropdown button
4. Raised a cryptic runtime exception `Error: Unknown element <[object Object]> in collection.`

Additional info: I'm using Vivaldi, a derivative of chrome.

### Expected behavior

I expected the dropdown to show and to not raise an exception.

### Screenshots or Videos

https://github.com/nextui-org/nextui/assets/31884435/455677d5-ee0f-45f6-bf88-97660457b20a

### Operating System Version

Fedora 38 KDE Plasma; Linux laptop 6.3.11-200.fc38.x86_64 #1 SMP PREEMPT_DYNAMIC Sun Jul  2 13:17:31 UTC 2023 x86_64 GNU/Linux

### Browser

Other (add additonal context)

robodove:
Happens on Tabs too @tabs/2.0.11

For now wrapping them in a client component ("use client") makes them work.

jrgarciadev:
Hey @iyxan23 we are waiting for the next [react-aria](https://github.com/adobe/react-spectrum) release to get this issue fixed, the current collection approach is compiled by next.js in a different way so it removes the `use client;` directive, in the meantime please add the directive on top of your files for collection based components -> Table, Accordion, Dropdown, and Tabs.

iyxan23:
> Hey @iyxan23 we are waiting for the next [react-aria](https://github.com/adobe/react-spectrum) release to get this issue fixed, the current collection approach is compiled by next.js in a different way so it removes the `use client;` directive, in the meantime please add the directive on top of your files for collection based components -> Table, Accordion, Dropdown, and Tabs.

Alright! I'll be keeping an eye on new updates 😉
Thanks for the quick response

guivalerioS:
Hy guys, Do we have an estimate of when this will be fixed?

JimmyRice:
The Select component doesn't seem to be working either.

[...]

PR #3811, opened 2024-09-27
docs(accordion): note `use client;` directive

jjangga0214:
Refer to https://github.com/nextui-org/nextui/issues/1403

Note: I think other components like Dropdown should be also documented, but did not include it in this PR.

## Summary by CodeRabbit

- **Documentation**
	- Added a note in the accordion component documentation to specify the `use client;` directive explicitly until issue [#1403](https://github.com/nextui-org/nextui/issues/1403) is resolved.


wingkwong:
It should apply for all collection-based components. However, RA team has supported custom collection renderers so probably it should be resolved after upgrading RA dependencies. If not, we will add the notes then.

Scholar01:
Any news? @wingkwong

wingkwong:
@Scholar01 you may subscribe PR 3732 for the status of RA version bump.

PR #3829, opened 2024-10-01
feat(autocomplete): add autoHighlight prop

dgz9:
Closes # N/A

## 📝 Description

> Introduces a new `autoHighlight ` prop to the Autocomplete component, which automatically highlights the first item in the dropdown list as the user types.

> Currently, the Autocomplete component does not automatically highlight any option when the user types. Users need to manually navigate through the list to select an option.

## 🚀 New behavior

> - A new `autoHighlight ` prop is added to the Autocomplete component.
> - When `autoHighlight ` is set to true, the first item in the dropdown list is automatically highlighted as the user types.
> - Users can immediately press Enter to select the highlighted option without needing to navigate through the list.
> - If `isAutoHighlighted` is false or not provided, the component behaves as it did previously.

> No

## 📝 Additional Information

> Added SB test and sample to documentation.

## Summary by CodeRabbit

- **New Features**
  - Introduced an autocomplete component for selecting animals with automatic highlighting for the first suggestion.
  - Added a new property, `autoHighlight`, to enhance user experience in both autocomplete and listbox components.

- **Documentation**
  - Updated documentation for the `Autocomplete` component to include new features and clarify existing properties.
  - Expanded API section with detailed descriptions of new properties and events related to the autocomplete functionality.

- **Bug Fixes**
  - Improved logic for highlighting items in the listbox based on user interactions.

### DIFF
--- a/packages/components/autocomplete/stories/autocomplete.stories.tsx
+++ b/packages/components/autocomplete/stories/autocomplete.stories.tsx
@@ -65,7 +65,7 @@ export default {
         type: "boolean",
       },
     },
-    isAutoHighlight: {
+    autoHighlight: {
       control: {
         type: "boolean",
       },
@@ -810,7 +810,7 @@ const WithReactHookFormTemplate = (args: AutocompleteProps) => {
 
 const AutoHighlightTemplate = ({color, variant, ...args}: AutocompleteProps<Animal>) => (
   <Autocomplete
-    isAutoHighlight
+    autoHighlight
     className="max-w-xs"
     color={color}
     defaultItems={animalsData}
--- a/packages/components/autocomplete/stories/autocomplete.stories.tsx
+++ b/packages/components/autocomplete/stories/autocomplete.stories.tsx
@@ -65,6 +65,11 @@ export default {
         type: "boolean",
       },
     },
+    isAutoHighlight: {
+      control: {
+        type: "boolean",
+      },
+    },
     validationBehavior: {
       control: {
         type: "select",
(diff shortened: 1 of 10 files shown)


wingkwong (changes requested):
1. the main logic would be taken in https://github.com/nextui-org/nextui/pull/2186. Please sync the latest canary.
2. please add tests for that
3. please add changeset

dgz9:
1. Removed my logic and only left the prop to enabled/disable the highlight feature in #2186
2. Added tests for the new prop.
3. Added changeset.

wingkwong (changes requested):
please rename the prop to `autoHighlight`.

dgz9:
Prop renamed to autoHiglight

packages/components/listbox/src/use-listbox-item.ts
```diff
@@ -21,6 +21,7 @@ import {ListState} from "@react-stately/list";
 interface Props<T extends object> extends ListboxItemBaseProps<T> {
   item: Node<T>;
   state: ListState<T>;
+  autoHighlighted?: boolean;
```
wingkwong, line 24: put it into `ListboxItemBaseProps`

packages/components/autocomplete/stories/autocomplete.stories.tsx
```diff
@@ -1061,3 +1081,10 @@ export const FullyControlled = {
     ...defaultProps,
   },
 };
+
+export const WithAutoHighlight = {
+  render: AutoHighlightTemplate,
```
wingkwong, line 1086: there's no need to create a new template for that. You can simply use `Template` one and just pass `autoHighlight` to args below.

packages/components/listbox/src/use-listbox.ts
```diff
@@ -92,6 +92,11 @@ interface Props<T> extends Omit<HTMLNextUIProps<"ul">, "children"> {
    * The menu items classNames.
    */
   itemClasses?: ListboxItemProps["classNames"];
+  /**
+   * Whether to automatically highlight the first item in the list.
```
wingkwong, line 96: first non-disabled item

wingkwong (changes requested):
The following case failed. Please add a new test case to cover that as well.

1. go to autocomplete story
2. open autocomplete (either by clicking input or by clicking selection button)
3. the listbox is open, and the first item is highlighted (expected)
4. close it
5. open it again, the first item is not highlighted.

dgz9:
@wingkwong  If you're okay with it, I think this can be closed. I initially just wanted the first item on the list to be selected by default and didn't realize someone else had already pushed that to canary. I don't think a prop is necessary to enable/disable the auto highlight feature.

PR #3861, opened 2024-10-09
fix(theme): show margin only with label in Switch component

### LINKED ISSUE #3791: [BUG] - SWITCH Component Width Excessely
### NextUI Version

2.4.8

### Describe the bug

image

`<Switch
        onClick={() =>
          theme === "light" ? setTheme("dark") : setTheme("light")
        }
        defaultSelected
        size="lg"
        color="primary"
        startContent={<SunIcon />}
        endContent={<MoonIcon />}
         className="border-1"
      ></Switch>`

_No response_

Code above

### Expected behavior

Should be no more width

### Screenshots or Videos

_No response_

### Operating System Version

macOS, Windows

### Browser

Chrome

ShrinidhiUpadhyaya:
Closes #3791 

## 📝 Description
Removed the extra right margin of the wrapper.

image

## 🚀 New behavior
image

## Summary by CodeRabbit

- **New Features**
	- Updated styling for the toggle component, enhancing its visual appearance.
  
- **Bug Fixes**
	- Adjusted spacing for the toggle component in small, medium, and large sizes to improve layout consistency.
	- Improved layout behavior of the Switch component by showing margin only when a label is present.


macci001:
Hey @ShrinidhiUpadhyaya, could you please add the changeset as well.
You can run `pnpm changeset` for the same.

packages/core/theme/src/components/toggle.ts
```diff
@@ -100,7 +100,7 @@ const toggle = tv({
     },
     size: {
       sm: {
-        wrapper: "w-10 h-6 mr-2 rtl:ml-2 rtl:mr-[unset]",
+        wrapper: "w-10 h-6 rtl:ml-2 rtl:mr-[unset]",
```
macci001, line 103: Maybe if we are going forward with this. We should also fix the left margin in case of `rtl`.
Screenshot 2024-10-10 at 12 07 12 PM

wingkwong: also please don't use `rtl:`. we're trying to get rid of it. ref: https://tailwindcss.com/blog/tailwindcss-v3-3#simplified-rtl-support-with-logical-properties

macci001: Ack, noted. Thanks for the blog link.
It would make the styles much more clean imo with logical properties.

wingkwong:
@ShrinidhiUpadhyaya please resolve the conflict and click re-request review once it's done.

packages/core/theme/src/components/toggle.ts
```diff
@@ -100,7 +100,7 @@ const toggle = tv({
     },
     size: {
       sm: {
-        wrapper: "w-10 h-6 rtl:ml-2 rtl:mr-[unset]",
+        wrapper: "w-10 h-6 me-2",
```
macci001, line 103: The occurance of `me-2` should be removed because they will be adding margin in the end i.e. `mr-2` in case of ltr and `ml-2` in case of rtl.

wingkwong:
@ShrinidhiUpadhyaya there's no change to review.

ShrinidhiUpadhyaya:
> @ShrinidhiUpadhyaya there's no change to review.

I guess we can close this PR, because #3868 had similar changes for the fix and it has been merged with canary.

wingkwong:
@ShrinidhiUpadhyaya  That PR has been merged into canary but I could still reproduce the issue on canary branch.

image

ShrinidhiUpadhyaya:
> @ShrinidhiUpadhyaya That PR has been merged into canary but I could still reproduce the issue on canary branch.
> 
> image


Okay i will have look into it.

packages/core/theme/src/components/toggle.ts
macci001: @ShrinidhiUpadhyaya @wingkwong.
Doing this will resolve the issue.

wingkwong (changes requested):
As I mentioned in the original issue, the margin is not required when there is no label. If you just remove `me-2`, it will affect Switch with Label. In other word, margin is required when there is a label. This is what you need to solve, not just remove right margin. 

your pr:
image

prod:
image

ShrinidhiUpadhyaya:
> As I mentioned in the original issue, the margin is not required when there is no label. If you just remove `me-2`, it will affect Switch with Label. In other word, margin is required when there is a label. This is what you need to solve, not just remove right margin.
> 
> your pr: image
> 
> prod: image

Ohh okay, i missed reading that in the original issue. So can i just add left margin to the label `label: "relative text-foreground select-none ml-2",`. Is that fine?

**With Label:-**
image


**Without Label:-**
image

wingkwong:
don't use `ml`. this would fail in RTL. use `ms` instead which covers RTL case.

wingkwong (approved):

PR #3902, opened 2024-10-16
fix: onOpenChange triggering when modal opens

### LINKED ISSUE #3887: [BUG] - Modal onOpenChange only fires on close
### NextUI Version

2.4.8

### Describe the bug

I'm trying to (lazy) load a video in a nextui modal when it opens and pause it when it closes.
When I use the most basic modal example I can find, the `onOpenChange` event only fires when the modal closes.

_No response_

When running the most basic example, you will see the `onOpenChange` only fires when closing the modal:

```
import {Modal, ModalContent, ModalHeader, ModalBody, ModalFooter, Button, useDisclosure} from "@nextui-org/react";

  const {isOpen, onOpen, onOpenChange} = useDisclosure();

  return (
    <>
      <Button onPress={onOpen}>Open Modal</Button>
      <Modal
            isOpen={isOpen}
            onOpenChange={isOpen => {
                  console.log(isOpen) // only fires on close
                  onOpenChange()
            }}
      >
        <ModalContent>
          {(onClose) => (
            <>
              <ModalHeader className="flex flex-col gap-1">Modal Title</ModalHeader>
              <ModalBody>
                <p> 
                  Lorem ipsum dolor sit amet, consectetur adipiscing elit.
                  Nullam pulvinar risus non risus hendrerit venenatis.
                  Pellentesque sit amet hendrerit risus, sed porttitor quam.
                </p>
                <p>
                  Lorem ipsum dolor sit amet, consectetur adipiscing elit.
                  Nullam pulvinar risus non risus hendrerit venenatis.
                  Pellentesque sit amet hendrerit risus, sed porttitor quam.
                </p>
                <p>
                  Magna exercitation reprehenderit magna aute tempor cupidatat consequat elit
                  dolor adipisicing. Mollit dolor eiusmod sunt ex incididunt cillum quis. 
                  Velit duis sit officia eiusmod Lorem aliqua enim laboris do dolor eiusmod. 
                  Et mollit incididunt nisi consectetur esse laborum eiusmod pariatur 
                  proident Lorem eiusmod et. Culpa deserunt nostrud ad veniam.
                </p>
              </ModalBody>
              <ModalFooter>
                <Button color="danger" variant="light" onPress={onClose}>
                  Close
                </Button>
                <Button color="primary" onPress={onClose}>
                  Action
                </Button>
              </ModalFooter>
            </>
          )}
        </ModalContent>
      </Modal>
    </>
  );
}
```

### Expected behavior

I would expect the `onOpenChange` would also fire when the modal is fully opened and all content is loaded, but it does not?

### Screenshots or Videos

_No response_

### Operating System Version

macOS

### Browser

Chrome

sanuj21:
Closes #3887 

## 📝 Description
The changes in this PR makes the onOpenChange fire when modal opens, making it consistent.

- onOpenChange fires only when modal closes and not when it opens.
- onOpenChange in useDisclosure decides which action to fire on isOpen state.

## 🚀 New behavior

- onOpenChange fires on both states consistently.
- Added useEffect, so when isOpen changes to true, onOpenChange is called.
- onOpenChange in useDisclosure decides which action to fire on basis of isOpen argument, used to do isOpen state earlier.

## 📝 Additional Information
- onClose function in modal is basically coming from useOuterlayTriggerState, but onOpen is just a normal function. Thats why we need that useEffect to fire onOpenChange explicitly.
- Make sure to pass the isOpen variable when you are wrapping onOpenChange inside a function. Like this onOpenChange(isOpen).

### DIFF
--- a/packages/components/modal/src/use-modal.ts
+++ b/packages/components/modal/src/use-modal.ts
@@ -3,7 +3,7 @@ import type {HTMLMotionProps} from "framer-motion";
 
 import {AriaModalOverlayProps} from "@react-aria/overlays";
 import {useAriaModalOverlay} from "@nextui-org/use-aria-modal-overlay";
-import {useCallback, useId, useRef, useState, useMemo, ReactNode} from "react";
+import {useCallback, useId, useRef, useState, useMemo, ReactNode, useEffect} from "react";
 import {modal} from "@nextui-org/theme";
 import {
   HTMLNextUIProps,
@@ -128,6 +128,12 @@ export function useModal(originalProps: UseModalProps) {
       }
     },
   });
+
+  useEffect(() => {
+    if (isOpen) {
+      onOpenChange?.(isOpen);
+    }
+  }, [isOpen]);
 
   const {modalProps, underlayProps} = useAriaModalOverlay(
     {
--- a/packages/hooks/use-disclosure/src/index.ts
+++ b/packages/hooks/use-disclosure/src/index.ts
@@ -44,11 +44,14 @@ export function useDisclosure(props: UseDisclosureProps = {}) {
     onOpenPropCallbackRef?.();
   }, [isControlled, onOpenPropCallbackRef]);
 
-  const onOpenChange = useCallback(() => {
-    const action = isOpen ? onClose : onOpen;
+  const onOpenChange = useCallback(
+    (isOpen: boolean) => {
+      const action = isOpen ? onOpen : onClose;
 
-    action();
-  }, [isOpen, onOpen, onClose]);
+      action();
+    },
+    [isOpen, onOpen, onClose],
+  );
 
   return {
     isOpen: !!isOpen,
--- a/.changeset/early-ghosts-fix.md
+++ b/.changeset/early-ghosts-fix.md
@@ -1,0 +1,7 @@
+---
+"@nextui-org/modal": minor
+"@nextui-org/use-disclosure": minor
+---
+
+Added useEffect in useModal to fire onOpenChange when onOpen changes to true.
+In useDisclosure, onOpenChange now depends on arg onOpen, than state(onOpen).


packages/components/modal/__tests__/modal.test.tsx
```diff
@@ -3,6 +3,7 @@ import {render, fireEvent} from "@testing-library/react";
 import userEvent from "@testing-library/user-event";
 
 import {Modal, ModalContent, ModalBody, ModalHeader, ModalFooter} from "../src";
+import {Button} from "../../button/src";
```
wingkwong, line 6: button is already in devDependency. 

```
import {Button} from "@nextui-org/button";
```

packages/components/modal/__tests__/modal.test.tsx
```diff
@@ -91,6 +92,33 @@ describe("Modal", () => {
     expect(onClose).toHaveBeenCalled();
   });
 
+  test("should fire 'onOpenChange' callback when open button is clicked", async () => {
```
wingkwong, line 95: This test case passes in canary branch which means it doesn't cover the change you added in this PR.

wingkwong (changes requested):

packages/components/modal/src/use-modal.ts
```diff
@@ -129,6 +129,12 @@ export function useModal(originalProps: UseModalProps) {
     },
   });
 
+  useEffect(() => {
+    if (isOpen) {
+      onOpenChange?.(isOpen);
+    }
+  }, [isOpen]);
+
```
weeix, line 137: I came across this interesting resource on the React docs ([https://react.dev/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes](https://react.dev/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes)) that suggests alternative approaches to `useEffect` in some scenarios. It might be worth considering if it applies to this particular use case.

In the example provided, they achieve a similar outcome by storing information from previous renders.  I took a stab at adapting it here (untested):

```jsx
const [prevIsOpen, setPrevIsOpen] = useState(isOpen);
if (isOpen !== prevIsOpen) {
  setPrevIsOpen(isOpen);
  if (isOpen) {
    onOpenChange?.(isOpen);
  }
}
```

What do you think?

sanuj21: Thanks, It seems a very nice idea to avoid useEffect. I'll test and see if it works.

sanuj21: I tested, actually the onOpenChange is eventually changing the isOpen state. And we shouldn't directly update parent state inside child without useEffect.

I tried though and got the warning : "Cannot update a component (`App`) while rendering a different component (`NextUI.Modal`)".  

Let me know if I am missing something.

packages/components/modal/src/use-modal.ts
```diff
@@ -129,6 +129,12 @@ export function useModal(originalProps: UseModalProps) {
     },
   });
 
+  useEffect(() => {
+    if (isOpen) {
+      onOpenChange?.(isOpen);
+    }
+  }, [isOpen]);
```
ryo-manba, line 136: It looks like the issue might be with `useOverlayTriggerState` not properly triggering `onOpenChange`. Could you leave a comment about that? It would also be helpful if you could file an issue upstream if possible.

https://github.com/nextui-org/nextui/blob/8a33eabb2583202fcc8fbc33e8f2ed23bb45f1a4/packages/components/modal/src/use-modal.ts#L121-L130

ryo-manba: Is there a possibility that this change might cause `onOpenChange` to be called multiple times?

packages/hooks/use-disclosure/src/index.ts
```diff
@@ -44,11 +44,14 @@ export function useDisclosure(props: UseDisclosureProps = {}) {
     onOpenPropCallbackRef?.();
   }, [isControlled, onOpenPropCallbackRef]);
 
-  const onOpenChange = useCallback(() => {
-    const action = isOpen ? onClose : onOpen;
+  const onOpenChange = useCallback(
+    (isOpen: boolean) => {
+      const action = isOpen ? onOpen : onClose;
```
ryo-manba, line 49: Since `isOpen` is managed within `useDisclosure` whether it is controlled or uncontrolled, is it necessary to pass it as an argument?

packages/components/modal/src/use-modal.ts
sanuj21: > It looks like the issue might be with `useOverlayTriggerState` not properly triggering `onOpenChange`. Could you leave a comment about that? It would also be helpful if you could file an issue upstream if possible.
> 
> https://github.com/nextui-org/nextui/blob/8a33eabb2583202fcc8fbc33e8f2ed23bb45f1a4/packages/components/modal/src/use-modal.ts#L121-L130

Basically, `useOverlayTriggerState` fires `onOpenChange` when its internal `onClose` and `onOpen` are called. 
But in our case `onOpen` is coming from `useDisclosure` and `onClose`  is the internal function of `useOverlayTriggerState`  which is passed while mounting modal.
It has its own state `isOpen`, so its doesn't react automatically to outside state, as per my understanding.

> https://github.com/nextui-org/nextui/blob/8a33eabb2583202fcc8fbc33e8f2ed23bb45f1a4/packages/components/modal/src/use-modal.ts#L205

sanuj21: > Is there a possibility that this change might cause `onOpenChange` to be called multiple times?

Don't think so , because I am only calling when the `onOpen` changes to `true`.  However, one issue could be if we call the `onClose` fxn which is return by `useDisclosure`, than same problem will arise, `onOpenChange` will not be called.

packages/hooks/use-disclosure/src/index.ts
sanuj21: Yes, in this case its needed. Because when the  `onClose` is called its modifiying the internal state as I said earlier ( `isOpen` becomes `false`), but at this point the `isOpen` which is in `useDisclosure`  is `false`.
Initially the call was like this  `isOpen ? onClose : onOpen`, thats why it was working  when `onClose` used to get called. 
But now that will cause infinite render due to `useEffect`, thats why I reversed it and and added and argument.

packages/components/modal/src/use-modal.ts
sanuj21: Also, I thought if we can somehow expose the both the fxn of `useOverlayTriggerState`  in useDisclosure, than that will be very neat fix. But that is not possible unless the modal is mounted.

packages/hooks/use-disclosure/src/index.ts
ryo-manba: Isn't that the expected behavior?

1. You press the close button → `state.close` (internal state becomes `false`).
2. `onOpenChange` is called (at this point, `useDisclosure` state is still `true`).
3. Since `isOpen` is `true`, `onClose` gets called.
4. The `useDisclosure` state becomes `false` (syncing the internal state with useDisclosure).

Accepting an argument would be a breaking change, so it might be better to leave it as is for backward compatibility.

ryo-manba: Could you clarify in what cases the `useEffect` bug occurs?

packages/components/modal/src/use-modal.ts
ryo-manba: Oh, I see. Thank you for the clarification.

packages/hooks/use-disclosure/src/index.ts
sanuj21: > Isn't that the expected behavior?
> 
> 1. You press the close button → `state.close` (internal state becomes `false`).
> 2. `onOpenChange` is called (at this point, `useDisclosure` state is still `true`).
> 3. Since `isOpen` is `true`, `onClose` gets called.
> 4. The `useDisclosure` state becomes `false` (syncing the internal state with useDisclosure).
> 
> Accepting an argument would be a breaking change, so it might be better to leave it as is for backward compatibility.



> 1. You press the open button → external state becomes true. That will remount modal Component.
> 2. `useOverlayTriggerState` will re-run with isOpen `true`. After this, `useEffect` will be called and it will call the outer onOpenExchange(which was received as prop to modal), without passing argument((at this point, `useDisclosure` state is still `true`).
> 3. Since `isOpen` is `true`, `onClose` gets called(at this point, `useDisclosure` state is becomes `false`, but internal state is true).
> 4. Now, you press the close button → `state.close` (internal state becomes `false`).
> 5. `onOpenChange` is called (at this point, `useDisclosure` state is `false`), which means `onOpen` gets called but here `onClose` needs to get called.

Ya, i thought that it might be a breaking change, but we are also passing state while calling onOpenChange from `useOverlayTriggerState`

> https://github.com/nextui-org/nextui/blob/8a33eabb2583202fcc8fbc33e8f2ed23bb45f1a4/packages/components/modal/src/use-modal.ts#L124

ryo-manba: Oh, so the `useEffect` added in `useModal` is the issue. Hmm, with the current implementation, it might introduce new bugs. Could you consider exploring alternative implementation methods?

ryo-manba: What do you think? @jrgarciadev

chirokas:
- The current commit causes `onOpenChange` to trigger twice when in an uncontrolled state.
- Should we treat [this](https://github.com/nextui-org/nextui/issues/3887) as incorrect usage rather than a bug? 🤔 See [here](https://stackblitz.com/edit/vitejs-vite-sufbgs?file=src%2FApp.tsx).

ryo-manba:
@chirokas 
Thanks for the comment! In basic use cases, it seems that the approach in your example would work well. However, I think this behavior of the `Modal` is also a bug.  I'll be sure to pay attention to `onChange` in `useDisclosure` when making fixes. Thank you!

PR #3912, opened 2024-10-17
fix(theme): removed pseudo cancel btn from input

### LINKED ISSUE #3907: [BUG] - Two(instead of one) clear icons pop up from `<Input />`
### NextUI Version

2.4.8

### Describe the bug

Two clear icons pop up.

스크린샷 2024-10-17 오전 9 53 38

_No response_

```tsx
import { Input } from '@nextui-org/react'

<Input isClearable type="search" variant="underlined" />
```

### Expected behavior

I expected only a single clear icon appear.

### Screenshots or Videos

_No response_

### Operating System Version

macOS

### Browser

Chrome

sanuj21:
Closes #3907 

## 📝 Description
Pseudo Cancel Button used to come with input element in chrome unintentionally when the input type is search.

Pseudo Cancel Button comes with input element in chrome when the type is search.
Screenshot from 2024-10-18 04-38-40

## 🚀 New behavior
Removed the cancel button which was unintentional.
Screenshot from 2024-10-18 04-41-06

No

## Summary by CodeRabbit

- **New Features**
	- Removed the pseudo cancel button from input components for a cleaner appearance.

- **Bug Fixes**
	- Addressed the issue of the pseudo cancel button display in input components as noted in issue #3907.

### DIFF
--- a/packages/core/theme/src/components/input.ts
+++ b/packages/core/theme/src/components/input.ts
@@ -46,6 +46,7 @@ const input = tv({
       "data-[has-end-content=true]:pe-1.5",
       "file:cursor-pointer file:bg-transparent file:border-0",
       "autofill:bg-transparent bg-clip-text",
+      "input-search-cancel-button-none",
     ],
     clearButton: [
       "p-2",
--- a/packages/core/theme/src/components/input.ts
+++ b/packages/core/theme/src/components/input.ts
@@ -46,7 +46,6 @@ const input = tv({
       "data-[has-end-content=true]:pe-1.5",
       "file:cursor-pointer file:bg-transparent file:border-0",
       "autofill:bg-transparent bg-clip-text",
-      "input-search-cancel-button-none",
     ],
     clearButton: [
       "p-2",
@@ -191,7 +190,7 @@ const input = tv({
     },
     isClearable: {
       true: {
-        input: "peer pe-6",
+        input: "peer pe-6 input-search-cancel-button-none",
         clearButton: "peer-data-[filled=true]:opacity-70 peer-data-[filled=true]:block",
       },
     },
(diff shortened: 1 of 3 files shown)


wingkwong (changes requested):
it's still here. please double check.

```
<Input isClearable type="search" variant="underlined" />
```

image

sanuj21:
@wingkwong I checked it again its working while running with storybook.
However, I tried to replicate the same scenerio using a test project and than linking to the nextui using pnpm link.
But while linking, it creates a symlink of folder nextui in node_modules, but we import from @nextui-org right.
Is it an alias issue, or its a build folder ? 
Or is there any specific way to setup this ?

wingkwong:
I also tested in storybook only. Which browser were you testing in?

packages/core/theme/src/components/input.ts
```diff
@@ -46,6 +46,7 @@ const input = tv({
       "data-[has-end-content=true]:pe-1.5",
       "file:cursor-pointer file:bg-transparent file:border-0",
       "autofill:bg-transparent bg-clip-text",
+      "input-search-cancel-button-none",
```
wingkwong, line 49: i think we should hide it only when isClearable is true. Otherwise, it will break the existing behaviour for the false case.

sanuj21: But lets say, isClearable is false. Then it will still have a clear button in chrome(when the type is search), which shouldn't be the case right ?
Also I tested in chrome, did you build it ? Because the changes are not getting reflected without build.

wingkwong: - if `isClearable` is `false` and `type=search`, then it should show webkit-search-cancel-button because we don't want to break the native behaviour. 
- if `isClearable` is `true` and `type=search`, then it should only show NextUI Close button (since it supports onClear handle) and hide webkit-search-cancel-button one.
- True. The theme package needs to be built.

wingkwong (approved):

PR #3968, opened 2024-10-31
refactor(docs): breadcrumbs dx

wingkwong:
Closes #


apps/docs/content/components/breadcrumbs/custom-items.raw.jsx
```diff
@@ -0,0 +1,68 @@
+import {
```
wingkwong, line 1: seems something wrong with import with multiple lines

apps/docs/components/docs/components/code-demo/utils.ts
```diff
@@ -2,7 +2,7 @@ import {get} from "lodash";
 
 import {FileCode} from "./types";
 
-const importRegex = /^(import)\s(?!type(of\s|\s)(?!from)).*?$/gm;
+const importRegex = /^(import\s+(?!type\s+\{)[\s\S]*?;)/gm;
```
winchesHe, line 5: `(?!type(of\s|\s)(?!from))` change to `(?!type\s+\{)` why need `\{`

winchesHe (approved):

apps/docs/components/docs/components/code-demo/utils.ts
wingkwong: cover `import type { Selection } from "@nextui-org/react";` case

PR #4008, opened 2024-11-07
fix: react 19 as peer dep

jrgarciadev:
Closes #


follow-prince:
🚀 Great job on the React 19 upgrade, @jrgarciadev ! Thanks for tackling all those changes to keep everything compatible. This will make adopting React 19 so much smoother. Awesome work! 👏🎉


This pull request updates the peer dependencies for various components to include support for React 19. The changes ensure compatibility with both React 18 and the upcoming React 19 release candidate.

PR #4112, opened 2024-11-18
fix(system): allow Promise<void> for provider navigate prop

CanRau:
Redid https://github.com/nextui-org/nextui/pull/4098 to get rid of all those unrelated changes

## 📝 Description

At least TanStack Router's `router.navigate` returns `Promise<void>` so if this PR won't break anything would be nice to support both return types ❤️ 

```
Type '(to: any, options: any) => Promise<void>' is not assignable to type '(path: string) => void'.
```

The `to: any, options: any` is another topic I think

## 🚀 New behavior

No type errors

Shouldn't break anything I think 🤔 

## Summary by CodeRabbit

- **New Features**
	- Enhanced the `navigate` function to support asynchronous operations by allowing it to return a promise.

This change improves flexibility in navigation handling within the application.


wingkwong (changes requested):
please add a changeset

CanRau:
like so?

wingkwong (changes requested):
not pre.json. Please see [here](https://github.com/changesets/changesets/blob/main/docs/adding-a-changeset.md). It should just patch on `@nextui-org/system`

CanRau:
I see, like this? 😇

CanRau:
That's why it's returning `void | Promise<void>` so it should support both right?

You mean like:
```
navigate(path: Href, routerOptions: RouterOptions | undefined) => void | Promise<void>?
```

wingkwong:
Actually this `navigate` will be passing to react-aria's router. Their navigate return type is `void`. And the parameters got changed so I agree we should update that part only.

see https://react-spectrum.adobe.com/react-spectrum/routing.html#tanstack-router for the usage.

wingkwong:
Closing this PR.

- After revisiting this PR again, the root cause actually is not the return type but the parameters. There is no need to change the return type here since such assignment is allowed. If you are interested, you can see this [playground](https://www.typescriptlang.org/play/?#code/CYUwxgNghgTiAEEQBd4DcD2BLYAxAdgFzwAUAlPALwB862wA3ALABQoksCSqADjBgFssAZxAFi5KrQAK-IaIA8mHNWYtWyvPirw+gkWPwN4AehPwA7hhgBrYaz3zDOzQWNn4AMyhYIwoA).
- For the parameter, I've made a PR in https://github.com/nextui-org/nextui/pull/4163. This should solve the reported typing issue.

PR #4156, opened 2024-11-28
fix: menu item classNames not work

### LINKED ISSUE #4119: [BUG] - DropdownItem ignore variants
### NextUI Version

~~2.4.2~~ 2.4.8

### Describe the bug

I created custom component <DropdownItem /> which extends variants ( extendVariants used ). When I use this component inside <DropdownSection /> everything is rendered correctly but when this component is a direct child of <DropdownMenu />   then <DropdownItem /> get variants but styles are ignored.

_No response_

1. Create custom Dropdown items for example
```
const DropdownItem = extendVariants(NextDropdownItem, {
  variants: {
    variant: {
      solid: {
        base: ['rounded-none'],
      },
    },
    size: {
      md: {
        base: 'py-4',
      }
    },
  },
  defaultVariants: {
    variant: 'solid',
    size: 'md',
  },
});
```
2. Put this component inside <DropdownMenu />

### Expected behavior

styles from custom component should be applied 

### Screenshots or Videos

_No response_

### Operating System Version

macOS

### Browser

Chrome

winchesHe:
close #4119

## Summary by CodeRabbit

- **Bug Fixes**
	- Resolved issues with class names not functioning correctly for menu items in the menu component.

- **Tests**
	- Added new test cases to ensure proper functionality of class names for `MenuItem` components, verifying custom class name application and precedence handling.

- **Refactor**
	- Improved class name handling in the `Menu` component, allowing for more dynamic and flexible assignment of class names based on item properties.


.changeset/brave-trains-wave.md
```diff
@@ -0,0 +1,5 @@
+---
+"@nextui-org/menu": patch
+---
+
+Fix menu item classNames not work
```
wingkwong, line 5: include issue number

packages/components/menu/src/menu.tsx
```diff
@@ -47,11 +47,12 @@ function Menu<T extends object>(props: Props<T>, ref: ForwardedRef<HTMLUListElem
           hideSelectedIcon,
           ...item.props,
         };
+        const mergedItemClasses = itemProps.classNames ?? itemClasses;
```
wingkwong, line 50: may double check if the expected behaviour is "merge" or "override". The naming suggest it's "merge" but the logic is "override". If it's the latter, we may also need to point that out in documentation.

wingkwong (commented):
may double check if this PR is really related to the linked issue or not

packages/components/menu/src/menu.tsx
winchesHe: updated the documentation

wingkwong (commented):
what about just call it `menuItemClasses` instead of `mergedItemClasses`? since either one will be used instead of merging them. The naming makes me feel `cn` is missing here while it's not.

.changeset/brave-trains-wave.md
```diff
@@ -0,0 +1,5 @@
+---
+"@nextui-org/menu": patch
```
jrgarciadev, line 2:
```suggestion
"@nextui-org/menu": patch
"@nextui-org/theme": patch
```

PR #4157, opened 2024-11-28
fix: should not export list item internal variants type

winchesHe:
Closes # 

## 📝 Description

Related https://github.com/nextui-org/nextui/pull/4105


packages/components/menu/src/menu.tsx
```diff
@@ -71,10 +71,7 @@ function Menu<T extends object>(props: Props<T>, ref: ForwardedRef<HTMLUListElem
   );
 }
 
-export type MenuProps<T extends object = object> = Omit<
```
winchesHe, line 74: I found type `Props<T>` don't have these types export so i change to Omit in the MenuItemVariantProps

PR #4171, opened 2024-11-29
fix(docs): typecheck errors

wingkwong:
Closes # 

## 📝 Description

image

## 🚀 New behavior

image


apps/docs/content/components/alert/custom-styles.raw.tsx
```diff
@@ -3,73 +3,66 @@ import type {AlertProps} from "@nextui-org/react";
...
+      {children}
+    </Alert>
+  );
+};
 
 CustomAlert.displayName = "CustomAlert";
 
 export default function App() {
-  const colors = ["default", "primary", "secondary", "success", "warning", "danger"];
+  const colors = ["default", "primary", "secondary", "success", "warning", "danger"] as any;
```
jrgarciadev, line 57: what about `as const` ?

PR #4215, opened 2024-12-03
fix(theme): support description with a single space using zero-width …

iqingting:
## 📝 Description

Support `description=" "` to maintain a consistent layout when some form controls have descriptions and others do not.

CleanShot 2024-12-03 at 17 20 55@2x

## 🚀 New behavior

CleanShot 2024-12-03 at 17 19 51@2x


packages/components/date-input/src/date-input-group.tsx
```diff
@@ -64,7 +64,9 @@ export const DateInputGroup = forwardRef<"div", DateInputGroupProps>((props, ref
         {isInvalid && errorMessage ? (
           <div {...errorMessageProps}>{errorMessage}</div>
         ) : description ? (
-          <div {...descriptionProps}>{description}</div>
+          <div {...descriptionProps}>
+            {description === " " ? <span>&#8203;</span> : description}
```
Peterl561, line 68: 1. It's not immediately clear for users that `" "` causes a span to appear where normally you'd expect text
2. Doesn't seem ideal for accessibility

Peterl561 (changes requested):
Aligning inputs is a styling problem that can be fixed through CSS (e.g. `align-items: flex-start;` or `items-start` class in tailwind). It feels unnecessary to make this change to input description's behavior.

packages/components/date-input/src/date-input-group.tsx
Peterl561: Also this won't work if neighboring input have multiple lines of description.

iqingting:
> Aligning inputs is a styling problem that can be fixed through CSS (e.g. `align-items: flex-start;` or `items-start` class in tailwind). No need to make this change to input description's behavior.

It's not just a style issue; the problem is that when no description is provided, the error messages disrupt the layout. For example, if the form is in a modal, the current behavior causes the modal's height to change as the form's validation status updates.

https://stackblitz.com/edit/vitejs-vite-mj4jjg?file=src%2FApp.tsx


https://github.com/user-attachments/assets/4ea978b6-926a-40b2-82e8-cb520b73bf7b

Peterl561:
> It's not just a style issue; the problem is that when no description is provided, the error messages disrupt the layout. For example, if the form is in a modal, the current behavior causes the modal's height to change as the form's validation status updates.

That is a definitely not great UI-wise, but I think solving it with some CSS is preferable to the implementation proposed in this PR, for the reasons I mentioned in my previous comments.

iqingting:
> > It's not just a style issue; the problem is that when no description is provided, the error messages disrupt the layout. For example, if the form is in a modal, the current behavior causes the modal's height to change as the form's validation status updates.
> 
> 
> 
> That is definitely not great UI-wise, but I think solving it with some CSS is preferable to the implementation proposed in this PR, for the reasons I mentioned in my previous comments.

1. Why do you think don't disrupt the form layout when validate status change is not great UI-wise.
2. `flex-start` doesn't work unless you set a static height for every form control, if you agree that preventing the form layout is a reasonable requirement.

Peterl561:
> 1. Why do you think don't disrupt the form layout when validate status change is not great UI-wise.
> 2. `flex-start` doesn't work unless you set a static height for every form control, if you agree that preventing the form layout is a reasonable requirement.

Sorry for the confusion. I meant that the form height changing is not great. I do think it would be nice if we could come up with a generalized fix, but I don't have any decent ideas.

The actual CSS implementation will of course depend on your requirements. If you are interested in discussing a specific CSS solution for your situation, you could try asking in Discord.

PR #4219, opened 2024-12-04
fix(select): hideEmptyContent API

### LINKED ISSUE #2422: [BUG] - Select component won't open listbox if list is empty
## Description
The select components uses the listBox components inside it, and the listBox component has `hideEmptyContent` attribute which controls if the listBox should be shown if the content is empty, I've tried to pass `false` to the listboxProps attribute of the select component but still clicking on the select component will do nothing.

After going through the select component source code I found [this line](https://github.com/nextui-org/nextui/blob/0cc8c79c53ae416586117540e05801631080e52e/packages/hooks/use-aria-multiselect/src/use-multiselect-state.ts#L79) which controls the clicking behavior and it looks like it forcefully won't open the listBox if the list is empty.

## Expected
Being able to control the clicking behavior of the select component when the list is empty.

## Suggested
Adding an extra attribute to the select component that gets passed to the useMultiSelect hook to control the clicking behavior.

Peterl561:
Closes #2422

## 📝 Description

- added `hideEmptyContent` to `Select`, which toggles whether to show dropdown when there is no content
- note that it defaults to `false` in `useSelect`, which is different from current behavior
- added `Select` storybook for `hideEmptyContent`

see video below

## 🚀 New behavior

old on the left, new on the right

https://github.com/user-attachments/assets/df04818f-1bc7-4d53-8e55-59741c671658

No

## 📝 Additional Information

N/A

## Summary by CodeRabbit

- **New Features**
  - Introduced `hideEmptyContent` property in the `Select` component, allowing users to control the visibility of empty content in the dropdown.
  - Added examples in the storybook demonstrating the behavior of the `Select` component with empty content scenarios.

- **Documentation**
  - Enhanced documentation for the `Select` component, detailing the new `hideEmptyContent` property and updating existing property descriptions for clarity.

- **Tests**
  - Added new test cases to verify the behavior of the `Select` component when using the `hideEmptyContent` property.

### DIFF
--- a/packages/components/select/src/use-select.ts
+++ b/packages/components/select/src/use-select.ts
@@ -166,6 +166,11 @@ export type UseSelectProps<T> = Omit<
      * @default undefined
      */
     isVirtualized?: boolean;
+    /**
+     * Whether the listbox should be openable when there are no items.
+     * @default true
+     */
+    hideEmptyContent?: boolean;
   };
 
 export function useSelect<T extends object>(originalProps: UseSelectProps<T>) {
@@ -209,6 +214,7 @@ export function useSelect<T extends object>(originalProps: UseSelectProps<T>) {
     onClose,
     className,
     classNames,
+    hideEmptyContent = true,
     ...otherProps
   } = props;
 
@@ -263,6 +269,7 @@ export function useSelect<T extends object>(originalProps: UseSelectProps<T>) {
     isDisabled: originalProps.isDisabled,
     isInvalid: originalProps.isInvalid,
     defaultOpen,
+    hideEmptyContent,
     onOpenChange: (open) => {
       onOpenChange?.(open);
       if (!open) {
--- a/packages/components/select/src/use-select.ts
+++ b/packages/components/select/src/use-select.ts
@@ -167,8 +167,8 @@ export type UseSelectProps<T> = Omit<
      */
     isVirtualized?: boolean;
     /**
-     * Whether the listbox should be openable when there are no items.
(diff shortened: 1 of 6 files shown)


Peterl561:
Not sure if the doc changes need to be a separate PR or not, as recommended by the PR template. Let me know and I'll open another PR for it.

packages/components/select/src/use-select.ts
```diff
@@ -166,6 +166,11 @@ export type UseSelectProps<T> = Omit<
      * @default undefined
      */
     isVirtualized?: boolean;
+    /**
+     * Whether the listbox should be openable when there are no items.
+     * @default true
+     */
+    hideEmptyContent?: boolean;
```
jrgarciadev, line 173:
```suggestion
    /**
     * Whether the listbox be prevented from opening when there are no items
     * @default false     
     */
    hideEmptyContent?: boolean;
```

packages/components/select/src/use-select.ts
```diff
@@ -209,6 +214,7 @@ export function useSelect<T extends object>(originalProps: UseSelectProps<T>) {
     onClose,
     className,
     classNames,
+    hideEmptyContent = true,
```
jrgarciadev, line 217:
```suggestion
    hideEmptyContent = false,
```

jrgarciadev (changes requested):
Great addition, thanks @Peterl561 please just fix the comments

jrgarciadev (approved):

PR #3346, opened 2024-06-27
fix(pagination): active element incorrect position on first load

arindam1997007:
Closes #3151 

## 📝 Description

Fixed Active Element positionon on first load, when pagination is rendered initially with `display:hidden`

## ⛳️ Current behavior (updates)

On first load, instead of left arrow button,  **Page 1** active element is displayed

## 🚀 New behavior

Active Element position is fixed even on first load, using IntersectionObserver.

## 💣 Is this a breaking change (Yes/No):

No

## 📝 Additional Information

According to code submitted by users which can be reproduced, issue occurs if pagination component is rendered using `display:hidden` for the first time, and then switching to `display:block`.

Using **IntersectionObserver**, checking for display of pagination componenet in the viewport, and then running the required functions.

https://github.com/nextui-org/nextui/assets/23236141/9c5d1a2e-a06c-49b5-beba-8711b1beca0f

## Summary by CodeRabbit

- **Bug Fixes**
  - Resolved an issue with the Pagination component where the active element's position was incorrect when `display: hidden` was applied.

- **New Features**
  - Introduced a new `SubTabs` component to manage tabs and content display within the pagination story.
  - Added multiple new SVG icon components: `ChevronDownIcon`, `PlusIcon`, `SearchIcon`, and `VerticalDotsIcon`.
  - Introduced a `TableData` component with advanced table features like filtering, sorting, pagination, and customizable column visibility.

- **Enhancements**
  - Improved the `usePagination` function with better scrolling behavior and visibility checks.
  - Enhanced table data management with predefined columns, status options, and user information for display purposes.

PR #4222, opened 2024-12-04
Fix/pagination cursor hidden position

### LINKED ISSUE #3151: [BUG] Pagination left arrow shows page 1 on the first load
### NextUI Version

2.4.1

### Describe the bug

Pagination component is showing the left arrow button as page number 1.
This only happens for the first time when the page is loaded. If I click a different page, the left arrow will be displayed correctly. 

image

_No response_

I'm not able to reproduce this in Stackblitz using static data. What I'm doing is:
- Load data from a Graphql endpoint. 
- Calculate the number of pages needed based on the number of items loaded, and a `rowsPerPage` parameter:

```
  const [rowsPerPage, setRowsPerPage] = useState<number>(10);
  const [page, setPage] = useState<number>(1);

  const pages = useMemo(
    () => Math.ceil(items.length / rowsPerPage),
    [items.length, rowsPerPage]
  );
```

- Pass the page number to the Pagination component.

```
      <div className="py-2 px-2 flex items-center justify-center">
        <Pagination
          isCompact
          showControls
          showShadow
          color="primary"
          page={page}
          total={pages}
          onChange={(page) => setPage(page)}
        />
      </div>
```

### Expected behavior

Left arrow should be correctly displayed. 

### Screenshots or Videos

_No response_

### Operating System Version

macOS

### Browser

Chrome

Peterl561:
Closes #3151
Alternative to #3346 

## 📝 Description

- when `Pagination` first loads, if it is not visible, offset of the initial page node will be 0
- if animation is enabled, this causes the cursor to jump to 0, which blocks the left control button if it is enabled
- to rectify this, do not show cursor until `Pagination` component is visible in the DOM
- this effect can be achieved with `IntersectionObserver`

No

## 📝 Additional Information

- initial page still doesn't work for asynchronously loaded page totals
- should be addressed as a separate fix, probably some sort of loading state or something

### DIFF
--- a/.changeset/tidy-parents-cheat.md
+++ b/.changeset/tidy-parents-cheat.md
@@ -1,0 +1,5 @@
+---
+"@nextui-org/pagination": patch
+---
+
+fix pagination cursor position on first load
--- a/packages/components/pagination/__tests__/pagination.test.tsx
+++ b/packages/components/pagination/__tests__/pagination.test.tsx
@@ -79,4 +79,74 @@ describe("Pagination", () => {
 
     expect(cursor).toHaveAttribute("aria-hidden", "true");
   });
+
+  it("the pagination cursor should not have data-moving attribute before intersection", () => {
+    // save callback and options to later emulate intersection event
+    let intersectCallback: IntersectionObserverCallback | undefined;
+    let intersectOptions: IntersectionObserverInit | undefined;
+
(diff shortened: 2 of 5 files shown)


jrgarciadev (approved):
Thanks @Peterl561 🙏🏻, you're doing a great job!

PR #4257, opened 2024-12-07
fix: disable number change in case of scrolling

ankit-ml12:
Closes  https://github.com/nextui-org/nextui/issues/4238

## 📝 Description
added a prop disableScrollChange to disable change in the input when the type is number

disable change in number on scrolling with mouse wheel or touchpad, since it can change value without user noticing when in scrollable container

## 🚀 New behavior

prop added to disable if needed

No
## Summary by CodeRabbit

- **New Features**
	- Introduced an optional property `disableScrollChange` to enhance input control during scrolling.
	- Added functionality to prevent value changes when the input is scrolled, improving user experience.

### DIFF
--- a/packages/components/input/src/use-input.ts
+++ b/packages/components/input/src/use-input.ts
@@ -82,6 +82,10 @@ export interface Props<T extends HTMLInputElement | HTMLTextAreaElement = HTMLIn
    * React aria onChange event.
    */
   onValueChange?: (value: string) => void;
+  /**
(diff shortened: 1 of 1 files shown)


wingkwong:
1. please don't work on assigned issue
2. we plan to have NumberField so we shouldn't add a new prop to input.

ankit-ml12:
Ok thanks @wingkwong

PR #4264, opened 2024-12-07
fix: missing react-types package added

jrgarciadev:
Closes # 

## Summary by CodeRabbit

- **Bug Fixes**
	- Resolved a missing dependency issue by adding `react-types` to enhance functionality and type safety.

- **New Features**
	- Updated the dependency `@react-types/shared` to version `3.26.0` across multiple packages, improving overall functionality and compatibility.

### DIFF
--- a/packages/core/system/package.json
+++ b/packages/core/system/package.json
@@ -61,6 +61,7 @@
     "@react-aria/overlays": "3.23.4",
     "@react-aria/utils": "3.26.0",
     "@react-stately/utils": "3.10.4",
-    "@react-types/datepicker": "3.8.3"
+    "@react-types/datepicker": "3.8.3",
+    "@react-types/shared": "3.25.0"
   }
 }
--- a/packages/core/system/package.json
+++ b/packages/core/system/package.json
@@ -62,6 +62,6 @@
     "@react-aria/utils": "3.26.0",
     "@react-stately/utils": "3.10.4",
     "@react-types/datepicker": "3.8.3",
-    "@react-types/shared": "3.25.0"
+    "@react-types/shared": "3.26.0"
   }
 }
--- a/.changeset/gorgeous-mice-think.md
+++ b/.changeset/gorgeous-mice-think.md
@@ -1,0 +1,5 @@
+---
+"@nextui-org/system": patch
+---
+
+Fix, mising react-types package added to the system package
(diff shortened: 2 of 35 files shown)


packages/core/system/package.json
```diff
@@ -61,6 +61,7 @@
     "@react-aria/overlays": "3.23.4",
     "@react-aria/utils": "3.26.0",
     "@react-stately/utils": "3.10.4",
-    "@react-types/datepicker": "3.8.3"
+    "@react-types/datepicker": "3.8.3",
+    "@react-types/shared": "3.25.0"
```
AnYiEE, line 65: Why not use the latest 3.26.0?
To avoid muti-vervisn issue like https://github.com/nextui-org/nextui/issues/4243

jrgarciadev: because latest versions of `react-aria` are still not supported/migrated/tested

AnYiEE: @jrgarciadev The focus is on the mixed use of multiple versions. Although NextUI specifies a certain version, such as 3.25.0 of `@react-types/shared`, it will actually install 3.26.0 of `@react-types/shared`, and the types provided by both are not completely the same. You can try creating a new folder and initializing a project, then only install the `@nextui-org/react` package, and search the lockfile. You will see that some packages of `@react-aria/*` are using 3.26.0 of `@react-types/shared`. When imports some types from `@nextui-org/react`, such as `SortDescriptor`, for me (I use pnpm), the imported version is 3.25.0, while the component props only accept other versions, and the `SortDescriptor` does not overlap, leading to type errors.

```ts
// 3.26.0
export interface SortDescriptor {
  /** The key of the column to sort by. */
  column: Key,
  /** The direction to sort by. */
  direction: SortDirection
}

// 3.25.0
export interface SortDescriptor {
  /** The key of the column to sort by. */
  column?: Key,
  /** The direction to sort by. */
  direction?: SortDirection
}
```

wingkwong: @AnYiEE just fyi. I'm currently working on it on https://github.com/nextui-org/nextui/pull/4258.

jrgarciadev:
@AnYiEE fixed, thanks for the suggestions, just checked it out and everything works as expected with the `v3.26.0`

Issue #4223, opened 2024-12-04
[BUG] - Buttons wrapped with Link not working on mobile canary


christophmeise:
### NextUI Version

2.5.0

### Describe the bug

<Button> that are wrapped with a <Link> are not clickable on mobile devices on latest canary 2.5.0

On click, nothing happens.

### Steps to Reproduce the Bug or Issue

Example:

<Link href="/">
<Button color="default" variant="light">
click
</Button>
</Link>

### Expected behavior

Click should work mobile

### Operating System Version

Mobile

### Browser

Chrome

christophmeise:
Probably related to: https://github.com/nextui-org/nextui/issues/4210

@macci001

miljkovicjovan:
Experiencing the same right now

christophmeise:
> Experiencing the same right now (tested on Safari and Brave)

My workaround is to only use <Button as={Link}> for now.

This is definitely a bug though.

wingkwong:
Please upgrade to the latest version.

PR #4284, opened 2024-12-08
refactor: button & card ripple

sudongyuer:
Closes # 

## 📝 Description

Refactor Button & Card ripple

In the latest version of react-aria they added the pointer events to the usePress hook so we no longer need to hack this in the components that use ripple

https://react-spectrum.adobe.com/react-aria/examples/ripple-button.html

## 🚀 New behavior

https://github.com/user-attachments/assets/619d65dd-d3fd-4fdc-be82-bba9705d53c6

## Summary by CodeRabbit

- **New Features**
  - Enhanced button and card components to support press events, improving interaction clarity and accessibility.
  
- **Bug Fixes**
  - Updated event handling for ripple effects to ensure they trigger correctly under specified conditions.

- **Refactor**
  - Refined internal logic for button and card components to distinguish between click and press events, streamlining event handling.

- **Documentation**
  - Updated method signatures to reflect changes in event handling for better clarity.

### DIFF
--- a/.changeset/rude-cobras-grab.md
+++ b/.changeset/rude-cobras-grab.md
@@ -1,0 +1,7 @@
+---
+"@nextui-org/button": patch
+"@nextui-org/ripple": patch
+"@nextui-org/card": patch
+---
+
+Refactor Button & Card Ripple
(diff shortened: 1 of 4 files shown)


macci001:
The PR fixes the ripple issue on touch devices, which we saw after upgrading the RA.
Also side-note: seems like after moving to onPress from onClick, we can also get rid of this issue #4223

jrgarciadev (approved):

PR #4285, opened 2024-12-09
[DRAFT] Feat(table): virtualization

vinroger:
Closes #


wingkwong (commented):
1. please add a changeset
2. why is the default value of `isVirtualized`  `undefined` instead of `false`?

vinroger:
@wingkwong 
> 1. please add a changeset 
Will do. Thank you for reminding!

> 2. why is the default value of `isVirtualized`  `undefined` instead of `false`?
I am following Junior's edit on autocomplete props component. https://github.com/nextui-org/nextui/pull/4094/files#diff-f28f2caca204e466c560d07fd4d4ca22c4618f1430ea13b915548db84cf7aad8

My thoughts: this is because even if the user does not provide the isVirtualized prop, we enable virtualization if the list item is more than 50.

packages/components/table/src/table-row-group.tsx
```diff
@@ -11,7 +11,7 @@ export interface TableRowGroupProps extends HTMLHeroUIProps<"thead"> {
   classNames?: ValuesType["classNames"];
 }
 
-const TableRowGroup = forwardRef<"thead", TableRowGroupProps>((props, ref) => {
+const TableRowGroup = forwardRef<any, TableRowGroupProps>((props, ref) => {
```
wingkwong, line 14: any better type than `any`?

wingkwong (approved):

packages/components/table/src/table.tsx
```diff
@@ -30,6 +36,10 @@ const Table = forwardRef<"table", TableProps>((props, ref) => {
     ref,
   });
 
+  const {isVirtualized, rowHeight = 40, maxTableHeight = 600} = props;
+
+  const shouldVirtualize = values.collection.size > 50 || isVirtualized;
+
```
jrgarciadev, line 42: @vinroger let's add a prop to make this component virtualized (at least for v.2.7.0) then we can apply this automatically, but only when we are sure it is stable, do the same with the `select`

jrgarciadev (changes requested):

packages/components/table/src/table.tsx
wingkwong: currently `isVirtualized` is the prop to make it virtualized. in select, there is also such logic (i.e. virtualized automatically when size > 50)

PR #4296, opened 2024-12-09
fix: autofocus the input-otp when auto-focus prop is passed

### LINKED ISSUE #4250: [BUG] - OTP Autofocus doesn't work
### NextUI Version

2.6.3

### Describe the bug

So my OTP componenet is loaded when the user inserted their e-mail, Then the OTP appears, the autoFocus prop is set to true but theres no focus when the componnent appears. 

_No response_

<InputOtp
              length={4}
              size="lg"
              autoFocus
            />

and you can jsut insert it or you can set a button to hide/show the OTP and then you can see the problem

### Expected behavior

Autofocus to the first box

### Screenshots or Videos

https://github.com/user-attachments/assets/13a4e3b9-db2e-4d76-8fc2-ce5b144244a7

### Operating System Version

Windows

### Browser

Chrome

macci001:
Closes #4250

## 📝 Description

Fixes the autofocus functionality in input-otp component## ⛳️ Current behavior (updates)

## Current behavior

https://github.com/user-attachments/assets/bddd90c6-d83f-4ecc-bf7d-4037ed92dd87

## 🚀 New behavior

https://github.com/user-attachments/assets/d77c19bc-cbc3-4dd4-9f5a-31eed5d3c40d

### DIFF
--- a/packages/components/input-otp/__tests__/input-otp.test.tsx
+++ b/packages/components/input-otp/__tests__/input-otp.test.tsx
@@ -168,6 +168,21 @@ describe("InputOtp Component", () => {
     expect(onComplete).toHaveBeenCalledTimes(2);
     expect(onComplete).toHaveBeenCalledWith("1234");
   });
+
+  it("should autofocus when autofocus prop is passed.", async () => {
+    // eslint-disable-next-line jsx-a11y/no-autofocus
+    render(<InputOtp autoFocus length={4} />);
+    const segments = screen.getAllByRole("presentation");
+
+    expect(segments[0]).toHaveAttribute("data-focus", "true");
+  });
+
+  it("should not autofocus when autofocus prop is not passed.", async () => {
+    render(<InputOtp length={4} />);
+    const segments = screen.getAllByRole("presentation");
+
+    expect(segments[0]).not.toHaveAttribute("data-focus", "true");
+  });
 });
 
 describe("InputOtp with react-hook-form", () => {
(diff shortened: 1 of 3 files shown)


packages/components/input-otp/src/use-input-otp.ts
```diff
@@ -230,6 +230,7 @@ export function useInputOtp(originalProps: UseInputOtpProps) {
         ref: inputRef,
         name: name,
         value: value,
+        autoFocus: originalProps.autoFocus,
```
wingkwong, line 233: why we need to extract from `originalProps` instead of just using `autoFocus` in L107?

macci001: Done in: fffa4bbc2b6baba16f51b970c49d439833b988f3

packages/components/input-otp/__tests__/input-otp.test.tsx
```diff
@@ -168,6 +168,21 @@ describe("InputOtp Component", () => {
     expect(onComplete).toHaveBeenCalledTimes(2);
     expect(onComplete).toHaveBeenCalledWith("1234");
   });
+
+  it("should autofocus when autofocus prop is passed.", async () => {
```
wingkwong, line 172: async is not required

wingkwong (changes requested):
please also add `autoFocus` in https://nextui.org/docs/components/input-otp#inputotp-props

wingkwong (approved):

PR #4309, opened 2024-12-10
Export useDisclosure with  `@nextui-org/drawer`

omarshehab221:
Closes # 

## 📝 Description

Export useDisclosure with `@nextui-org/drawer`

the useDisclosure can currently be only imported from `@nextui-org/modal` or `@nextui-org/use-disclosure`

## 🚀 New behavior

useDisclosure is now exported with `@nextui-org/drawer` so it becomes completely independent from other components like the `nextui-org/modal`

## Summary by CodeRabbit

- **New Features**
	- Introduced a new `useDisclosure` hook for enhanced functionality in the drawer component.

### DIFF
--- a/packages/components/drawer/src/index.ts
+++ b/packages/components/drawer/src/index.ts
@@ -7,6 +7,7 @@ export type {DrawerProps} from "./drawer";
 
 // export hooks
 export {useDrawer} from "./use-drawer";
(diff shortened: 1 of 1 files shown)


jrgarciadev:
It would cause the `@nextui-org/react` package to fail since it will export the `useDisclosure` twice

PR #4313, opened 2024-12-10
v2.6.6

jrgarciadev:
Closes #


daveycodez:
#3987 - Fix Avatar Flashing

This issue is marked as completed but it never got merged in or completed, I'm still getting consistent avatar flashing when navigating my app which is a bit jarring

Issue #4292, opened 2024-12-09
[BUG] - Interactive Elements Not Responding on iOS After Updating to NextUI 2.6.0


scarrascoalvarez:
### NextUI Version

2.6.0

### Describe the bug

After updating to NextUI 2.6.0, interactive elements such as cards, buttons, and clickable components stopped working on iOS devices when navigating through the web application. Tapping these elements does not trigger any action or event, making the interface non-functional on affected devices.

### Steps to Reproduce the Bug or Issue

	1.	Update the project to NextUI 2.6.0.
	2.	Deploy the application or run it locally.
	3.	Access the web app on an iOS device (Safari/Chrome).
	4.	Attempt to interact with buttons, cards, or clickable elements.

### Expected behavior

Interactive elements such as buttons and cards should respond to taps or clicks as expected, triggering the correct actions/events.

### Screenshots or Videos

https://github.com/user-attachments/assets/85630dd1-f479-4568-9f92-f8ec819ca738

### Operating System Version

iOS

### Browser

Safari

jrgarciadev:
We're investigating this issue, fix should be out soon

jimmailcamp:
It occurs when using buttons in Portals on ios

bckd00r:
Small clarifacition about issue, if you change from onclick to onpress and remove preventdefault issue is disappears @jimmailcamp @scarrascoalvarez @sudongyuer

sudongyuer:
Guys, have you tried use **onPress** instead of **onClick** ?

jimmailcamp:
Why is this necessary tho?

bckd00r:
> Guys, @scarrascoalvarez @jimmailcamp @bckd00r have you tried use **onPress** instead of **onClick** ?

yes, i have tried onpress and it's working now but why onClick not working on mobile devices tho?

sudongyuer:
> Why is this necessary tho?

This is just a temporary solution, will fix it soon

bckd00r:
> > Why is this necessary tho?
> 
> This is just a temporary solution, will fix it soon

thank you for quick answer i'm very glad to hear that gonna be fixed soon <3

jrgarciadev:
Hey guys please test this in the v2.6.7

sudongyuer:
> Hey guys please test this in the v2.6.7

Not woking, v2.6.7 just solve the ripple issue. I'v already found the reason, will make PR soon.

jrgarciadev:
[...]

PR #4322, opened 2024-12-10
fix: on click event on interactive elements

jrgarciadev:
Closes: #4292

## 📝 Description

This PR addresses issues related to the `onClick` event handling in various interactive components and introduces improvements by deprecating `onClick` in favor of `onPress`. Additionally, unnecessary packages have been removed, and React Aria internal utilities have been updated.

## ⛳️ Current Behavior (Updates)

- The `onClick` event is used across interactive components like `MenuItem`, `ListboxItem`, `Link`, and `Button`.
- Some internal utilities from React Aria are outdated.
- Unnecessary packages deleted

## 🚀 New Behavior

- Fixed the `onClick` event for `MenuItem`, `ListboxItem`, `Link`, `Button`, and other interactive components to ensure better consistency.
- Deprecated `onClick` in favor of `onPress` for interactive elements to align with modern accessibility standards.
- Removed unnecessary packages to reduce bundle size and improve maintainability.
- Updated React Aria internal utilities for improved performance and compatibility.

## 💣 Is this a breaking change (Yes/No): **Yes**

- **Impact**: Users currently relying on the `onClick` event will need to migrate to `onPress` as soon as possible.
- **Migration Path**: Update all instances of `onClick` to `onPress` in your codebase. 

## 📝 Additional Information

- The changes improve accessibility and align with the latest React Aria standards.
- Removed dependencies streamline the project and reduce complexity.


AnYiEE:
@jrgarciadev

After testing, passing the `onClick` event to both the `Button` and `Link` components reveals inconsistent behavior on mobile devices between versions 2.6.7 and 2.6.8, with two differences.

First, the event type for `onClick` has changed from `SyntheticBaseEvent`/`PointerEvent` to `PressEvent`.

Second, I believe that `onClick` and `onPress` are fundamentally different events. For me, when using `onPress`, if there is another element with an event listener beneath the button (at the same position but on a different layer), touching the button may also trigger that element's listener, whereas this does not happen with `onClick`. Additionally, in the case of `NavbarMenu`, sometimes touching a link with `onPress` does not automatically close the menu, while it does so reliably with `onClick`.

Overall, the behavior of `onClick` is more stable, everything was fine on 2.6.7.

I'm not sure if the two situations I mentioned are due to a bug in `react-aria` or something else, but before 2.6.8 I could work around them using `onClick`, whereas 2.6.8 changed them all to `onPress` and did not provide any option to exit this behavior. I did not reproduce the problem addressed by this PR; instead, it introduced new issues.

Thank you for your work, but could you look into any other solutions? I don't want to keep the version at 2.6.7 because future versions may have other bug fixes that I am aware of.

AnYiEE:
In version 2.6.7, when using `onClick`, touching the button above the blue avatar does not trigger the event for the blue avatar, while using `onPress` may trigger it, so I used `onClick` to avoid this issue. However, in version 2.6.8, this no longer works. Even if `onClick` is passed, it ultimately all becomes `onPress`.

In the image is 2.6.7, you can also see that the event is `SyntheticBaseEvent`/`PointerEvent` instead of `PressEvent`.

image

PR #4327, opened 2024-12-11
fix(shared-utils): process?.env will not transform to env by transcompiler

s524797336:
When `warn` message such as button `onClick` deprecated warn, throw error when there is no global process var, transcompiler will not transform `process?.env` to env string with `?` in it.

## 📝 Description
The `warn` method cause page error because `process` is not defined.

Page break render

## 🚀 New behavior
Page correct render

No

## Summary by CodeRabbit

- **Bug Fixes**
	- Enhanced error handling in the warning function to ensure it raises an error if the process object is undefined, improving stability in various environments.

### DIFF
--- a/.changeset/clever-teachers-live.md
+++ b/.changeset/clever-teachers-live.md
@@ -1,0 +1,5 @@
+---
+"@nextui-org/shared-utils": patch
+---
+
+fix process is not defined
--- a/packages/utilities/shared-utils/src/console.ts
+++ b/packages/utilities/shared-utils/src/console.ts
@@ -8,7 +8,7 @@ export function warn(message: string, component?: string, ...args: any[]) {
   if (warningStack[log]) return;
   warningStack[log] = true;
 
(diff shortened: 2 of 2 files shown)


jrgarciadev:
Hey @s524797336 could you add the error your getting?  / Replicaton ?

s524797336:
> Hey @s524797336 could you add the error your getting? / Replicaton ?

image

A compiler like Webpack automatically replaces process.env.NODE_ENV with a string, such as 'development'. However, if you write it as process?.env?.NODE_ENV, Webpack won’t be able to replace it. This results in the process variable being directly executed in the browser, which is not defined by code, causing an error to be thrown.

jrgarciadev:
@wingkwong please check

wingkwong (commented):
~~Do you have any repo that can show us how you configure or reproduce the issue?~~

UPD: ok i can reproduce it now.

wingkwong (approved):

PR #4338, opened 2024-12-11
fix(input-otp): filter disallowed characters from value prop

### LINKED ISSUE #4329: [BUG] - `InputOtp` does not work `allowedKeys` when using React Hook Form
### NextUI Version

2.6.8

### Describe the bug

The `allowedKeys` prop in the `InputOtp` component does not work when using the React Hook Form. Both the default value and custom prop value are not applied.

_No response_

1. Go to https://nextui.org/docs/components/input-otp#react-hook-form
2. Enter an OTP code that does not consist of numerical digits
3. The example in the screenshot

image

### Expected behavior

By default, the `allowedKeys` parameter is set to `^[0-9]*$`. The same behavior is expected when using React Hook Form.

### Operating System Version

Any

### Browser

Chrome

macci001:
Closes #4329

## 📝 Description

Change ensures that the input value does not accept any disallowed characters when the value prop contains them.

https://github.com/user-attachments/assets/45cb3e05-908f-46d2-8c94-a83ca8acac3e

## 🚀 New behavior
* Disallowed characters typed will be avoided. 

https://github.com/user-attachments/assets/5f5dbd8c-cbbd-4ec3-8345-67b3459190a7

## Summary by CodeRabbit

- **New Features**
  - Enhanced input integrity for the OTP component by restricting disallowed characters.
  - Integrated `react-hook-form` functionality for improved form handling of the OTP input.

- **Bug Fixes**
  - Improved validation logic to ensure only allowed characters are accepted in the input.

- **Tests**
  - Added a new test case to verify that the input correctly reflects allowed characters when using `react-hook-form`.

### DIFF
--- a/packages/components/input-otp/src/use-input-otp.ts
+++ b/packages/components/input-otp/src/use-input-otp.ts
@@ -119,6 +119,7 @@ export function useInputOtp(originalProps: UseInputOtpProps) {
     pasteTransformer,
     containerClassName,
     noScriptCSSFallback,
+    onChange,
     ...otherProps
   } = props;
 
@@ -209,6 +210,14 @@ export function useInputOtp(originalProps: UseInputOtpProps) {
           }),
           filterDOMProps(props),
         ),
+        onChange: (e: React.ChangeEvent<HTMLInputElement>) => {
+          const val = e.target?.value;
+          const regex = new RegExp(allowedKeys);
+
+          if (regex.test(val)) {
+            onChange?.(e);
+          }
+        },
       };
     },
     [baseDomRef, slots, baseStyles, isDisabled, isInvalid, isRequired, isReadOnly, value, length],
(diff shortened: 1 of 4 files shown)


packages/components/input-otp/src/use-input-otp.ts
```diff
@@ -127,6 +127,14 @@ export function useInputOtp(originalProps: UseInputOtpProps) {
   const inputRef = useDOMRef<HTMLInputElement>(ref);
   const baseDomRef = useDOMRef<HTMLDivElement>(baseRef);
 
+  const regex = new RegExp(allowedKeys);
+  let valueFromProps = props.value
+    ? props.value
+        .split("")
+        .filter((char) => regex.test(char))
+        .join("")
+    : undefined;
```
wingkwong, line 136: wondering why we test each char one by one instead of just testing the whole `props.value`?

wingkwong (commented):
you may check the following case to see if it is expected.

1. type 1-2-3-4
2. click the first input (i.e. digit 1)
3. type an invalid key, says `a`
4. digit 1 is gone. it becomes 2-3-4 (my question here: should inputting an invalid character erase the existing value?)

PR #4482, opened 2025-01-02
fix(tooltip): tooltip stays open on fast movement

### LINKED ISSUE #4301: [BUG] - Tooltip stays open more than it should
### NextUI Version

@nextui-org/tooltip@2.2.4 (or the one that is in the documentation)

### Describe the bug

When you rapidly move cursor over elements with Tooltip element the tooltip stays sometimes open.

image

In the provided image, there are multiple buttons that I moved my mouse over.

https://nextui.org/docs/components/tooltip#placements

1. Go to provided URL link
2. Scroll to [**Placements**](https://nextui.org/docs/components/tooltip#placements)
3. Rapidly move mouse cursor over those buttons and you should see that some of the tooltips are staying longer then expected 

### Expected behavior

The tooltips should close after while and when I dont hover over any of them the Tooltip should close.

### Screenshots or Videos

image
image
clideo_editor_8bd713f5ec1a413bafb1bd040230c624

### Operating System Version

Ubuntu 24.04.1 LTS

### Browser

Chrome

Connorelsea:
Closes #4301 

## 📝 Description

This PR fixes an issue where tooltips when hovered in quick succession would sometimes remain open in NextUI. It updates AnimatePresence and motion key handling to comply with motion's typical usage patterns.

When moving a cursor quickly across a screen with many tooltips, some tooltips would stay open after mouse has left. This can be easily replicated by making a list of tooltips and moving the cursor up and down them at medium-quick speed, especially when `closeDelay` and `openDelay` are set to zero to increase the chance that interruptions of the close animation will occur.

## 🚀 New behavior
Tooltip no longer stays open on fast mouse movements across a list of tooltips. This ensures `AnimatePresence` is not prematurely unrendered on tooltip close by ensuring `AnimatePresence` is always rendered when `disableAnimation` is false, and not unrendered when `isOpen` is false. Previously animate presence was being unrendered on close due to the conditional combining `isOpen` and `disableAnimation` checks.  This also adds the proper keys that allow framer-motion to properly track these animations. Removing the keys causes the regression to return.

No breaking changes, small internal fix to tooltip with no effect on current usage patterns.

## 📝 Additional Information

I tried to think of a way to test this programatically, but I could not think of anything good. I added a storybook where I was able to test manually and replicate this issue locally on the canary branch. You can replicate by removing the keys and moving your cursor quickly up and down the list. Adding the keys back, alongside the change in AnimatePresence rendering, seemingly eliminates this behavior.

If someone can think of a better way to write a test for this, or wants to remove the storybook (or present it differently) as it was mostly used for testing this issue, let me know! I am available to make changes to this PR.

Replicated on public docs:

https://github.com/user-attachments/assets/84909b63-873b-44f9-8b8f-8fb22beedd92

Replicated on the storybook I added (pre-fix)

telegram-cloud-photo-size-1-5168126915008114304-y

## Summary by CodeRabbit

- **New Features**
  - Enhanced tooltip rendering with improved animation handling.
  - Added a new story template demonstrating multiple tooltips with immediate open/close behavior.

- **Improvements**
  - Simplified tooltip content rendering logic.
  - Improved element identification during tooltip transitions.
  - Fixed an issue where tooltips remained open during fast movements.


jrgarciadev:
Hey @Connorelsea thanks, please add a changeset

Connorelsea:
Hello @jrgarciadev, I added a changeset, tried to follow the linked changeset documentation as I have never used that tool before. If it's incorrect I can fix it.

Ejdyz:
Hello, i dont know if its done or not, but in [this preview](https://heroui-sb-git-fork-connorelsea-bug-4301-toolt-460a35-nextui-org.vercel.app/) i was still able to replicate the issue.

OS: Ubuntu 24.04.1 LTS x86_64
Browser: 132.0.6834.110 (Official Build) (64-bit)

https://github.com/user-attachments/assets/29b82b05-1e74-485d-82ea-5af997c6aec2

packages/components/tooltip/stories/tooltip.stories.tsx
```diff
@@ -263,6 +263,35 @@ const ControlledTemplate = (args: TooltipProps) => {
...
+    </Tooltip>
+    <Tooltip {...args} closeDelay={0} content="Tooltip 7" openDelay={0}>
+      <Button>Then hover me</Button>
+    </Tooltip>
+    <Tooltip {...args} closeDelay={0} content="Tooltip 8" openDelay={0}>
+      <Button>Then hover me</Button>
+    </Tooltip>
+  </div>
+);
+
```
wingkwong, line 294: may include it in `Multiple` instead creating a new one

wingkwong (changes requested):
still reproducing the issue

image

Connorelsea:
@wingkwong Will look into it and see if I can reproduce and fix. And I'll update the Multiple storybook

AntoineArt:
Just a post to thank @Connorelsea, I use tooltips a lot and this bug was really annoying my users and I, I look forward to have the fix in official release!

fullyDeepak:
Hi, any update on this?

Jimflip:
Just to add this is causing me problems too, look forward to fix.

Connorelsea:
@wingkwong I was able to replicate it still failing, and was able to fix by ensuring the LazyMotion loader also included the AnimatePresence component. Just pushed a fix

Apologies for the delay!

packages/components/tooltip/stories/tooltip.stories.tsx
```diff
@@ -156,16 +156,41 @@ const OffsetTemplate = (args: TooltipProps) => (
...
+  }: {
+    content: React.ReactNode;
+    tooltipProps: TooltipProps;
+  }) => {
+    return (
+      <Tooltip {...tooltipProps} closeDelay={0} content={content} openDelay={0}>
+        <Button>Hover me</Button>
+      </Tooltip>
+    );
+  };
```
wingkwong, line 172: don't think it's necessary to create this.

packages/components/tooltip/stories/tooltip.stories.tsx
```diff
@@ -156,16 +156,41 @@ const OffsetTemplate = (args: TooltipProps) => (
...
+
+  const tooltipIndex = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21];
```
wingkwong, line 174: instead of hardcoding the index, you can do

```
{Array.from({ length: 21 }, (_, i) => (...))}
```

packages/components/tooltip/src/tooltip.tsx
```diff
@@ -62,41 +62,44 @@ const Tooltip = forwardRef<"div", TooltipProps>((props, ref) => {
...
       {trigger}
-      {disableAnimation && isOpen ? (
-        <OverlayContainer portalContainer={portalContainer}>
-          <div ref={tooltipRef} id={id} style={style} {...otherTooltipProps}>
-            <Component {...getTooltipContentProps()}>{content}</Component>
-          </div>
-        </OverlayContainer>
+      {disableAnimation ? (
+        isOpen && (
+          <OverlayContainer portalContainer={portalContainer}>
```
wingkwong, line 87: do we need the key here?

packages/components/tooltip/src/tooltip.tsx
```diff
@@ -62,41 +62,44 @@ const Tooltip = forwardRef<"div", TooltipProps>((props, ref) => {
...
-          variants={TRANSITION_VARIANTS.scaleSpring}
-          {...mergeProps(motionProps, otherTooltipProps)}
-          style={{
-            ...getTransformOrigins(placement),
-          }}
-        >
-          <Component {...getTooltipContentProps()}>{content}</Component>
-        </m.div>
-      </LazyMotion>
+    <div key={id + "-tooltip-content"} ref={tooltipRef} id={id} style={style}>
```
wingkwong, line 65: prefer the following instead of using `+`. (apply to other keys)

```tsx
key={`${id}-tooltip-content`}
```

wingkwong (changes requested):
can you share more why we wrap `LazyMotion` even `disableAnimation` is `true`?

packages/components/tooltip/src/tooltip.tsx
Connorelsea: The motion docs recommend that any children of `AnimatePresence` always have a key, but I did some testing and the tooltip works properly with all keys removed, on the `OverlayContainer` and the keys inside `animatedContent`. So I could probably remove those as well.

I did some research and apparently the reason it works with no keys is because framer motion has internal fallback mechanisms (component type as key, internal instance keys, component order) for tracking presence when there are no keys, and it works well here since it is a single child with a predictable structure.

I could see the argument for the `OverlayContainer` key go either way, up to you.

Connorelsea:
@wingkwong Good suggestions, made the changes. I did some testing and it works with all keys removed, so keeping them could go either way. And you are right about loading lazy motion, it's not required if animation is disabled.

wingkwong (approved):

wingkwong:
@Connorelsea seems I couldn't reproduce on the current production. Are you able to reproduce?

wingkwong:
I'm not talking about the PR. I mean the original issue. After the recent release, I could no longer reproduce the issue in the production.

Arch-Storm:
Will this actually get merged soon? it's been approved and in the milestone 3 times now but is just getting pushed back every time. I have quite a powerful machine with high refresh monitors and i get this problem so bad to the point that half my screen gets filled with tooltips after a few seconds of usage. The image shows how bad it currently is. And that is from regularly moving the mouse. I am not moving it fast on purpose.

image

aaronhawkey:
> Will this actually get merged soon? it's been approved and in the milestone 3 times now but is just getting pushed back every time. I have quite a powerful machine with high refresh monitors and i get this problem so bad to the point that half my screen gets filled with tooltips after a few seconds of usage. The image shows how bad it currently is. And that is from regularly moving the mouse. I am not moving it fast on purpose.
> 
> image

This is infact effecting our production application as well. If its a bandwidth issue from a contributor stand point, I'll raise my hand to help out.

Connorelsea:
No issues on my end, as the opener of this PR. I think it's just waiting on final review and merge. I am still around to make any updates if necessary, but I've found that these changes seem to fix the issue

wingkwong:
I'll bring this up in tomorrow's meeting. Previously none of members could reproduce the issue on our side so it's hard to verify this PR.

@aaronhawkey @Arch-Storm are you guys able to check on your dev environment with this PR build version `https://pkg.pr.new/@heroui/react@4482`?

Arch-Storm:
> @aaronhawkey @Arch-Storm are you guys able to check on your dev environment with this PR build version `https://pkg.pr.new/@heroui/react@4482`?

Have just tested using the pr preview and the issue is indeed fully fixed on my end!

jrgarciadev (approved):

PR #4562, opened 2025-01-15
updating the accordian api's and introducing the disclosure component

macci001:
Closes # 

## 📝 Description

Breaking changes in Accordion:
- selectionMode control is changed to allowsMultipleExpanded
  - for single selection allowsMultipleExpanded is false
  - for multiple selection allowsMultipleExpanded is true
- selectionBehavior will no longer exist.
- DefaultSelected is changed to DefaultExpanded.
- SelectedKeys is changed to ExpandedKeys
- onSelectionChange is now onExpandedChange. Note: onExpandedChanged do not expect "all" as argument.
- disallowEmptySelection will not exist anymore
- motionProps will not exist, the motion can be changes with the tailwind styles
- renamed key to id

### DIFF
--- a/apps/docs/content/components/accordion/bordered-variant.raw.jsx
+++ b/apps/docs/content/components/accordion/bordered-variant.raw.jsx
@@ -6,13 +6,13 @@ export default function App() {
 
   return (
     <Accordion variant="bordered">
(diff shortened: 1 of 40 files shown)


macci001:
@coderabbitai pause

PR #4580, opened 2025-01-17
docs(installation): Update Hoisted dependencies step to help users with Prettier/EsLint

J4v4Scr1pt:
## 📝 Description
Updated Hoisted dependencies step in installation.mdx.
Added extra information about hoisted pnpm packages. Because without prior knowledge about hoisting you may end up with broken Prettier and EsLint.  Read more here: https://github.com/pnpm/pnpm/issues/8878

Docs don't inform new users that their Prettier and EsLint may stop working because of how Pnpm and hoisting works.

## 🚀 New behavior

Now the user will have the information needed to prevent non-working Prettier and EsLint.

### DIFF
--- a/apps/docs/content/docs/guide/installation.mdx
+++ b/apps/docs/content/docs/guide/installation.mdx
@@ -151,6 +151,13 @@ If you are using pnpm, you need to add the following line to your `.npmrc` file 
 ```bash
 public-hoist-pattern[]=*@heroui/*
 ```
+If you are using Prettier and/or EsLint include them accordingly.
+
+```bash
+public-hoist-pattern[]=*@heroui/*
+public-hoist-pattern[]=*eslint*
+public-hoist-pattern[]=*prettier*
+```
 
 After modifying the `.npmrc` file, you need to run `pnpm install` again to ensure that the dependencies are installed correctly.
 


AnYiEE:
It seems that `public-hoist-pattern[]=*eslint-plugin*` is enough.

J4v4Scr1pt:
> It seems that `public-hoist-pattern[]=*eslint-plugin*` is enough.

The reason I added Prettier was because of it being default in the pnpm array. But maybe that is for early versions of Prettier?

AnYiEE:
> > It seems that `public-hoist-pattern[]=*eslint-plugin*` is enough.
> 
> The reason I added Prettier was because of it being default in the pnpm array. But maybe that is for early versions of Prettier?

There may be other uses. For a Next.js + ESLint 9 project, using Flat Config, with `public-hoist-pattern[]=*eslint-plugin*` is indeed sufficient. ESLint 8 doesn't even require `public-hoist-pattern`. (Note: Also using Prettier 3 along with `eslint-config-prettier` and `prettier-plugin-tailwindcss`)

J4v4Scr1pt:
> > > It seems that `public-hoist-pattern[]=*eslint-plugin*` is enough.
> > 
> > 
> > The reason I added Prettier was because of it being default in the pnpm array. But maybe that is for early versions of Prettier?
> 
> There may be other uses. For a Next.js + ESLint 9 project, using Flat Config, with `public-hoist-pattern[]=*eslint-plugin*` is indeed sufficient. ESLint 8 doesn't even require `public-hoist-pattern`. (Note: Also using Prettier 3 along with `eslint-config-prettier` and `prettier-plugin-tailwindcss`)

Ahh ok, maybe I should add som more info context the step then maybe? What you think?

AnYiEE:
> > > > It seems that `public-hoist-pattern[]=*eslint-plugin*` is enough.
> > > 
> > > 
> > > The reason I added Prettier was because of it being default in the pnpm array. But maybe that is for early versions of Prettier?
> > 
> > 
> > There may be other uses. For a Next.js + ESLint 9 project, using Flat Config, with `public-hoist-pattern[]=*eslint-plugin*` is indeed sufficient. ESLint 8 doesn't even require `public-hoist-pattern`. (Note: Also using Prettier 3 along with `eslint-config-prettier` and `prettier-plugin-tailwindcss`)
> 
> Ahh ok, maybe I should add som more info context the step then maybe? What you think?

Yes, it seems that a case-by-case explanation is needed.

J4v4Scr1pt:
> 
> Yes, it seems that a case-by-case explanation is needed.
Ok great then I will fix this 👍

J4v4Scr1pt:
I have updated the docs now to be more clear.
From my understand correct me if I'm wrong.
Prettier <3 need the hoist pattern.
EsLint <9 need the hoist pattern. But there are people in v9 with flat-config (e.g. me) that experience issues and need to have it also, because some setups with ESLint v9 still require hoisting.

AnYiEE:
> EsLint <9 need the hoist pattern.

Only for me. My project previously used Next.js 14, EsLint 8 and Prettier 3.4, without needing any `public-hoist-pattern` (except for `@nextui-org/*`). After upgrading to Next.js 15 and EsLint 9, I need an additional `public-hoist-pattern[]=*eslint-plugin*` only.

J4v4Scr1pt:
> > EsLint <9 need the hoist pattern.
> 
> Only for me. My project previously used Next.js 14, EsLint 8 and Prettier 3.4, without needing any `public-hoist-pattern` (except for `@nextui-org/*`). After upgrading to Next.js 15 and EsLint 9, I need an additional `public-hoist-pattern[]=*eslint-plugin*` only.

I had a similar experience expect that neither EsLint 8 or 9 worked 😅. But according to the PR they removed the defaults because of version 9 and the flat config. That's why I tried to inform that in the information.

What you think it's ok now? :)
image

apps/docs/content/docs/guide/installation.mdx
```diff
@@ -152,6 +152,26 @@ If you are using pnpm, you need to add the following line to your `.npmrc` file
 public-hoist-pattern[]=*@heroui/*
 ```
 
+<Blockquote color="primary">
+**Note**: In version <10 of pnpm the default values of the hoist pattern contains ['*eslint*', '*prettier*'] and by adding the hoist pattern they will get removed and you may experince the tools not working.
```
ryo-manba, line 156: This is originally a pnpm issue, so I'm not sure if we should include setup details for external libraries like Prettier or ESLint.Providing useful information is important, but we should also avoid making the documentation overly lengthy. I'd appreciate hearing other members' thoughts.

ryo-manba (commented):
This is originally a pnpm issue, so I'm not sure if we should include setup details for external libraries like Prettier or ESLint. Providing useful information is important, but we should also avoid making the documentation overly lengthy. I'd appreciate hearing other members' thoughts.

J4v4Scr1pt:
> This is originally a pnpm issue, so I'm not sure if we should include setup details for external libraries like Prettier or ESLint. Providing useful information is important, but we should also avoid making the documentation overly lengthy. I'd appreciate hearing other members' thoughts.

I totally agree with you 👍😊! Because my intention is just to help others by having some information about this. Because I spent quite some time trying to understand why my Eslint failed 😅. 
We could maybe just add a smaller note saying something like this:
_"If you are on pnpm <10 Prettier and EsLint is default in they hoist array and depending on your setup you may need to be add them back."_

PR #4628, opened 2025-01-23
Fix/routerized links

### LINKED ISSUE #4624: [BUG] - useLink overwrites router-ized prop
### HeroUI Version

2.6.13

### Describe the bug

This is my current theory as to why my links in RR with useHref are not working:

1. [useLink](https://github.com/heroui-inc/heroui/blob/cddba8281cb161d0a8afb42e20c05094974af981/packages/components/link/src/use-link.ts#L45) calls useAriaLink()
2. [useAriaLink](https://github.com/heroui-inc/heroui/blob/cddba8281cb161d0a8afb42e20c05094974af981/packages/hooks/use-aria-link/src/index.ts#L43) calls useLinkProps from react-aria
3. [useLinkProps](https://github.com/adobe/react-spectrum/blob/2a1c28b8370f4a19e4bdb92e0daf14d07990f27a/packages/%40react-aria/utils/src/openLink.tsx#L174) will use useHref to change the href prop and send it back to useAriaLink
4. useAriaLink passes back to useLink
5. useLink does a mergeProps with the original props ('otherProps') after linkProps https://github.com/heroui-inc/heroui/blob/cddba8281cb161d0a8afb42e20c05094974af981/packages/components/link/src/use-link.ts#L113 - overwriting the useHref'ed href

This can be seen as the relative link is "bare" (ie does not have the rest of the route path that a plain useHref locally would show)

ie const href = useHref("blogs") would return '1/blogs' in my case, but <Link href="blogs" /> stays as "blogs" in the <a> element href

This breaks all relative link navigations for me.

https://stackblitz.com/edit/wfdcu9f2 

React Router setup with HeroUIProvider (navigation and href); try relative links.  I have a complex setup I can demonstrate on a screenshare if necessary.

### Expected behavior

Relative links work

### Screenshots or Videos

_No response_

### Operating System Version

macOS

### Browser

Chrome

jprosevear:
Closes #4624 

## 📝 Description

Use the routerized href if available for both the onClick event and the href prop on the <a> element

Currently the original href is used (ignoring the application of useHref) on the onClick and so navigate() for react-router acts on the wrong href and fails for relative links.   So if you are at /blogs already and you put href="14" it will try to go to /14 instead of /blogs/14 and the <a> href is decorate with href="14"

## 🚀 New behavior

navigate() for react-router works, going to /blogs/14 and the <a> element is decorated correctly (/blogs/14) .

Possible, but I don't think it worked before with relative links in RR; absolute links should still work.


packages/components/link/src/use-link.ts
```diff
@@ -110,6 +112,8 @@ export function useLink(originalProps: UseLinkProps) {
       "data-focus": dataAttr(isFocused),
       "data-disabled": dataAttr(originalProps.isDisabled),
       "data-focus-visible": dataAttr(isFocusVisible),
+      // The `href` prop may be routerized by useAriaLink, so we may merge over top of it
+      href,
```
jprosevear, line 116: This makes sure that the href on the <a> is set properly

packages/hooks/use-aria-link/src/index.ts
```diff
@@ -118,10 +118,10 @@ export function useAriaLink(props: AriaLinkOptions, ref: RefObject<FocusableElem
           // If props are applied to a router Link component, it may have already prevented default.
           !e.isDefaultPrevented() &&
           shouldClientNavigate(e.currentTarget, e) &&
-          props.href
+          routerLinkProps.href
         ) {
           e.preventDefault();
-          router.open(e.currentTarget, e, props.href, props.routerOptions);
+          router.open(e.currentTarget, e, routerLinkProps.href, props.routerOptions);
```
jprosevear, line 124: This makes sure that we use the useHref'ed href for the navigation

wingkwong:
may revisit this PR again after #4611

jprosevear:
> may revisit this PR again after #4611

Is the hope that the upgrade will fix this or something else?

wingkwong:
@jprosevear RA recently bumped their version and we are still using the second last latest version. The version diff would cause some routing issues. So I wanna double check again after syncing the versions.

jprosevear:
> @jprosevear RA recently bumped their version and we are still using the second last latest version. The version diff would cause some routing issues. So I wanna double check again after syncing the versions.

Ok, but based on my patch and investigation the issue is that RA was returning the right info already, it just wasn't being used correctly in Hero.

jprosevear:
Rebased.

wingkwong:
@jprosevear do you have a simple sandbox that can reproduce the issue? I need it to test the before & after.

jprosevear:
> @jprosevear do you have a simple sandbox that can reproduce the issue? I need it to test the before & after.

I do not, unless there is a good starting point for heroui / RR (or Tanstack Router, as per Discord I think it has the same problem) its probably a couple hours of work at least is my guess which is possible but painful obviously.

What is the right thing to do from a testing infrastructure standpoint long term?

packages/components/link/src/use-link.ts
```diff
@@ -73,6 +74,7 @@ export function useLink(originalProps: UseLinkProps) {
   const {linkProps} = useAriaLink(
     {
       ...otherProps,
+      href,
```
wingkwong, line 77: `href` originally is in `otherProps`

packages/hooks/use-aria-link/src/index.ts
wingkwong, line 124: by doing so, one couldn't do any absolute paths, e.g. says I'm at `/blogs` and if i set `href` to `/foo/1`, I will be navigated to `/blogs//foo/1` instead of `/foo/1`.

wingkwong (commented):
FYI, our `use-aria-link` mirrors `useLink` from `@react-aria`. We modified the deprecated warning part only. I think navigating to `routerLinkProps.href` isn't expected since it will cover the current path.

packages/components/link/src/use-link.ts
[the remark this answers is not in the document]
jprosevear: In case you are not using a routerized link you still need the original href I thought

packages/hooks/use-aria-link/src/index.ts
jprosevear: Why not?  useHref  from RR should produce /foo/1 in that case.  I updated https://stackblitz.com/edit/wfdcu9f2 with that case.

packages/components/link/src/use-link.ts
jprosevear: I know; see below.

packages/hooks/use-aria-link/src/index.ts
wingkwong: RRLink can navigate to `/foo/1`correctly. With your PR change, I got `//blogs/foo/1`. Note that even I'm at `/blogs` originally, it would still add the slash causing `//` at the end. You may link those components locally to play around.

packages/components/link/src/use-link.ts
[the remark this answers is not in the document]
wingkwong: `href` will be passed to `useLinkProps`, and eventually it will be in `linkProps ` anyway.

packages/hooks/use-aria-link/src/index.ts
jprosevear: https://github.com/adobe/react-spectrum/blob/016590a4afc585eea18d5a12853c9d1e2d82a19c/packages/%40react-aria/utils/src/openLink.tsx#L174 for reference - unless the useLinkProps href is used, why call it it at all?  Also, if the href is something like "https://www.heroui.com/" i believe the router.open call won't be made on it, a regular link will happen.

jprosevear:
> > @jprosevear do you have a simple sandbox that can reproduce the issue? I need it to test the before & after.
> 
> I do not, unless there is a good starting point for heroui / RR (or Tanstack Router, as per Discord I think it has the same problem) its probably a couple hours of work at least is my guess which is possible but painful obviously.
> 
> What is the right thing to do from a testing infrastructure standpoint long term?

https://stackblitz.com/edit/wfdcu9f2

packages/components/link/src/use-link.ts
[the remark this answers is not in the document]
jprosevear: It still needs to be extracted from otherProps though so mergeProps doesn't overwrite the linkProps href with the otherPropsHref.

packages/hooks/use-aria-link/src/index.ts
jprosevear: > RRLink can navigate to `/foo/1`correctly. With your PR change, I got `//blogs/foo/1`. Note that even I'm at `/blogs` originally, it would still add the slash causing `//` at the end. You may link those components locally to play around.

Can you give more details?  All this patch really does is ensure the href produced by useLinkProps from react-aria is actually used in Hero (at least that was my intention)

packages/components/link/src/use-link.ts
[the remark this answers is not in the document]
jprosevear: the key for me is that useLinkProps can *change* href

wingkwong:
UPD: was using the outdated stackblitz for my initial review. will take a look again.

jprosevear:
Upstream discussion: https://github.com/adobe/react-spectrum/pull/7695

jprosevear:
@jrgarciadev I would not include this in 2.7.x until there is some resolution upstream

PR #4655, opened 2025-01-26
refactor: migrate external packages to internal to avoid version incompatibility

### LINKED ISSUE #4567: [Feature Request] Mitigate need for users to keep @internationalized/date version in sync with NextUI

Today, in order to use`<DatePicker>`, users of NextUI also have to take on a dependency on the `@internationalized/date` package.

It's also necessary to keep your version of this library exactly in sync with NextUI. (for example, a version mismatch can cause an error like `Type 'ZonedDateTime' is not assignable to type 'CalendarDate | CalendarDateTime | ZonedDateTime | null | undefined'.`, seen in the screenshot below)

(I'm guessing this isn't a surprise, as the [docs](https://nextui.org/docs/components/date-picker) provide instructions to install 3.6.0 specifically)

What if NextUI re-exported symbols from their own version of `@internationlized/date`? So users could switch from this:

```ts
import { DatePicker, DatePickerProps } from "@nextui-org/date-picker";
import { CalendarDate, parseDate } from "@internationalized/date";
```

To something like this:

```ts
import { DatePicker, DatePickerProps } from "@nextui-org/date-picker";
import { CalendarDate, parseDate } from "@nextui-org/date-picker/internalized-date";
```

Then, if I'm only using `@internationalized/date` for NextUI, then I won't need to add it to my package.json (and won't accidentally update it).

Or, if I _am_ using it elsewhere, I'll have a chance of being able to update it without breaking NextUI. (though I'd probably run into version conflicts elsewhere)

Another issue suggests not requiring `@internationalized/date` at all: https://github.com/nextui-org/nextui/issues/4401

### Screenshots or Videos

image

wingkwong:
Closes #4567

## 📝 Description

Currently users may need to install some external packages. However, since we are using fixed versions for them in HeroUI to avoid unexpected issues, users still may install the latest version (even the doc explicitly defines the exact version). The version incompatibility causes issues from time to time.

This PR is to create a mirror version of the following package so that users will use the exact same version.

- `@internationalized/date` -> `@heroui/internationalized-date`
- `@react-aria/i18n` -> `@heroui/react-aria-i18n`
- `framer-motion` -> `@heroui/motion` (also migrated to `motion` v12)

These three packages will not be included in `@heroui/react`. The reasons are
- to minimize the package size
- to avoid naming conflicts

Sort of. Users can still use those 3 packages, but they need to make sure to use the exact version. Otherwise, simply install the package and rename the imports.

`@heroui/framer-utils` will be merged to `@heroui/motion`


packages/external/motion/package.json
```diff
@@ -1,11 +1,12 @@
...
+  "name": "@heroui/motion",
+  "version": "2.0.0",
+  "description": "Mirror package of framer-motion",
   "keywords": [
-    "framer-utils"
+    "hero ui",
+    "motion"
   ],
-  "author": "Junior Garcia <jrgarciadev@gmail.com>",
+  "author": "WK Wong <wingkwong.code@gmail.com>",
```
jrgarciadev, line 9: @wingkwong let's refactor all packages' authorm add "HeroUI <support@heroui.com>"

jrgarciadev:
@wingkwong please check build on Vercel
CleanShot 2025-02-05 at 17 41 57@2x

packages/external/motion/package.json
wingkwong: sure. will update this one and handle the existing components in another PR.

ssijak:
Any news on this? In particular migration to motion

wingkwong:
@ssijak It's scheduled in 2.8.0. Currently we plan to have a few more bug fix releases first.

toxsick:
Hey guys, can you please share why you decided to drop this. I felt like this sounds like a good idea.

Thanks a lot!

wingkwong:
@toxsick in v3, we tend to keep everything simple and those may not be required anymore and we do not want to confuse users in 2.8.0 by migrating these packages and dropping them in the next major version.

PR #3746, opened 2024-09-13
feat(select): add isClearable

abhinav700:
Closes #2239 

## 📝 Description
Added clear button functionality to select component.
Screenshot from 2024-09-13 18-02-21
Screenshot from 2024-09-13 18-02-11
Screenshot from 2024-09-13 18-01-17

## Summary by CodeRabbit

- **New Features**
	- Introduced a selection component for choosing favorite animals with a clearable feature.
	- Added a `PetBoldIcon` component to enhance the dropdown menu.
	- Users can now reset their selection easily with the clearable functionality.
	- Documentation updated to include the new `isClearable` property and `onClear` event handler for the Select component.

- **Bug Fixes**
	- Enhanced the styling and responsiveness of the clear button across different component sizes.

PR #4785, opened 2025-02-02
feat(select): adding isClearable to select [ Recreation ]

abhinav700:
this PR is duplicate of #3746
it adds a clear functionality to the select component.

image
image
image


abhinav700:
@wingkwong can you check this new PR please

apps/docs/content/components/select/is-clearable.raw.jsx
```diff
@@ -0,0 +1,65 @@
+import {Select, SelectItem} from "@nextui-org/react";
```
wingkwong, line 1: please update the package name

.changeset/three-socks-glow.md
```diff
@@ -0,0 +1,5 @@
+---
+"@heroui/select": patch
```
wingkwong, line 2: include theme package also

wingkwong (changes requested):
1. please add tests
2. please update theme version in select peer dep in package.json

apps/docs/content/components/select/is-clearable.raw.jsx
abhinav700: are you referring to `@nextui-org` part?

.changeset/three-socks-glow.md
abhinav700: `@heroui/theme` package?

wingkwong: yes

apps/docs/content/components/select/is-clearable.raw.jsx
wingkwong: yes. now we've rebranded. please change to `@heroui/react`

abhinav700:
@wingkwong I have made the changes, and changed one of the tests a little bit.

packages/components/select/src/select.tsx
```diff
@@ -54,12 +54,35 @@ const Select = forwardRef(function Select<T extends object>(
...
+      return (
+        <div className="flex end-18">
+          {clearButton}
+          {endContent && <span className="ms-3">{endContent}</span>}
+        </div>
+      );
+    }
+
+    return endContent && <span className="mb-4">{endContent}</span>;
+  }, [clearButton, endContent]);
```
wingkwong, line 84: styles should be in theme package, not in component level.

wingkwong (changes requested):

packages/components/select/src/select.tsx
abhinav700: will fix it

apps/docs/content/components/select/is-clearable.raw.jsx
```diff
@@ -0,0 +1,65 @@
...
+  return (
+    <div className="flex w-screen justify-center items-center">
+      <Select className="max-w-xs my-5" isClearable={true} label="Favorite Animal">
+        {animals.map((animal) => (
+          <SelectItem key={animal.key}>{animal.label}</SelectItem>
+        ))}
+      </Select>
+    </div>
+  );
+}
```
wingkwong, line 65: please align with other examples, i.e. display on the left.

image

apps/docs/content/components/select/is-clearable.raw.jsx
```diff
@@ -0,0 +1,65 @@
...
+        d="M20.5496 12.9313C21.6266 12.9313 22.4996 12.0582 22.4996 10.9812C22.4996 9.90429 21.6266 9.03125 20.5496 9.03125C19.4727 9.03125 18.5996 9.90429 18.5996 10.9812C18.5996 12.0582 19.4727 12.9313 20.5496 12.9313Z"
+        fill="currentColor"
+      />
+      <path
+        d="M3.94 10.9816C5.28757 10.9816 6.38 9.88914 6.38 8.54156C6.38 7.19399 5.28757 6.10156 3.94 6.10156C2.59243 6.10156 1.5 7.19399 1.5 8.54156C1.5 9.88914 2.59243 10.9816 3.94 10.9816Z"
+        fill="currentColor"
+      />
+    </svg>
+  );
+};
```
wingkwong, line 53: not in use. please delete.

packages/components/select/src/use-select.ts
```diff
@@ -341,11 +347,23 @@ export function useSelect<T extends object>(originalProps: UseSelectProps<T>) {
     triggerRef,
   );
 
+  const handleClear = useCallback(() => {
+    state.setSelectedKeys(new Set([]));
+    onClear?.();
+    domRef.current?.focus();
```
wingkwong, line 353: select an item -> click clear button -> press a space bar.

image

packages/components/select/src/use-select.ts
```diff
@@ -671,6 +691,32 @@ export function useSelect<T extends object>(originalProps: UseSelectProps<T>) {
...
+    },
+    [slots, classNames?.endWrapper],
+  );
+
+  const getEndContentProps: PropGetter = useCallback(
+    (props = {}) => {
+      return {
+        ...props,
+        "data-slot": "end-content",
+        className: slots.endWrapper({
```
wingkwong, line 712: should it be `slots.endContent`?

wingkwong (changes requested):
#### 1: After setting a value, endContent would shift to the middle, which doesn't look good. I think initially the endContent can be displayed next to the selector button. In this way , even clear button shows up, endContent won't be shifted.

image

image

---

#### 2: After setting a value, see the console

image

apps/docs/content/components/select/is-clearable.raw.jsx
abhinav700: @wingkwong should I display endContent next to selector button even when isClearable is false?

abhinav700: Okay, will fix that. I thought that alignment would change depending on the isClearable

abhinav700:
@wingkwong I have implemented the fixes. can you check please?

wingkwong (changes requested):
overall it looks good. just one final comment. when there is no end content, the clear button should be placed closer to the selector button (just like end content in the second screenshot)

image

image

abhinav700:
is this fine? @wingkwong 
Screenshot from 2025-06-02 22-50-11
Screenshot from 2025-06-02 22-49-46

wingkwong:
yes

wingkwong (changes requested):
The style is a bit off for underline variant (with endContent + clear button). You may compare with other variants.

image

image

With label placement. please test with all combinations.

image

wingkwong (changes requested):
clearable + underlined + size = md

image

endContent + isClearable=false + size = lg

image

PR #5044, opened 2025-03-13
fix(slider): prevent NaN error for equal min / max values

### LINKED ISSUE #5014: [BUG] - Range slider breaks when min and max are the same
### HeroUI Version

@heroui/slider@2.4.12

### Describe the bug

If the values ​​are the same, it starts acting somehow inadequately, writing Invalid end: NaN

https://codesandbox.io/p/devbox/intelligent-pike-wwcmnw

_No response_

1. Set minValue={100}
2. Set maxValue={100}
3. Try dragging

### Expected behavior

Fix the error and make it possible to set min max to be the same

### Screenshots or Videos

_No response_

### Operating System Version

macOS

### Browser

Chrome

kaishuige:
Closes #5014 

## 📝 Description

Fixed an issue where the slider component would display "Invalid end: NaN" when minimum and maximum values are the same. The fix maintains slider interactivity while preventing NaN calculations.

No error will be reported. NAN will not be displayed. Fixed values will be displayed. You can't slide and click.

## 🚀 New behavior

NO

NO

## Summary by CodeRabbit

- **Bug Fixes**
	- Enhanced the slider component to correctly handle scenarios where the minimum and maximum values are the same. The slider now reliably displays the fixed value, prevents “NaN” outputs, and maintains proper accessibility and interaction behavior.
- **Tests**
	- Added a new test case to ensure the slider behaves correctly when minimum and maximum values are identical.


wingkwong (changes requested):
1. please add changeset.
2. I think we don't need to change slider-thumb, and use-slider-thumb at all. `NaN` is caused from RA side since supposedly min and max shouldn't be same. So my idea is that you can simply create a new variable called `stateValues` in `use-slider.ts`. If it is a fixed value, then remove filter out all NaN values, else take state.values. Then replace all existing `state.values` with stateValue. In this case, it would be much cleaner. Also you may return after L420 if it is a fixed value.

kaishuige:
OK, I'll try it later.

packages/components/slider/src/use-slider.ts
```diff
@@ -270,6 +297,7 @@ export function useSlider(originalProps: UseSliderProps) {
       "data-orientation": state.orientation,
       "data-slot": "base",
       "data-hover": isHovered,
+      "data-fixed-value": isFixedValue,
```
wingkwong, line 300: not necessary

packages/components/slider/src/use-slider.ts
```diff
@@ -248,19 +250,44 @@ export function useSlider(originalProps: UseSliderProps) {
...
+    if (isFixedValue) return 0;
+
+    return state.getThumbPercent(index);
+  };
+
+  const getValuePercent = (value: number) => {
+    if (isFixedValue) return 0;
+
+    return state.getValuePercent(value);
+  };
```
wingkwong, line 263: not required if you don't include L273 - L276.

packages/components/slider/src/use-slider.ts
```diff
@@ -248,19 +250,44 @@ export function useSlider(originalProps: UseSliderProps) {
...
+    if (isFixedValue) {
+      return state.values.filter((val) => !Number.isNaN(val));
+    }
+
+    return state.values;
+  }, [isFixedValue, state.values]);
+
+  if (isFixedValue && stateValues.length === 0) {
+    stateValues.push(minValue);
+  }
```
wingkwong, line 275: not required.

packages/components/slider/src/use-slider.ts
```diff
@@ -248,19 +250,44 @@ export function useSlider(originalProps: UseSliderProps) {
...
-    state.values.length > 1
-      ? state.getThumbPercent(0)
+    stateValues.length > 1
+      ? getThumbPercent(0)
       : fillOffset !== undefined
-      ? state.getValuePercent(fillOffset)
+      ? getValuePercent(fillOffset)
       : 0,
-    state.getThumbPercent(state.values.length - 1),
+    getThumbPercent(stateValues.length - 1),
```
wingkwong, line 283: still use `state.getThumbPercent`

packages/components/slider/src/use-slider.ts
```diff
@@ -248,19 +250,44 @@ export function useSlider(originalProps: UseSliderProps) {
...
+  }
+
   const [startOffset, endOffset] = [
-    state.values.length > 1
-      ? state.getThumbPercent(0)
+    stateValues.length > 1
+      ? getThumbPercent(0)
       : fillOffset !== undefined
-      ? state.getValuePercent(fillOffset)
+      ? getValuePercent(fillOffset)
```
wingkwong, line 281: still use `state.getValuePercent`

packages/components/slider/src/use-slider.ts
```diff
@@ -248,19 +250,44 @@ export function useSlider(originalProps: UseSliderProps) {
...
+
+  if (isFixedValue && stateValues.length === 0) {
+    stateValues.push(minValue);
+  }
+
   const [startOffset, endOffset] = [
-    state.values.length > 1
-      ? state.getThumbPercent(0)
+    stateValues.length > 1
+      ? getThumbPercent(0)
```
wingkwong, line 279: still use `state.getThumbPercent`

packages/components/slider/src/use-slider.ts
```diff
@@ -248,19 +250,44 @@ export function useSlider(originalProps: UseSliderProps) {
...
 
-  const value =
-    state.values.length === 1
-      ? numberFormatter.format(state.values[0])
-      : numberFormatter.formatRange(state.values[0], state.values[state.values.length - 1]);
+  const value = isFixedValue
+    ? numberFormatter.format(minValue)
+    : stateValues.length === 1
+    ? numberFormatter.format(stateValues[0])
+    : numberFormatter.formatRange(stateValues[0], stateValues[stateValues.length - 1]);
```
wingkwong, line 290: rollback to original code

packages/components/slider/src/use-slider.ts
```diff
@@ -360,13 +388,14 @@ export function useSlider(originalProps: UseSliderProps) {
...
       renderThumb,
+      isFixedValue,
       formatOptions: tooltipValueFormatOptions,
       className: slots.thumb({class: classNames?.thumb}),
     } as SliderThumbProps;
   };
 
   const getStepProps = (index: number) => {
-    const percent = state.getValuePercent(index * step + minValue);
+    const percent = isFixedValue ? 0 : state.getValuePercent(index * step + minValue);
```
wingkwong, line 398: not required

packages/components/slider/src/use-slider.ts
```diff
@@ -379,7 +408,7 @@ export function useSlider(originalProps: UseSliderProps) {
   };
 
   const getMarkProps = (mark: SliderStepMark) => {
-    const percent = state.getValuePercent(mark.value);
+    const percent = isFixedValue ? 0 : state.getValuePercent(mark.value);
```
wingkwong, line 411: not required

packages/components/slider/src/use-slider.ts
```diff
@@ -394,12 +423,18 @@ export function useSlider(originalProps: UseSliderProps) {
       onMouseDown: (e: React.MouseEvent) => e.stopPropagation(),
       onPointerDown: (e: React.PointerEvent) => e.stopPropagation(),
       onClick: (e: any) => {
+        if (isFixedValue) {
+          e.stopPropagation();
+
+          return;
+        }
```
wingkwong, line 430: return after L432.

packages/components/slider/__tests__/slider.test.tsx
```diff
@@ -64,6 +64,26 @@ describe("Slider", () => {
...
+    );
+
+    const slider = getByRole("slider");
+    const group = getByRole("group");
+    const output = getByRole("status");
+
+    expect(slider).toHaveProperty("min", "10");
+    expect(slider).toHaveProperty("max", "10");
+
+    expect(group).toHaveAttribute("data-fixed-value", "true");
```
wingkwong, line 79: not need to check this.

packages/components/slider/src/use-slider.ts
```diff
@@ -173,6 +173,8 @@ export function useSlider(originalProps: UseSliderProps) {
     ...otherProps
   } = props;
 
+  const isFixedValue = minValue === maxValue;
```
wingkwong, line 176: probably throw a warning in console if `isFixedValue` is true

wingkwong (changes requested):

packages/components/slider/src/use-slider.ts
kaishuige: Does it mean that you only need to control the warning?

wingkwong: use `warn` from `@heroui/shared-utils` to tell users minValue and maxValue shouldn't be same

kaishuige:
It's done. I understand what you mean.

packages/components/slider/__tests__/slider.test.tsx
```diff
@@ -64,6 +64,21 @@ describe("Slider", () => {
...
+
+    const slider = getByRole("slider");
+    const output = getByRole("status");
+
+    expect(slider).not.toHaveAttribute("disabled");
+
+    expect(slider).toHaveAttribute("aria-valuetext", "10");
+    expect(output).toHaveTextContent("10");
+    expect(output).not.toHaveTextContent("NaN");
+  });
```
wingkwong, line 80: This test can pass in canary branch, which means this test case doesn't test the original issue. Ideally it should fail in canary branch and pass in your PR branch.

packages/components/slider/src/use-slider.ts
wingkwong: unnecessary change here

wingkwong (changes requested):

packages/components/slider/__tests__/slider.test.tsx
kaishuige: Sorry, I don't quite understand what you said. This is my idea and solution.

1. Modify the clampValue function to ensure that it returns minValue when isFixedValue is true. 
3. When isFixedValue is true, directly set validatedValue to minValue. 
3. Provide an empty callback function when isFixedValue is true. 
4.  5. When isFixedValue is true, set defaultValue to minValue.

Then I updated the test case.
If you have a better way, you can modify it on my basis.

wingkwong: The test case you add should fail without your changes. If it is passing, that means your test case doesn't catch the reported issue.

kaishuige: I have revised it according to my ideas. You can have a look.

packages/components/slider/src/use-slider.ts
```diff
@@ -185,31 +191,37 @@ export function useSlider(originalProps: UseSliderProps) {
...
 
   const clampValue = useCallback(
-    (valueToClamp: number) => Math.min(Math.max(valueToClamp, minValue), maxValue),
-    [minValue, maxValue],
+    (valueToClamp: number) => {
+      if (isFixedValue) return minValue;
+
+      return Math.min(Math.max(valueToClamp, minValue), maxValue);
+    },
+    [minValue, maxValue, isFixedValue],
```
wingkwong, line 199: Undo it. `clampValue` is called from `validatedValue`, where you return `minValue` if `isFixedValue` is `true`. Hence, the change you added won't take any effect.

packages/components/slider/src/use-slider.ts
```diff
@@ -185,31 +191,37 @@ export function useSlider(originalProps: UseSliderProps) {
...
   );
 
   const validatedValue = useMemo(() => {
+    if (isFixedValue) return minValue;
```
wingkwong, line 203: add a new line after it

packages/components/slider/src/use-slider.ts
```diff
@@ -185,31 +191,37 @@ export function useSlider(originalProps: UseSliderProps) {
...
     isDisabled: originalProps?.isDisabled ?? false,
     orientation,
     step,
     minValue,
     maxValue,
     numberFormatter,
-    onChange,
-    onChangeEnd,
+    onChange: isFixedValue ? () => {} : onChange,
+    onChangeEnd: isFixedValue ? () => {} : onChangeEnd,
```
wingkwong, line 224: undo

.changeset/fix-slider-nan.md
```diff
@@ -0,0 +1,5 @@
+---
+"@heroui/slider": patch
+---
+
+fix: handle slider component NaN values when min and max are the same value
```
wingkwong, line 5:
```
fixed slider component NaN values when min and max are the same value  (#5014)
```

packages/components/slider/__tests__/slider.test.tsx
```diff
@@ -64,6 +64,35 @@ describe("Slider", () => {
     expect(slider).toHaveProperty("max", "20");
   });
 
+  it("should handle fixed value when minValue and maxValue are the same", async () => {
+    const onChangeMock = jest.fn();
+
+    const {getByRole} = render(
+      <Slider
+        aria-label="Fixed Slider"
+        label="Fixed Value"
```
wingkwong, line 73: should be same

packages/components/slider/__tests__/slider.test.tsx
```diff
@@ -64,6 +64,35 @@ describe("Slider", () => {
...
+        maxValue={10}
+        minValue={10}
+        onChange={onChangeMock}
```
wingkwong, line 76: no need to test onChange

packages/components/slider/__tests__/slider.test.tsx
```diff
@@ -64,6 +64,35 @@ describe("Slider", () => {
...
+      />,
+    );
+
+    const slider = getByRole("slider");
+    const output = getByRole("status");
+
+    expect(slider).not.toHaveAttribute("disabled");
```
wingkwong, line 83: no need to test disabled

wingkwong (changes requested):

packages/components/slider/__tests__/slider.test.tsx
kaishuige: It's all taken care of. Do we need a new pr?

packages/components/slider/src/use-slider.ts
```diff
@@ -185,23 +191,28 @@ export function useSlider(originalProps: UseSliderProps) {
...
     }
 
     return clampValue(valueProp);
-  }, [valueProp, clampValue]);
+  }, [valueProp, clampValue, isFixedValue, minValue]);
 
   const state = useSliderState({
     ...otherProps,
     value: validatedValue,
+    defaultValue: otherProps.defaultValue,
```
wingkwong, line 215: doesn't it include in L213 already?

wingkwong (changes requested):

PR #5247, opened 2025-04-28
fix(accordion): double click to focus when we have input or another component

### LINKED ISSUE #4981: [BUG] - Input inside AccordionItem requires 2 clicks to focus
### HeroUI Version

2.7.4

### Describe the bug

Inside a controlled accordion, any inputs require 2 clicks to become focused and to enter any text there.
It starts working if you click anywhere inside the accordion item.
If the item is controlled but initially collapsed, it works too, because to expand it, you actually have to click on the item, and it works then.

I tried checking some events propagation on the accordion or on the item, it didn't help.

https://stackblitz.com/edit/qtnjgugf?file=App.jsx

1. Open the reproducible example on StackBlitz
2. Click on the input field
3. See nothing happens
4. Try to type something on keyboard
5. Nothing appears in the input field
6. Click again on the input field
7. Now it works

### Expected behavior

I expect the input field to become focused from the 1st click

### Screenshots or Videos

_No response_

### Operating System Version

macOS, includes Chrome, Safari and WebViews

### Browser

Chrome

Hova25:
Closes #4981

## 📝 Description

In HeroUI v2.7.4, a bug was discovered where interactive elements (like inputs) inside controlled accordion items required two clicks to become focused. This issue occurred when:

- The accordion was controlled (using `selectedKeys` or `defaultSelectedKeys`)
- The accordion item was initially expanded
- Users tried to interact with form elements inside the accordion

Users had to click twice on form elements to focus them and begin typing, leading to a poor user experience.

When we have a focusable component in a default open accordion, we need to click 2 times to focus the element. This issue occurs due to event handling not properly managing focus for interactive elements inside the accordion.

## 🚀 New behavior

When we have a focusable component in a default open accordion, we focus the element in one click. The implementation adds a new handler with clear, explicit conditions.

This solution:

1. Explicitly checks if the element is focusable with `target.tabIndex >= 0`
2. Verifies the element is not disabled with `!target.hasAttribute("disabled")`
3. Ensures the focus method exists and is a function with `typeof target.focus === "function"`
4. Only proceeds with focus handling when all three conditions are met

```js
onClickCapture: (e) => {
  const target = e.target;

  if (target instanceof HTMLElement) {
    // Check if the element is focusable and not disabled
    const isFocusable = target.tabIndex >= 0;
    const isNotDisabled = !target.hasAttribute("disabled");
    const hasFocusMethod = typeof target.focus === "function";

    if (isFocusable && isNotDisabled && hasFocusMethod) {
      e.stopPropagation();
      target.focus();
    }
  }
}
```

## Benefits

- Interactive elements inside accordion items now properly focus on the first click
- The code is more maintainable with clear variable names describing the conditions
- The logical flow is more predictable, fixing the focus behavior across all accordion use cases

No

### DIFF
--- a/.changeset/wise-trees-buy.md
+++ b/.changeset/wise-trees-buy.md
@@ -1,0 +1,5 @@
+---
+"@heroui/accordion": patch
+---
+
+fix(accordion): double click to focus when we have input or another component (#4981)
--- a/packages/components/accordion/src/use-accordion-item.ts
+++ b/packages/components/accordion/src/use-accordion-item.ts
@@ -192,6 +192,21 @@ export function useAccordionItem<T extends object = {}>(props: UseAccordionItemP
         "data-disabled": dataAttr(isDisabled),
         "data-slot": "content",
         className: slots.content({class: classNames?.content}),
+        onClickCapture: (e) => {
+          const target = e.target;
+
+          if (target instanceof HTMLElement) {
+            // Check if the element is focusable and not disabled
+            const isFocusable = target.tabIndex >= 0;
+            const isNotDisabled = !target.hasAttribute("disabled");
+            const hasFocusMethod = typeof target.focus === "function";
+
+            if (isFocusable && isNotDisabled && hasFocusMethod) {
+              e.stopPropagation();
+              target.focus();
+            }
+          }
+        },
         ...mergeProps(regionProps, props),
       };
     },
--- a/packages/components/accordion/stories/accordion.stories.tsx
+++ b/packages/components/accordion/stories/accordion.stories.tsx
@@ -461,6 +461,15 @@ export const WithForm = {
   },
 };
 
+export const WithFormAndOpenAccordion = {
+  render: WithFormTemplate,
+
+  args: {
+    ...defaultProps,
+    defaultSelectedKeys: ["1"],
+  },
+};
+
 export const CustomMotion = {
   render: Template,
 


wingkwong (changes requested):
please include tests

carlcho12 (approved):
lgtm입니다

ybbarng:
Hello, I encountered the same issue and attempted to resolve it by applying the changes from this PR. However, I ran into a new problem: a `Switch` component inside the content stopped working. After some investigation, I discovered that this new issue could be resolved by removing the line that calls `e.stopPropagation()`.

I wonder whether this line is actually necessary to fix the issue from this PR. In any case, all UI combinations must be tested before this PR is merged.

After further testing, I found that if `e.stopPropagation()` is removed, the `Switch` no longer receives focus on the first click.

Hova25:
Hello, 
In my project i have updated the solution with this

```
onClickCapture={(e) => {
  const target = e.target as HTMLElement;

  // Check if the element is focusable and not disabled
  const isFocusable = target.tabIndex >= 0;
  const isNotDisabled = !target.hasAttribute("disabled");

  if (isFocusable && isNotDisabled && target instanceof HTMLInputElement && target.type === "text") {
    e.stopPropagation();
    target.focus();
  }
}}
```

but i don't know if it is really best solution..

ybbarng:
It seems that the number type also has the same issue.

PR #5252, opened 2025-04-29
Fix/input clear button rendering

### LINKED ISSUE #5069: [BUG] - Input Clear button is not dismissed when it is pressed in mobile view
### HeroUI Version

2.6.11

### Describe the bug

My team is using the Input component from `@nextui-org/react` 2.6.11. we noticed that in mobile view, the clear button is not dissmissed after its clicked, but it is working on desktop view. i have also verified it [here](https://www.heroui.com/docs/components/input#clear-button), which i assume should be running on the latest  heroui.

_No response_

1. Create a vite reactjs project with latest version.
2. Install Tailwind CSS.
3. Install HeroUI.
4. Add an Input component with an isClearable props.
5. Toggle inspector and and run on any mobile device simulator
6. Type in the input field and click on the clear button

### Expected behavior

the clear button should be dismissed.

### Screenshots or Videos

https://github.com/user-attachments/assets/d15db172-45d3-4f13-bfbf-abf85179c26c

### Operating System Version

macOS

### Browser

Chrome

Vishvsalvi:
Closes #5069  

## 📝 Description

Issue:
In the mobile view, the clear button of the input field remains visible even after the input has been cleared. This behavior persists due to the opacity settings configured for the button.

Current Behavior:

The button has two opacity states:

0.7 when active

1.0 on hover

After tapping the clear button on mobile devices, the active opacity is removed, but the hover opacity (1.0) remains applied—despite the fact that "hover" doesn't apply in touch environments.

Proposed Solution:
Restrict the hover opacity (1.0) to larger screen sizes only. This will ensure consistent and expected behavior across mobile devices, preventing the clear button from remaining visible after the input is cleared.

## 🚀 New behavior

https://github.com/user-attachments/assets/c136066d-d510-4702-b233-a8a94edf9308

No


.changeset/soft-spoons-march.md
```diff
@@ -0,0 +1,5 @@
+---
+"@heroui/toast": patch
+---
+
+Remove the bottom extenstion of the toast
```
wingkwong, line 5: remove this file as you only modified `theme` package.

packages/core/theme/src/components/input.ts
```diff
@@ -61,7 +61,7 @@ const input = tv({
       "outline-none",
       "select-none",
       "opacity-0",
-      "hover:!opacity-100",
+      "md:hover:!opacity-100",
```
wingkwong, line 64: Normally we don't need in this way. instead, remove this line and add `hover:!opacity-100` to `size.md` (L137) and `size.lg` (L148). Just FYI. `hover` will not be applied on mobile in TW4. In TW3, currently it is applied to some browsers only.

wingkwong (changes requested):
please undo changes done in `packages/core/theme/src/components/toast.ts`. seems not related to this issue.

packages/core/theme/src/components/input.ts
Vishvsalvi: Done 👍

wingkwong (approved):

jrgarciadev (approved):

PR #5276, opened 2025-05-06
fix(image): removed extra height style

### LINKED ISSUE #3813: [BUG] - Image Component Stretches, Ignoring Parent Element Constraints
### NextUI Version

2.4.8

### Describe the bug

I've encountered an issue with the `Image` component from `@nextui-org/react` where the image stretches beyond the parent element, ignoring its constraints. This behavior is different when compared to using a native `` element, which renders the image properly within its container.

_No response_

### Steps to reproduce:

**Using the NextUI `Image` component (causes the issue):**

   ```tsx
   "use client";
   
   import React from 'react';
   import { Card, CardBody } from "@nextui-org/card";
   import { Image } from "@nextui-org/react";
   
   const Test2 = () => {
       return (
           <Card>
               <CardBody>
                   <Image
                       width={2000}
                       height={2000}
                       alt="test2"
                       src="https://tailwindui.com/img/ecommerce-images/product-page-03-product-01.jpg"
                   />
               </CardBody>
           </Card>
       );
   };
   
   export default Test2;
   ```

   In this case, the image stretches and does not fit within the parent `Card` component, no matter how I adjust the `width`/`height` properties or apply classes such as `w-full` or `h-auto`.

The `Image` component should behave like the native `` element, fitting appropriately inside the parent container without stretching.

### Expected behavior

**Using a native `` or `<Image>` (from next) tag  (works as expected):**

   ```tsx
   "use client";
   
   import React from 'react';
   import { Card, CardBody } from "@nextui-org/card";
   
   const Test1 = () => {
       return (
           <Card>
               <CardBody>
                   <img
                       width={2000}
                       height={2000}
                       alt="test1"
                       src="https://tailwindui.com/img/ecommerce-images/product-page-03-product-01.jpg"
                   />
               </CardBody>
           </Card>
       );
   };
   
   export default Test1;
   ```

   The image fits perfectly inside the `Card` component.

### Screenshots or Videos

https://github.com/user-attachments/assets/b2905fa6-1b0f-4f3a-a800-ad0fa72a9f40

### Operating System Version

Windows

### Browser

Firefox

Vishvsalvi:
Closes #3813  

## 📝 Description

Removed the additional style prop which was providing height

Screenshot 2025-05-06 143853

The current behavious is distorting the image vertically

## 🚀 New behavior
The image will only have one height prop and not the additional `style="height: Ypx"`

Screenshot 2025-05-06 143908

No

## Summary by CodeRabbit

- **Bug Fixes**
  - Removed unintended extra height styling from images to ensure correct image display.

### DIFF
--- a/.changeset/beige-boxes-cheat.md
+++ b/.changeset/beige-boxes-cheat.md
@@ -1,0 +1,5 @@
+---
+"@heroui/image": patch
+---
+
+Removed additional height style (#3813)
--- a/packages/components/image/src/use-image.ts
+++ b/packages/components/image/src/use-image.ts
@@ -116,7 +116,7 @@ export function useImage(originalProps: UseImageProps) {
 
   const domRef = useDOMRef(ref);
 
-  const {w, h} = useMemo(() => {
+  const {w} = useMemo(() => {
     return {
       w: props.width
         ? typeof props.width === "number"
@@ -162,7 +162,6 @@ export function useImage(originalProps: UseImageProps) {
       style: {
         // img has `height: auto` by default
         // passing the custom height here to override if it is specified
-        ...(otherProps?.height && {height: h}),
         ...props.style,
         ...otherProps.style,
       },
--- a/packages/components/image/src/use-image.ts
+++ b/packages/components/image/src/use-image.ts
@@ -123,13 +123,8 @@ export function useImage(originalProps: UseImageProps) {
           ? `${props.width}px`
           : props.width
         : "fit-content",
-      h: props.height
-        ? typeof props.height === "number"
-          ? `${props.height}px`
-          : props.height
-        : "auto",
     };
-  }, [props?.width, props?.height]);
+  }, [props?.width]);
 
   const showFallback = (!src || !isImgLoaded) && !!fallbackSrc;
   const showSkeleton = isLoading && !disableSkeleton;


wingkwong (changes requested):
By doing so, passing `height` to `Image` takes no effect since the default auto height would take higher precedence. Users should be able to disable aspect ratio if height is set.

Vishvsalvi:
In the storybook example, If I pass the height it takes effect and applies the height

image

See here:

image

Let me know if my understanding has failed

Vishvsalvi:
Thanks for explaining, I've understood. 

The skeleton will require height and width of the image to render.
The image will be distorted if height is passed.

One approach is that create a new skeletonHeight and skeletonWidth prop in the Image.

By this, the user will preserve the original aspect ratio without having to pass the height and avoid the image distortion.

Vishvsalvi:
> Thanks for explaining, I've understood.
> 
> The skeleton will require height and width of the image to render. The image will be distorted if height is passed.
> 
> One approach is that create a new skeletonHeight and skeletonWidth prop in the Image.
> 
> By this, the user will preserve the original aspect ratio without having to pass the height and avoid the image distortion.

@wingkwong let me know if you're open to this solution, I'll build it

wingkwong:
@Vishvsalvi I don't think it's a good idea to have extra props.

Vishvsalvi:
1. Load the image in the background, extract the width and height, and use it for skeleton.But, this will lead to poor user experience 

2. A new skeleton component 🤩

These are the final two solutions I can think of.

PR #5294, opened 2025-05-13
fix: no require.resolve in modern node

kovsu:
No `require.resolve`: https://nodejs.org/api/esm.html#differences-between-es-modules-and-commonjs

[import.meta.resolve](https://nodejs.org/api/esm.html#importmetaresolvespecifier) still has some problem here. So this is why I use `module.createRequire()`

image

## Summary by CodeRabbit

- **Chores**
	- Improved compatibility with ES module environments for package path resolution in Storybook configuration.


wingkwong:
seems not reproducible

image

PR #5383, opened 2025-06-14
refactor(Textarea): Keep consistency on the name of the type

matheusrocha89:
Closes # 

## 📝 Description

The interfaces of the properties of the components use a certain nomenclature pattern `<component-name>Props` (InputProps, ButtonProps). Textarea component was not following this pattern, to keep it consistent this PR rename it to the same pattern. From `TextAreaProps` to `TextareaProps`

Renaming `TextAreaProps` to `TextareaProps`

## 🚀 New behavior

Behavior doesn't change, it's just a renaming.

No

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected the casing of the `TextareaProps` type name across the component and related stories to ensure consistency and prevent potential type errors. No changes to component behavior or functionality.

### DIFF
--- a/packages/components/input/src/textarea.tsx
+++ b/packages/components/input/src/textarea.tsx
@@ -55,6 +55,9 @@ export interface TextareaProps extends Omit<UseInputProps<HTMLTextAreaElement>, 
   onHeightChange?: (height: number, meta: TextareaHeightChangeMeta) => void;
 }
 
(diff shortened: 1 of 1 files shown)


wingkwong:
After an internal discussion, we tend to not include breaking changes in v2. Ideally it should rename the component, not the type. Probably we may do it in v3.

PR #5405, opened 2025-06-20
fix(toast): Toast items close in reverse order

### LINKED ISSUE #5096: [BUG] - Toast Items Close in reverse order
### HeroUI Version

2.7.5

### Describe the bug

The Toast Component does function, but after some toasts are triggered, they disappear in the reverse order: the newest toast disappears at first. It is opposite to human's common cognitive.

It should be First In First Out at default, and at least a configurable prop if the developers tends to choose Last In First Out.

_No response_

1. Just go to the HeroUI website, Toast Component Page.
2. Click Toast Buttons with different colors or styles.
3. Look at the order they disappear.

### Expected behavior

The Toasts disapper `Last In First Out` instead of `First In First Out`

### Screenshots or Videos

_No response_

### Operating System Version

Windows

### Browser

Chrome

adi-ray:
Closes #5096 

## 📝 Description

Fixes the bug where toast items were closing in reverse order (LIFO instead of FIFO). The issue was in the timeout progression logic that was allowing only the oldest toast to progress its timeout, causing it to close first.

- Toast items close in LIFO order (Last In First Out)
- The oldest toast (first added) closes first
- This behavior is counterintuitive to users who expect newer toasts to disappear first

## 🚀 New behavior

- Toast items now close in proper FIFO order (First In First Out) 
- The newest toast (most recently added) closes first
- Behavior now aligns with user expectations and common UX patterns
- Maintains all existing visual styling, animations, and other functionality

**No** - This is a bug fix that corrects the intended behavior. While it changes the closing order, it fixes the behavior to match user expectations and the natural UX pattern. All existing APIs, props, and functionality remain unchanged.

## 📝 Additional Information

- **Testing:** All existing tests pass ✅
- **Compatibility:** No API changes, fully backward compatible

## Summary by CodeRabbit

- **Bug Fixes**
	- Fixed toast notifications to close in the order they appear (FIFO), ensuring older toasts dismiss before newer ones.
	- Improved toast timeout progress behavior for better consistency when multiple toasts are displayed.
	- Enhanced toast exit animations for smoother transitions when closing via timeout or close button.

### DIFF
--- a/packages/components/toast/src/use-toast.ts
+++ b/packages/components/toast/src/use-toast.ts
@@ -220,7 +220,7 @@ export function useToast<T extends ToastProps>(originalProps: UseToastProps<T>) 
         startTime.current = timestamp;
       }
 
-      if (isToastHovered || isRegionExpanded || index != 0) {
+      if (isToastHovered || isRegionExpanded || index !== 0) {
         pausedTime.current += timestamp - startTime.current;
         startTime.current = null;
         animationRef.current = requestAnimationFrame(updateProgress);
--- a/packages/components/toast/src/use-toast.ts
+++ b/packages/components/toast/src/use-toast.ts
@@ -202,6 +202,7 @@ export function useToast<T extends ToastProps>(originalProps: UseToastProps<T>) 
   }, []);
 
   const [isLoading, setIsLoading] = useState<boolean>(!!promiseProp);
+  const [isToastExiting, setIsToastExiting] = useState(false);
 
   useEffect(() => {
     if (!promiseProp) return;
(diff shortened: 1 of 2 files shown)


wingkwong (approved):

macci001:
Thanks for the PR, @adi-ray. I just have one comment.

packages/components/toast/src/use-toast.ts
```diff
@@ -220,7 +220,7 @@ export function useToast<T extends ToastProps>(originalProps: UseToastProps<T>)
         startTime.current = timestamp;
       }
 
-      if (isToastHovered || isRegionExpanded || index != total - 1) {
+      if (isToastHovered || isRegionExpanded || index !== 0) {
```
macci001:
```suggestion
      if (isToastHovered || isRegionExpanded) {
```
Now since we are planning to remove the toast by FIFO, we can keep the timer running for all the toasts. This is because with the current implementation, first toast will close and then the timer for the second toast will start. The current implementation will have following: 

https://github.com/user-attachments/assets/e54f4f9c-1018-47ca-98fe-db99076405b4

But I think keeping the timer on for all the toast will make more sense.

Also you might have to look for the proper animation for the exiting. You can use `isToastExiting` for the same.

macci001 (approved):

PR #5432, opened 2025-07-01
fix(autocomplete): auto-selection on blur when input is empty

### LINKED ISSUE #5396: [BUG] - Tabbing away from empty Autocomplete input selects the top option
### HeroUI Version

2.7.10

### Describe the bug

When tabbing away from an empty autocomplete input, the input will select the top entry in the dropdown rather than leave the input empty.

This is a departure from the React-Aria combobox behavior, which leaves the input empty when tabbing away if the input is already empty.

This bug is observable on the documentation site and does not require a reproduction.
<br/>
 
The impact of this bug is that it is impossible for a keyboard user to leave an input empty, which then breaks the functionality of optional inputs for keyboard users. This results in a somewhat severe accessibility issue.

https://www.heroui.com/docs/components/autocomplete

1. Open the Autocomplete documentation page
2. Select any autocomplete that does not use an AutocompleteSection
3. Hit tab
4. Observe that the input is no longer empty.

### Expected behavior

I expect that tabbing through an empty input should leave the input empty. I should not inadvertently enter any data that I have not consciously selected or typed.

### Screenshots or Videos

_No response_

### Operating System Version

macOS

### Browser

Chrome

adi-ray:
Closes #5396 

## 📝 Description

Fixed a bug where tabbing away from an empty Autocomplete input would automatically select the top option instead of leaving the input empty. This was causing accessibility issues for keyboard users who couldn't leave optional inputs empty.

- When an empty autocomplete input loses focus (e.g., by tabbing away), the component automatically selects the first available option
- This happens because the focus management logic sets a focused key on the first item when the dropdown opens, and the blur event saves this selection
- Keyboard users cannot leave optional autocomplete inputs empty, creating an accessibility barrier

## 🚀 New behavior

- When an empty autocomplete input loses focus, no option is automatically selected
- The input remains empty as expected
- The focused key and selected key are properly cleared when blurring from an empty input
- Keyboard users can successfully leave optional autocomplete inputs empty

**No** - This is a bug fix that corrects unintended behavior. The change makes the component behave as users would expect and improves accessibility without breaking existing functionality.

## Summary by CodeRabbit

* **Bug Fixes**
  * Resolved an issue where an autocomplete option was automatically selected when the input field was empty and lost focus. Now, no option will be selected in this scenario, preventing unintended selections.
* **Tests**
  * Added tests to verify blur behavior ensuring no option is selected when input is empty and focus is lost, and preserving selection when input has a value.  
  * Updated focus management tests to reflect new behavior for empty input scenarios.

### DIFF
--- a/packages/components/autocomplete/src/use-autocomplete.ts
+++ b/packages/components/autocomplete/src/use-autocomplete.ts
@@ -479,8 +479,8 @@ export function useAutocomplete<T extends object>(originalProps: UseAutocomplete
     state.displayValidation.isInvalid === false &&
     state.realtimeValidation.isInvalid === true;
 
-  const getInputProps = () => {
-    return {
+  const getInputProps = () =>
+    ({
       ...otherProps,
       ...inputProps,
       ...slotsProps.inputProps,
@@ -491,8 +491,7 @@ export function useAutocomplete<T extends object>(originalProps: UseAutocomplete
           ? errorMessage({isInvalid, validationErrors, validationDetails})
           : errorMessage || validationErrors?.join(" "),
       onClick: chain(slotsProps.inputProps.onClick, otherProps.onClick),
-    } as unknown as InputProps;
-  };
+    }) as unknown as InputProps;
 
   const getListBoxProps = () => {
     // Use isVirtualized prop if defined, otherwise fallback to default behavior


craigbehnke:
Sorry to nag, but I'm really looking forward to having this change be shipped. Is there an update on this?

packages/components/autocomplete/src/use-autocomplete.ts
```diff
@@ -475,8 +479,8 @@ export function useAutocomplete<T extends object>(originalProps: UseAutocomplete
     state.displayValidation.isInvalid === false &&
     state.realtimeValidation.isInvalid === true;
 
-  const getInputProps = () =>
-    ({
+  const getInputProps = () => {
+    return {
```
wingkwong: unnecessary change

packages/components/autocomplete/src/use-autocomplete.ts
```diff
@@ -487,7 +491,8 @@ export function useAutocomplete<T extends object>(originalProps: UseAutocomplete
           ? errorMessage({isInvalid, validationErrors, validationDetails})
           : errorMessage || validationErrors?.join(" "),
       onClick: chain(slotsProps.inputProps.onClick, otherProps.onClick),
-    }) as unknown as InputProps;
+    } as unknown as InputProps;
+  };
```
wingkwong: unnecessary change

wingkwong (changes requested):
This PR will alter the existing focusing the first non-disabled item behaviour. 

Expected:
- press tab
  - autocomplete is focused
  - listbox is shown with the first non-disabled item highlighted (your PR broke here)
- press tab again
  - autocomplete is closed
  - no value will be set in autocomplete

wingkwong:
@adi-ray May I know if there is any update on this PR?

wingkwong:
Closing - inactivity

Issue #5439, opened 2025-07-04
[BUG] - Form ID is not passed to checkbox input


kivojenko:
### HeroUI Version

2.7.6

### Describe the bug

In HTML it is possible to add form id to inputs that are not children of this form and they are being picked up for this form. This works as expected with HeroUI Input component, yet if you try to do the same with Checkbox, form parameter is only passed to `<label>` and therefore form cannot receive it's value.

`<Checkbox form="form-325680" />`

<img width="375" alt="Image" src="https://github.com/user-attachments/assets/39b15cfa-2446-487f-bd0b-fe54f85a78c9" />

### Steps to Reproduce the Bug or Issue

1.  Add function
```
  const onSubmit = (e: any) => {
    const data = Object.fromEntries(new FormData(e.currentTarget));
    console.log(data);
  };
```
2. Return 
```
<Checkbox form="test" name="checkbox-input" />
<Input form="test" name="text-input" />
<Form onSubmit={onSubmit} id="test">
   <Button type="submit" name="Test">Test</Button>
</Form>
```
3. Submit the form. The result in console would be `{text-input: ''}`, meaning that checkbox was not found by form.

### Expected behavior

I expect checkbox to be seen by form. To achieve this, `form="test"` needs to be passed down to <input> tag of the checkbox and not stay in label container.

### Operating System Version

MacOS

### Browser

Chrome

wingkwong:
Closing in favor of v3. If the issue is reproducible in v3, please create a new one with a minimal reproducible example.

PR #5440, opened 2025-07-04
fix(checkbox): pass form id to input tag

kivojenko:
Closes [#5439](https://github.com/heroui-inc/heroui/issues/5439)

## 📝 Description

In HTML it is possible to add form id to inputs that are not children of this form and they are being picked up for this form. 

```html
<Checkbox form="test-form" />
<Form id="test-form"/>
```

Adding form currently works for HeroUI Input component, but not for Checkbox. For Checkbox form parameter is being added to label tag, which will not make it work.

```html
<label form="test-form">
    <input type="checkbox" value="" />
</label>
```

## 🚀 New behavior

Form parameter is being passed to input tag, making it possible to use checkboxes outside of form.

```html
<label>
    <input form="test-form" type="checkbox" value="" />
</label>
```

No.

## Summary by CodeRabbit

* **New Features**
  * Checkboxes can now be associated with specific forms using a new property, allowing them to function correctly even when visually separated from the form element.
  * Added a new example demonstrating how to link a checkbox to a form when they are not nested together.

* **Documentation**
  * Updated storybook examples to showcase the new checkbox association feature.

### DIFF
--- a/packages/components/checkbox/src/use-checkbox.ts
+++ b/packages/components/checkbox/src/use-checkbox.ts
@@ -87,6 +87,7 @@ export function useCheckbox(props: UseCheckboxProps = {}) {
     children,
     icon,
     name,
+    form,
     isRequired,
     isReadOnly: isReadOnlyProp = false,
     autoFocus = false,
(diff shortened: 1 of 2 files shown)


wingkwong (changes requested):
1. Please include changeset
2. I think putting `form` into `ariaCheckboxProps` for `useReactAriaCheckboxGroupItem` and `useReactAriaCheckbox` doesn't make any effect as the returned `inputProps` won't include `form` anyway.

wingkwong:
just found it duplicate - #4979

PR #5518, opened 2025-07-24
fix(checkbox): pass form id to input tag

kivojenko:
Closes 

- [#5439](https://github.com/heroui-inc/heroui/issues/5439)

## 📝 Description
There already is similar request present [#4979](https://github.com/heroui-inc/heroui/pull/4979) but it looks abandoned.

In HTML it is possible to add form id to inputs that are not children of this form and they are being picked up for this form. 

```html
<Checkbox form="test-form" />
<Form id="test-form"/>
```

Adding form currently works for HeroUI Input component, but not for Checkbox. For Checkbox form parameter is being added to label tag, which will not make it work.

```html
<label form="test-form">
    <input type="checkbox" value="" />
</label>
```

## 🚀 New behavior

Form parameter is being passed to input tag, making it possible to use checkboxes outside of form.

```html
<label>
    <input form="test-form" type="checkbox" value="" />
</label>
```

No.

## Summary by CodeRabbit

* **New Features**
  * Checkboxes can now be associated with a specific form using the form ID, allowing inputs to be linked to forms even when rendered outside the form element.
  * Added a new example demonstrating how to use a checkbox outside of its associated form.

* **Documentation**
  * Updated stories to showcase the new checkbox and form association capability.

### DIFF
--- a/.changeset/tasty-glasses-marry.md
+++ b/.changeset/tasty-glasses-marry.md
@@ -1,0 +1,6 @@
+---
+"@heroui/checkbox": minor
+"@heroui/form": minor
+---
+
+Pass form id to checkbox input tag
--- a/packages/components/checkbox/src/use-checkbox.ts
+++ b/packages/components/checkbox/src/use-checkbox.ts
@@ -87,6 +87,7 @@ export function useCheckbox(props: UseCheckboxProps = {}) {
     children,
     icon,
     name,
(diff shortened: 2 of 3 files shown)


wingkwong (commented):
same comment. we need a general solution to handle all the cases rather than just passing one.

PR #5537, opened 2025-07-28
test: add provider tests to check if component is properly calling navigation

TheCodeDestroyer:
## 📝 Description

I have noticed in many version bumps of NextUI and now in HeroUI that suddenly components stopped respecting the navigation prop on the provider.

Right now in v2.8.1 Tabs reload the whole page.

## 🚀 New behavior

This will potentially detect the issue early before it gets released again

## 💣 Is this a breaking change (Yes/No): No

None, just some additional tests

## 📝 Additional Information

You can just see from [this query](https://github.com/heroui-inc/heroui/issues?q=sort%3Aupdated-desc%20is%3Aissue%20page%20reload) how many times this happened.

Not sure if this will actually catch the issue if not could you maybe point me in a better direction.
I know with these tests we are basically testing RA functionality but maybe this is a good thing?

P.S.: I am not really aware of all components using providers navigation so if I missed any could you please point me in the right direction.

## Summary by CodeRabbit

* **Tests**
  * Added new tests for Dropdown, Link, Listbox, and Tabs components to verify correct rendering and navigation behavior when used with or without a navigation context provider.
  * Ensured that clicking items with links triggers navigation only when the appropriate context is present.


wingkwong:
Actually this is caused by the upstream dependencies. These tests won't be useful if they release after we do.

TheCodeDestroyer:
> Actually this is caused by the upstream dependencies. These tests won't be useful if they release after we do.

Is there maybe a way to lock upstream deps?

wingkwong:
We are using fixed version but they are using caret version on their side. Anyway, I'm wrapping up v2.8.2. It will include the latest RA and hence solve the issue.

wingkwong:
Closing this PR.

TheCodeDestroyer:
@wingkwong I think this might make things easier for HeroUI in future https://github.com/adobe/react-spectrum/issues/7946#issuecomment-3271239216

wingkwong:
@TheCodeDestroyer Yes. We're informed and aware of their RFC.

PR #5590, opened 2025-08-09
fix(toast): prevent duplicate toasts when multiple providers are declared

### LINKED ISSUE #5559: [BUG] - Extra Toast
### HeroUI Version

2.8.2

### Describe the bug

https://github.com/user-attachments/assets/1cc6ec47-6251-4635-83ef-3d5eb6ddb93a

Check this screen record.

_No response_

1. Go to [HeroUI Toast](https://www.heroui.com/docs/components/toast)
2. Click any button in there
3. You can see 2 toast when you click the button once.

### Expected behavior

I expected toast was working well.

### Screenshots or Videos

https://github.com/user-attachments/assets/1cc6ec47-6251-4635-83ef-3d5eb6ddb93a

### Operating System Version

Windows

### Browser

Chrome

KumJungMin:
Closes #5559 

### 📝 Description
- Removes the extra `<ToastProvider>` declared in the example, 
- which was causing duplicate toasts when combined with the global provider. 
- This change ensures only the global provider instance is used in the docs/examples.
<br/>

_before_

https://github.com/user-attachments/assets/fa922af7-3589-4bc5-8ec1-c5d8211e9fff

<br/>

_after_

https://github.com/user-attachments/assets/4e91e4c2-fbc5-4bad-972c-52d177a2b497

<br/>

###  ⛳️ Current behavior (updates)
- Multiple `<ToastProvider>` declarations in the app cause the toast queue to render duplicates, 
- as each provider instance manages its own region rendering.

<br/>

###  🚀 New behavior
- Only the global provider instance is used. 
- The example no longer declares its own `<ToastProvider>`, preventing duplicate toast rendering.

<br/>

###  💣 Is this a breaking change (Yes/No):
- No

<br/>

###  📝 Additional Information
- As a future improvement, we can make `<ToastProvider>` internally manage a global singleton toast queue so that even if multiple providers are declared, they share the same instance and avoid duplicates:

```ts
const G = globalThis as any;
const QUEUE_KEY = "__heroui_toast_queue__";

export const getToastQueue = () => {
  if (!G[QUEUE_KEY]) {
    G[QUEUE_KEY] = new ToastQueue<ToastProps>({ maxVisibleToasts: Infinity });
  }
  return G[QUEUE_KEY];
};
```
- If the team agrees with this approach, I can apply it in a follow-up PR :)

## Summary by CodeRabbit

* **Refactor**
  * Toast notifications now appear only on blog pages (paths under /blog).
  * Other areas of the site no longer display or mount the toast container, reducing unexpected pop-ups outside the blog.
  * Existing toast behavior and interactions on blog pages remain unchanged.
  * No changes to public component APIs; users should experience a more focused UI with toasts confined to relevant blog content.

### DIFF
--- a/.changeset/empty-needles-hope.md
+++ b/.changeset/empty-needles-hope.md
@@ -1,5 +1,0 @@
----
-"@heroui/toast": patch
----
-
-prevent duplicate toasts when multiple providers are declared
--- a/.changeset/empty-needles-hope.md
+++ b/.changeset/empty-needles-hope.md
@@ -1,0 +1,5 @@
+---
+"@heroui/toast": patch
+---
+
+prevent duplicate toasts when multiple providers are declared
--- a/apps/docs/content/components/toast/close.raw.jsx
+++ b/apps/docs/content/components/toast/close.raw.jsx
@@ -3,12 +3,15 @@ import React from "react";
 
 export default function App() {
   const [toastKey, setToastKey] = React.useState([]);
+  const [isBlogPage, setIsBlogPage] = React.useState(false);
+
+  React.useEffect(() => {
+    setIsBlogPage(window.location.pathname.startsWith("/blog"));
+  }, []);
 
   return (
     <>
-      <div className="fixed z-[100]">
-        <ToastProvider />
-      </div>
+      <div className="fixed z-[100]">{isBlogPage && <ToastProvider />}</div>
       <div className="flex flex-wrap gap-2">
         <Button
           variant="flat"
--- a/apps/docs/content/components/toast/close.raw.jsx
+++ b/apps/docs/content/components/toast/close.raw.jsx
@@ -1,4 +1,4 @@
-import {addToast, Button, closeToast, closeAll} from "@heroui/react";
+import {addToast, Button, closeToast, closeAll, ToastProvider} from "@heroui/react";
 import React from "react";
 
 export default function App() {
@@ -6,7 +6,9 @@ export default function App() {
 
   return (
     <>
-      <div className="fixed z-[100]" />
+      <div className="fixed z-[100]">
+        <ToastProvider />
+      </div>
       <div className="flex flex-wrap gap-2">
         <Button
           variant="flat"
--- a/apps/docs/content/components/toast/close.raw.jsx
+++ b/apps/docs/content/components/toast/close.raw.jsx
@@ -1,4 +1,4 @@
-import {addToast, Button, closeToast, closeAll, ToastProvider} from "@heroui/react";
+import {addToast, Button, closeToast, closeAll} from "@heroui/react";
 import React from "react";
 
 export default function App() {
@@ -6,9 +6,7 @@ export default function App() {
 
   return (
     <>
-      <div className="fixed z-[100]">
-        <ToastProvider />
-      </div>
+      <div className="fixed z-[100]" />
       <div className="flex flex-wrap gap-2">
         <Button
           variant="flat"


wingkwong (changes requested):
1. if you just update doc, then no need to add changeset.
2. the changes would make the example https://heroui-rheco9d3p-heroui.vercel.app/blog/v2.8.0#toast not working. I think that is the reason why ToastProvider was included in close.raw.jsx.

KumJungMin:
> updated commit: [b6bf10d](https://github.com/heroui-inc/heroui/pull/5590/commits/b6bf10d08e6e2cd9f5bf09df9a2e5418f4e8d6ea)

Previously, ToastProvider was being rendered in both the blog page and
the higher-level layout, causing it to be registered twice.

#### Why not use a global “single registration” guard?
- A global guard (e.g., via globalThis or singleton variables) would
  prevent multiple registrations but also reduce flexibility.
- Some pages may require different Provider configurations or dynamic
  removal/re-addition, which a forced single registration would block.
- Keeping control at the application level is more maintainable.

#### Why use the current pathname-based approach?
- Restricting Provider rendering to when pathname starts with `/blog`
  prevents duplicate registration and avoids loading it on unrelated
  pages.
- This approach is simple, keeps logic in the app, and solves the issue
  without adding complexity to the library.

ChaserZ98:
> 2. the changes would make the example https://heroui-rheco9d3p-heroui.vercel.app/blog/v2.8.0#toast not working. I think that is the reason why ToastProvider was included in close.raw.jsx.

@wingkwong I am the one who wrote this.
~~That `ToastProvider` in the `close.raw.jsx` is essential for storybook test cases to work so simply removing that won't work.The duplicate toast is probably caused by another `ToastProvider` used by the doc website itself IMO.~~
I double-checked the original PR (#5125). The `ToastProvider` wasn't present there so possibly someone else added it to solve the blog issue you mentioned but unluckily caused this issue on the doc. ~~I couldn't find a clue about this change from the commit history of this file as it is all stacked up in #5473. Maybe you can help digging it out here?~~
Okay, I found it here (https://github.com/heroui-inc/heroui/pull/5473/commits/7f5b8e2ec07028d756fbf0923c5999c66a51ccaa). Kind of awkward bro :)

@KumJungMin Your solution may fix the duplicate toast issue on the doc. But it would also add some unrelated code to the toast example since the preview and the code section of the doc share the same code (if I remember correctly). Users may be confused to see this.

ChaserZ98:
Okay, I digged a bit in the doc code.
**THERE IS NO GLOBAL `ToastProvider` INSTANCE.**
If you removed the current `ToastProvider` in the `close.raw.jsx` file, the one that makes all the other toast working is the one in the `placement.raw.jsx` file. If you remove that one as well, boom, none of the toast will show up.

@wingkwong The blog one doesn't work because it doesn't include the code demo from `placement.raw.jsx` which contains the `ToastProvider`. That's why you want to include another `ToastProvider` in the `close.raw.jsx`, which caused the duplicate toast problem in the doc.

The problem behind this is that all the `ToastProvider` instances share the same `ToastQueue`. The `addToast` method puts the toast into that queue and all the `ToastProvider` instances will render the UI effects based on that queue. So basically if you have n `ToastProvider` instances on the current page, calling `addToast` once will add one toast into the `ToastQueue` but n `ToastProvider` instances will render it n times.

Maybe we could adopt the [multi toaster](https://react-hot-toast.com/docs/multi-toaster) design from `react-hot-toast`. We can assign each `ToastProvider` a unique id `toasterId` and let `addToast` accept a `toasterId` field so that each `ToastProvider` instance could know which toast in the `ToastQueue` belongs to it.

wingkwong:
Closing - agree with the multi toaster approach.

PR #5605, opened 2025-08-17
feat(toast): provide a way to close the toast itself within the endContent

ChaserZ98:
## 📝 Description

This PR changes the `endContent` attribute of the `ToastProps`. It can now not only receive a variable of type `ReactNode`, but also a function with a `onClose` param and a return type of `ReactNode`. The functionality of the `onClose` function is the same as the toast's close button's onPress event. When users call the `onClose` function, it closes the toast just like clicking on the close button.

With this PR, Users can now create customized close button in the `endContent`.

There is no way to create a close button in `endContent`. Although `closeToast` api provides a way to programmtically close a toast, it requires the toast's id which users can only access it after the toast is created. Since the toast's id won't be available when the user defines the `endContent`, there is no way to utilize the `closeToast` api to close itself.

## 🚀 New behavior

As mentioned in the description, users can now utilize the `onClose` function to customize a close button or whatever ui effect they may want within the `endContent`.

No

### DIFF
--- a/.changeset/grumpy-boxes-doubt.md
+++ b/.changeset/grumpy-boxes-doubt.md
@@ -1,0 +1,5 @@
+---
+"@heroui/toast": patch
+---
+
+Enable the toast's endContent to close the toast using the onClose function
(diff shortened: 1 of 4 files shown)


ChaserZ98:
Because, like I mentioned, you need the toast id to close it and it won't be available until you call `addToast`.

You might be able to implement it using a reference. It might be easy to manage the reference for one toast but when it comes to many toasts, that would be complicated. So exposing the `onClose` to components in `endContent` would be a much easier way, just like how users are using `onClose` within the `ModalContent` to close the modal.

PR #5623, opened 2025-08-23
fix(modal/drawer): prevent modal/drawer from closing on press instead…

### LINKED ISSUE #5616: [BUG] - Modals/Drawer are closing 'on press' instead of 'on press release'
### HeroUI Version

2.8.2

### Describe the bug

With version `2.8.2` the Modals and Dialogs are immediately closed when when the mouse is clicked, without releasing it or when a touch even is registered, without the finger being released which was not the default behaviour.

Before `2.8.2`, for example in `2.8.1`, the modals/drawers gets closed only when the mouse click/finger is released. 

This change impacts me because I used to listen on swipe back on mobile devices which was triggering a POP event in the navigation history so I was able to programaticaly close the dialogs/modals. Now, the modals/drawers closes as soon as I touch the screen, so the swipe back is no longer taken into consideration

https://www.heroui.com/docs/components/modal

1. Go to HeroUI documentation
2. Select Modal/Drawer component
3. Open a modal/drawer and click/press outside of it.
4. The modal/drawer closes as soon as it registers a click/press without waiting for the release of it.

### Expected behavior

I expect the dialog/modal to get closed when I click outside of them, but only when I realease the click/finger, not as soon as I click/touch the screen.

### Screenshots or Videos

_No response_

### Operating System Version

Windows

### Browser

Chrome

anshumandev2025:
Closes #5616 

## 📝 Description
This PR fixes a bug where the Modal/Drawer was closing on press instead of on press release, resulting in an unintended and abrupt user experience.

**Root Cause**

The issue was caused by the onHide function being triggered inside the **onInteractOutsideStart** event within useAriaOverlay. This event fires as soon as the user presses outside the modal, rather than waiting for the full press interaction (press + release), which led to the modal closing prematurely.

**Solution**

To align with expected behaviour—where a modal or drawer should only close after the complete interaction (i.e. press and release)—the onHide call has been removed from **onInteractOutsideStart**.

## Summary by CodeRabbit

- **Bug Fixes**
  - Overlays (modals, drawers) no longer close on outside press start; they now dismiss only after the outside interaction completes, avoiding premature closures. Event behavior on interaction end remains unchanged. No public API changes.

- **Tests**
  - Added tests verifying overlays close only on release (not on press start) for modal and drawer.

- **Chores**
  - Patch release entry added for the overlay fix.

### DIFF
--- a/packages/hooks/use-aria-overlay/src/index.ts
+++ b/packages/hooks/use-aria-overlay/src/index.ts
@@ -66,6 +66,10 @@ export function useAriaOverlay(props: UseAriaOverlayProps, ref: RefObject<Elemen
         }
       }
     }
+
+    if (getOverlayTypeFromRef(ref) !== "modalOrDrawer") {
+      onHide();
+    }
   };
 
   const onInteractOutside = (e: PointerEvent) => {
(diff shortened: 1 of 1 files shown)


anshumandev2025:
Thanks for you suggestion @wingkwong  i added the changeset and testcases.

PR #5637, opened 2025-08-26
fix(toast): show animation when closing all modals

### LINKED ISSUE #5620: [BUG] - Toast closeToast and closeAll closes toast without animation
### HeroUI Version

2.8.2

### Describe the bug

When using `closeToast` fn to close the toast and there is only one toast, it disappears without any animation. 

_No response_

example code

```typescript
'use client';

import { addToast, Button, closeAll, closeToast } from '@heroui/react';

function showToast() {
  const toastKey = addToast({
    title: 'Test Toast',
    endContent: (
      <>
        <Button
          onPress={() => {
            if (toastKey) {
              closeToast(toastKey);
            }
          }}
        >
          Close
        </Button>

        <Button onPress={closeAll}>Close all</Button>
      </>
    ),
  });
}

export default function TestPage() {
  return (
    <div className="flex h-dvh flex-col items-center justify-center">
      <h1>Test Page</h1>

      <Button onPress={showToast}>Show Toast</Button>
    </div>
  );
}

```

1. Click on the `Show Toast` then on the toast click `close` button
2. You will see that it just disappears without any animation
3. If there is more than one toast, only the last will disappear without animation

### Expected behavior

As a user I expect that there will be an animation when closing the toast

### Screenshots or Videos

https://github.com/user-attachments/assets/d017e61f-3295-4df2-803e-2d9254b4f1df

### Operating System Version

- macOS

### Browser

Chrome

anshumandev2025:
Closes #5620
## 📝 Description
The modals were not closing with animation because they were all closing simultaneously. I added a slight delay between each modal's closing to allow the animations to play smoothly.

https://github.com/user-attachments/assets/4de0597e-2aab-44a1-9db1-10572b035367

## Summary by CodeRabbit

* **Bug Fixes**
  * Closing all toasts now animates smoothly instead of disappearing instantly.
  * Toasts dismiss in a brief, staggered sequence, improving visual feedback and readability during bulk closures.
  * No changes to public APIs; existing integrations continue to work as before.

### DIFF
--- a/.changeset/twelve-needles-bathe.md
+++ b/.changeset/twelve-needles-bathe.md
@@ -2,4 +2,4 @@
 "@heroui/toast": patch
 ---
 
-show animation when closing all modals
+show animation when closing all modals (#5620)
--- a/.changeset/twelve-needles-bathe.md
+++ b/.changeset/twelve-needles-bathe.md
@@ -1,0 +1,5 @@
+---
+"@heroui/toast": patch
+---
+
+show animation when closing all modals
--- a/packages/components/toast/src/toast-provider.tsx
+++ b/packages/components/toast/src/toast-provider.tsx
@@ -82,7 +82,9 @@ export const closeAll = () => {
 
   const keys = globalToastQueue.visibleToasts.map((toast) => toast.key);
 
-  keys.map((key) => {
-    globalToastQueue?.close(key);
+  keys.forEach((key, index) => {
+    setTimeout(() => {
+      globalToastQueue?.close(key);
+    }, index * 100);
   });
 };


wingkwong (changes requested):
closing a single toast does't have animation

ChaserZ98:
I noticed this issue a few days ago. This is because there is no way to control the internal state of toast component due to the current design of toast component.

The `addToast`, `closeToast` and `closeAll` work by directly calling the `ToastQueue` APIs from `react-aria` (i.e. the `add` and `close` methods). The `ToastQueue` has an attribute `visibleToasts` and the toast region uses it to render all the toast components including animations of adding them and closing them. This means toasts that are playing the close animation are still in the queue and they are removed from the queue until the animation is over.

However, `closeToast` and `closeAll` will directly remove toasts from the queue without modifying the internal state of each toast to trigger their close animation. This is why there is no close animation when calling these two methods.

So a good solution would be providing a way to control the internal state of toast component, specifically `isToastExiting`, and then update `closeToast` and `closeAll` to modify that state of each toast.

wingkwong (approved):

ChaserZ98:
@wingkwong I don't see the point of merging this PR. 

1. A correct close animation should include a slide-out transition as the following video shows. Currently, only the close button of the toast could trigger this, which is one of the reasons I request #5605.

    https://github.com/user-attachments/assets/b2dee27a-9f05-493d-9a3b-631dc1e2d386

    The videos below are the current UI effects of using `closeToast` and `closeAll` after merging this PR.

    https://github.com/user-attachments/assets/03e311c1-9078-43fd-8a58-323704ef549d

    https://github.com/user-attachments/assets/ca4901c1-11c2-4072-8e8e-94142ce36ad8

    None of them shows the expected slide-out transition.

2. If this PR aims to add some sort of UI effect of closing each toast one by one to `closeAll`, it is easy to do the same by directly using the global toast queue (see code example below) and plus, you can customize the gap time of closing each toast or even add a start delay to the whole action or whatever you want. So this is completely doable with the current API. There is no need to change the current `closeAll` behavior without really solving the close animation issue.

    ```ts
    import { getToastQueue } from "@heroui/react";

    export function customCloseAll(){
        const globalToastQueue = getToastQueue();
        const keys = globalToastQueue.visibleToasts.map((toast) => toast.key);
        const startDelay = 300;
        const gap = 100;
        keys.forEach((key, index) => setTimeout(globalToastQueue?.close(key), startDelay + index * gap));
    }
    ```

Therefore, my point is this PR should be reverted and the original issue #5620 should be reopen until the `closeToast` and `closeAll` API can correctly trigger the close animation of the toast component.

wingkwong:
@ChaserZ98 thanks for pointing out. I missed slide-out transition as I was using Close the last one as a reference. I've created another PR for that.

PR #5647, opened 2025-08-30
refactor(toast): animation & toast region

wingkwong:
Closes #

### DIFF
--- a/packages/components/toast/src/toast-provider.tsx
+++ b/packages/components/toast/src/toast-provider.tsx
@@ -68,11 +68,23 @@ export const addToast = ({...props}: ToastProps & ToastOptions) => {
   return globalToastQueue.add(props);
 };
 
+const closingToasts = new Map<string, NodeJS.Timeout>();
+
 export const closeToast = (key: string) => {
   if (!globalToastQueue) {
     return;
   }
-  globalToastQueue.close(key);
+
+  if (closingToasts.has(key)) {
+    return;
+  }
+
+  const timeoutId = setTimeout(() => {
+    closingToasts.delete(key);
+    globalToastQueue?.close(key);
+  }, 300);
+
+  closingToasts.set(key, timeoutId);
 };
 
 export const closeAll = () => {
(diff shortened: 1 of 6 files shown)


packages/components/toast/src/toast-provider.tsx
```diff
@@ -68,23 +68,35 @@ export const addToast = ({...props}: ToastProps & ToastOptions) => {
...
   }
-  globalToastQueue.close(key);
+
+  if (closingToasts.has(key)) {
+    return;
+  }
+
+  const timeoutId = setTimeout(() => {
+    closingToasts.delete(key);
+    globalToastQueue?.close(key);
```
ChaserZ98, line 84: @wingkwong Hi. I am trying to rework on the multiple `ToastProvider` PR #5606 and having two questions after reviewing the new implementation.

1. Is this `globalToastQueue?.close(key)` redundant? In `use-toasts.ts` module, there is a `getToastProps` callback. It defines an `onTransitionEnd` prop which calls `state.close(key)` to remove the toast from the `ToastQueue` but it is only called when `disableAnimation` is set to `true`. I wonder why we should have this constraint? If the removal of the toast from the `ToastQueue` is handled there, why do we need extra one here?

2. The close animation is triggered as expected but I am confused about why this would trigger the re-rendering of the `ToastProvider`. If you call the `add` or `close` method of `ToastQueue`, it would trigger the render of the `ToastProvider` because the `useToastQueue` hook turns the `globalToastQueue` into React state. However, `closingToasts` does not have such a step which means any modification to it does not change its object address and hence, there won't be any state change in react. But in fact, the `closeToast` does trigger the re-render. Why is that?

wingkwong: 1. It seems so. I remember previously we got some issues on closing toasts with/without animation and probably that's why we handled separately. 
2. closeToast adds key to closingToasts map, isToastClosing(key) becomes true, isToastClosing is called and passed as isClosing prop which calls setIsToastExiting(true). This triggers re-render.

PR #5847, opened 2025-10-24
fix(system-rsc): extendVariants Fix/compound variants types

### LINKED ISSUE #5778: [BUG] - `extendVariants` does not show variants
### HeroUI Version

@heroui/react@2.8.5

### Describe the bug

After upgrading from @heroui/react@2.8.4 to @heroui/react@2.8.5, TypeScript doesn't list the custom variants I created.

_No response_

Image

### Expected behavior

I hope it lists HeroUI variants along with custom ones, like in the previous patch version.

### Screenshots or Videos

_No response_

### Operating System Version

Ubuntu

### Browser

Chrome

ITBoomBKStudio:
Closes #5778

## 📝 Description

Supersedes #5795.
Builds on the initial work by **@IsDyh01** to address the incorrect return type in `extendVariants`, and additionally fixes long-standing typing issues with **CompoundVariants** (observed since early NextUI days).

This PR:

* Ports and refines the `extendVariants` type fix.
* Corrects type inference for **CompoundVariants** when composing/extending variants.

**Credits:**
Original approach by **@IsDyh01** (commits preserved via cherry-pick / attribution).
Authored by **@ITBoomBKStudio**.

## ⚽️ Current behavior (updates)

* `extendVariants` can yield an incorrect component return type, leading to type errors when using the extended variant API.
* **CompoundVariants** types don’t consistently infer the resulting props when variants are composed or extended; edge cases mis-type or drop constraints, forcing manual casts.

## 🚀 New behavior

* `extendVariants` now preserves and exposes the correct component/props type in the returned value.
* **CompoundVariants** inference is stable when:

  * composing existing variants,
  * extending variant configs, and
  * combining multiple variant keys (no manual `as` needed).
* Added unit tests covering:

  * return type of `extendVariants`,
  * inference across composed/extended compound variants,
  * representative edge cases seen in real projects.

No runtime changes; all changes are type-level and test-only.

**No.**
The change improves type inference without removing or renaming public APIs. Existing correct usages continue to type-check. In cases where code previously relied on loose/incorrect inference, the compiler now catches mistakes earlier (desired tightening, not a breaking API change).

## 📝 Additional Information

* Motivation: This aligns Typescript inference with real-world usage patterns for extended/compound variants; it removes long-standing friction that required casts/workarounds.
* Attribution: Built on **@IsDyh01**’s PR (#5795). Happy to adjust structure (split PRs or tweak tests) per maintainer preference.

## Summary by CodeRabbit

* **Refactor**
  * Redesigned type-level variant system to be slot-aware and more expressive, with updated public signatures for variant helpers.
* **New Features**
  * Exported slot-aware variant prop shape and expanded public API to accept richer default/compound variant forms.
* **Bug Fixes**
  * More robust slot-detection logic for varied inputs.
* **Tests**
  * Test helpers accept optional style input and cover compound-variant propagation to slots.
* **Chores**
  * Added changesets preparing releases (major and minor).

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/packages/core/system-rsc/__tests__/extend-variants.test.tsx
+++ b/packages/core/system-rsc/__tests__/extend-variants.test.tsx
@@ -446,4 +446,75 @@ describe("extendVariants function - with slots", () => {
     expect(baseEl).toHaveClass("rounded-xl");
     expect(headerEl).toHaveClass("scale-150");
   });
+
+  test("should override base component slots with direct slots option", () => {
+    const Card2 = extendVariants(Card, {
+      slots: {
+        header: "!font-bold !text-lg",
+        footer: "!bg-red-500",
+      },
+    });
+
+    const {getByTestId} = render(<Card2>Card Content</Card2>);
+
+    const headerEl = getByTestId("header");
+    const footerEl = getByTestId("footer");
+
+    expect(headerEl).toHaveClass("!font-bold");
+    expect(headerEl).toHaveClass("!text-lg");
+    expect(footerEl).toHaveClass("!bg-red-500");
+  });
+
+  test("should merge direct slots with variant-based slots", () => {
+    const Card2 = extendVariants(Card, {
+      slots: {
+        header: "!font-bold",
+      },
+      variants: {
+        shadow: {
+          xl: {
+            base: "shadow-xl",
+          },
+        },
+      },
+      defaultVariants: {
+        shadow: "xl",
+      },
+    });
+
+    const {getByTestId} = render(<Card2>Card Content</Card2>);
+
+    const baseEl = getByTestId("base");
+    const headerEl = getByTestId("header");
+
+    expect(baseEl).toHaveClass("shadow-xl");
+    expect(headerEl).toHaveClass("!font-bold");
+  });
+
+  test("direct slots should override variant-based slots for the same slot", () => {
+    const Card2 = extendVariants(Card, {
+      slots: {
+        base: "!bg-blue-500",
+      },
+      variants: {
+        shadow: {
+          xl: {
+            base: "shadow-xl",
+          },
+        },
+      },
+      defaultVariants: {
+        shadow: "xl",
+      },
+    });
+
+    const {getByTestId} = render(<Card2>Card Content</Card2>);
+
+    const baseEl = getByTestId("base");
+
+    // Direct slots should be applied
+    expect(baseEl).toHaveClass("!bg-blue-500");
+    // Variant-based slots should also be applied (they merge)
+    expect(baseEl).toHaveClass("shadow-xl");
+  });
 });


ITBoomBKStudio:
> **Actionable comments posted: 1**
> 
> 🧹 Nitpick comments (2)
> > packages/core/system-rsc/src/extend-variants.d.ts (2)> `51-54`: **CompoundVariants slots support looks good; consider a small readability tweak.**
> > > You can express the value type via the existing helper to reduce repetition.
> > > Apply:
> > > ```diff
> > >  type CompoundVariants<V, SV, S> = Array<
> > >    VariantValue<V, SV> &
> > > -    ClassProp<S extends undefined ? ClassValue : ClassValue | SlotsClassValue<S>>
> > > +    ClassProp<ClassValue | GetSuggestedValues<S>>
> > >  >;
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > 
> > > `98-99`: **Use the already-bound S instead of recomputing ComponentSlots.**
> > > Keeps generics consistent and may help inference.
> > > ```diff
> > > -    CV extends CompoundVariants<V, SV, ComponentSlots<CP>>,
> > > +    CV extends CompoundVariants<V, SV, S>,
> > > ```
> 
> 📜 Review details

I’m keeping CV extends CompoundVariants<V, SV, ComponentSlots\<CP\>> on purpose.
Using S here narrows slot suggestions and validation to only the user-provided slots, while the extended component should expose all base component slots. ComponentSlots<CP> reflects the full slot surface of the base component and keeps compoundVariants authoring ergonomic and type-safe.

ITBoomBKStudio:
Added the final compoundVariants test - all cases now covered and passing
image

ITBoomBKStudio:
> **Actionable comments posted: 0**
> 
> 🧹 Nitpick comments (1)
> > packages/core/system-rsc/src/extend-variants.d.ts (1)> `103-107`: **Verify that making all keys optional is the intended behavior.**
> > > The current mapped type uses `?:` on line 104, making all keys optional—including those that may be required in the base component props `CP`. A past review raised this concern and suggested preserving `CP`'s original optionality by separating CP keys (which would retain their required/optional status) from V-only keys (which would be optional).
> > > That review was marked as addressed by updating tests to provide required props. If the intention is for extended components to always have optional props, the current implementation is correct. However, if base component contracts should be preserved (required props remain required), consider this alternative:
> > > ```ts
> > > {
> > >   // CP keys preserve their original optionality
> > >   [K in keyof CP]: K extends keyof V
> > >     ? CP[K] | StringToBoolean<keyof NonNullable<V[K]>>
> > >     : CP[K]
> > > } & {
> > >   // V-only keys are optional
> > >   [K in Exclude<keyof V, keyof CP>]?: StringToBoolean<keyof NonNullable<V[K]>>
> > > } & RefAttributes<InferRef<C>>
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > This would maintain stricter type-safety by preventing accidental omission of required base props while still allowing variant props to be optional.
> 
> 📜 Review details

Yes, this behavior is intentional for components created with extendVariants.
The goal is to make extended components easier to use - we don’t want developers to re-declare every required prop (like color, size, or variant for a Button) when composing or testing variants.

extendVariants is designed as a styling HOC, not a strict type-safe wrapper.
In this context, optional props provide flexibility and reduce boilerplate:
shorter JSX = better developer experience.

If we preserved CP’s required props, every extended component instance would force full prop specification, which defeats the purpose of variant composition.

ITBoomBKStudio:
Hey @jrgarciadev, both #5847 and #5848 are ready. They fix system-rsc type & slot inference issues. Would be great to get your review when you have a moment 👀

jrgarciadev:
@wingkwong please check this PR

.changeset/nine-apes-remain.md
```diff
@@ -0,0 +1,5 @@
+---
+"@heroui/system-rsc": major
```
wingkwong: use `patch` instead of `major`

wingkwong (changes requested):
1. The changes you added in `packages/core/system-rsc/__tests__/extend-variants.test.tsx` don't really handle the reported issue. i.e If I just run those new tests in canary branch, they are all passed.
2. Please merge your PR 5848 changes into this PR so that I could test them in one go.
3. Please also handle the following cases. 

The type error is on `to`.
```
export const CustomButton = extendVariants(Button, {});
<Button as={Link} to="/sign-in">
  Sign In
</Button>
``` 

The type error is on defaultVariants.classNames.tabList
```
const CustomTabs = extendVariants(Tabs, {
  slots: {
    tabList: 'bg-white p-0',
    base: 'w-full bg-white rounded-lg overflow-hidden',
    tab: 'py-6',
    tabContent: 'font-medium text-gray-800',
  },
  defaultVariants: {
    variant: 'underlined',
    color: 'primary',
    classNames: {
      tabList: 'bg-white p-0',
      base: 'w-full bg-white rounded-lg overflow-hidden',
      tab: 'py-6',
      tabContent: 'font-medium text-gray-800',
    },
  },
});
```

wingkwong:
@ITBoomBKStudio any updates?

ITBoomBKStudio:
> @ITBoomBKStudio any updates?

Hey, I haven’t forgotten about this and I'm planning to wrap it up this week

ITBoomBKStudio:
Hey @wingkwong,

I just pushed an update:

1. The reported issue was actually a type-level problem. The runtime behavior was already correct, so writing tests specifically for that case wouldn’t bring real value. The fix strictly addresses the incorrect type inference.

2. I merged both PRs into a single one so the full set of related changes can be reviewed together.

3. The edge cases are now handled. This part was genuinely tough and I explored several approaches and ended up with a compromise that works reliably, even though it means we lose some of the previous strictness in prop validation. Using _heroui_system.InternalForwardRefRenderFunction turned out to be the closest match to the intended behavior.

Happy to go through any part in detail if needed.

packages/core/system-rsc/src/extend-variants.d.ts
```diff
@@ -84,33 +82,35 @@ type InferRef<C> =
...
+>(
+  BaseComponent: C,
+  styles: {
+    variants?: V;
+    defaultVariants?: DV;
+    compoundVariants?: CV;
+    slots?: S;
+  },
+  opts?: Options,
+) => _heroui_system.InternalForwardRefRenderFunction<
```
wingkwong: I think we may just do `import type {InternalForwardRefRenderFunction} from "./types.js";` .

wingkwong (changes requested):
1. please combine two changeset into one and use patch instead of minor or major
2. those 2 cases mentioned in [comment](https://github.com/heroui-inc/heroui/pull/5847#pullrequestreview-3409316404) seems not resolved using this PR build. 
3. does this PR cover issue 4148? There's a few more related to extendVariants. Just wanna know which of them could be solved in this PR. 

In case you need faster communication / discussion, please free feel to ping me at discord (same user id)

mamarku:
Even though this should've been fixed in 2.8.6 according to release notes, it is not. Also, still an issue in 2.8.7. @wingkwong

ITBoomBKStudio:
Hey @wingkwong  👋

Just pushed a new commit and everything is working smoothly now.

- I created custom components, passed them into the `as` prop, and their props are inferred/validated correctly.
- The polymorphic behavior is now properly handled at the type level (runtime was already fine, but types are now aligned).

Also, regarding your note about the `to` prop for `Link`: I think that was a misunderstanding. `Link` expects `href` (not `to`), so the “type error on `to`” isn’t really a valid target for this fix, so it’s just not a supported prop for that component.

On top of that, I added tests for the relevant scenarios, and everything passes cleanly.

Lmk if you want me to highlight the specific type changes in the diff.

PR #5848, opened 2025-10-24
fix(system-rsc): correct slot detection in getSlots()

ITBoomBKStudio:
## 📝 Description

Fixes incorrect slot detection in the getSlots() helper.

This PR corrects how slot keys are collected from the variants object.
The previous implementation iterated over variant values and sometimes extracted wrong keys (for example numeric indices from arrays), which caused invalid or missing slot names in the resulting slots map.
The new logic walks through each variant group and its configs, gathering slot keys from nested objects while safely ignoring non-object values.

Related to [#5847](https://github.com/heroui-inc/heroui/pull/5847) — follow-up improvement that complements the extendVariants typing fixes.

•	getSlots() may return incorrect or incomplete slot maps.
•	compoundVariants often fail to apply classes when extending a component, because slot names do not match those of the base component.

## 🚀 New behavior
•	Correctly extracts all slot keys from variant configurations, including base component slots.
•	compoundVariants.class.{slot} now applies reliably.
•	Ignores arrays, strings, and non-plain objects to avoid noise.

No.

## 📝 Additional Information
•	Keeps compatibility with existing variant definitions.
•	Tested with components that extend base slots (value, trigger, label, input, base, etc.).
•	Complements the type-level PR [#5847](https://github.com/heroui-inc/heroui/pull/5847) which addressed CompoundVariants inference.

## Summary by CodeRabbit

* **Refactor**
  * Improved internal variant-processing to be safer and more predictable when extracting slot keys.
  * Strengthened input validation and guarded traversal to ignore malformed values and return an empty result for invalid input.
  * Ensured missing slot entries are initialized consistently; no changes to public APIs or outward behavior.

* **Chores**
  * Added a changeset entry marking a minor package release.

### DIFF
--- a/.changeset/four-guests-sneeze.md
+++ b/.changeset/four-guests-sneeze.md
@@ -1,0 +1,5 @@
+---
+"@heroui/system-rsc": minor
+---
+
+fix(system-rsc): correct slot detection in getSlots() to ensure proper slot key extraction and consistent compoundVariants behavior.


wingkwong:
As mentioned in https://github.com/heroui-inc/heroui/pull/5847 please merge the changes to the linked PR. Also please include some test cases and examples in your description (i.e. before & after). Thanks.

Issue #5844, opened 2025-10-24
[BUG] - Flickering when scrolling overflow on documentation page


Qhung3217:
### HeroUI Version

v2.8.5

### Describe the bug

When scrolling inside the **Code** tab of the **“Blurred Footer”** section on the Card documentation page, the content flickers or flashes repeatedly

### Steps to Reproduce the Bug or Issue

1. Go to 'https://www.heroui.com/docs/components/card#blurred-footer'
2. Click the **Code** tab
3. Scroll inside inside the code block area  
4. Observe the flickering effect on the code

### Expected behavior

The code block should scroll smoothly with no flickering or repaint artifacts.

### Screenshots or Videos

https://github.com/user-attachments/assets/86774eb9-f936-41ac-948c-be2c73ae81b0

### Operating System Version

Windows

### Browser

Chrome

villAsh:
Would like to work on this,
can you please assign it to me?

wingkwong:
@villAsh Assigned. Thanks.

villAsh:
@wingkwong Created this [PR](https://github.com/heroui-inc/heroui/pull/5849).

Please have a look and let me know any changes required or not.

BTW issues was in codeblock component so each and every codeblock of any component was having this issue.

PR #5849, opened 2025-10-25
fix(codeblock): codeblock flickering issue

villAsh:
Closes [Issue](https://github.com/heroui-inc/heroui/issues/5844)

## 📝 Description
When scrolling inside the Code tab of the any component section on the documentation page, the content flickers or flashes repeatedly.

It does not allows users to scroll smoothly.

## 🚀 New behavior

The code block now scroll smoothly with no flickering or repaint artifacts.

No

https://github.com/user-attachments/assets/502c888d-65d2-41d5-bcaf-d179266ac70c

## Summary by CodeRabbit

* **Bug Fixes**
  * Minor adjustment to code block rendering to more consistently preserve layout and horizontal scrolling for long lines by removing an unnecessary container class.

* **Style**
  * Tweaked code block container classes to refine overflow handling and whitespace preservation for improved readability without changing syntax highlighting.

### DIFF
--- a/apps/docs/components/docs/components/codeblock.tsx
+++ b/apps/docs/components/docs/components/codeblock.tsx
@@ -124,18 +124,10 @@ const CodeBlockHighlight = ({
                   preRef.current = element;
                 }
               }}
-              // eslint-disable-next-line prettier/prettier
-              className={clsx(
-                className,
-                classNameProp,
-                `language-${codeLang}`,
-                "max-w-full",
-                {
+              className={clsx(className, classNameProp, `language-${codeLang}`, "max-w-full", {
                 "flex-col": isMultiLine,
                 "overflow-x-scroll scrollbar-hide": hideScrollBar,
-                // eslint-disable-next-line prettier/prettier
-                },
-              )}
+              })}
               data-language={language}
               style={style}
             >


villAsh:
@wingkwong, I get it,

I'll revert a few changes and see if it works or not.

villAsh:
@wingkwong updated, please check.

apps/docs/components/docs/components/codeblock.tsx
```diff
@@ -97,6 +97,7 @@ const CodeBlockHighlight = ({
   return (
     <div
       ref={intersectionRef}
+      className="w-full overflow-hidden"
```
wingkwong: 1. `w-full` is already defined in L106.
2. what is the reason of adding `overflow-hidden`?

apps/docs/components/docs/components/codeblock.tsx
```diff
@@ -124,16 +125,10 @@ const CodeBlockHighlight = ({
...
-                "max-w-full contents",
-                {
-                  "flex-col": isMultiLine,
-                  "overflow-x-scroll scrollbar-hide": hideScrollBar,
-                },
-              )}
+              className={clsx(className, classNameProp, `language-${codeLang}`, "max-w-full", {
+                "flex-col": isMultiLine,
+                "overflow-x-scroll scrollbar-hide": hideScrollBar,
+              })}
```
wingkwong: prefer to keep the previous format if you just remove `contents`.

villAsh: I've to disable the prettier inline rule for that

apps/docs/components/docs/components/codeblock.tsx
villAsh: overflow-hidden is for that shimmer state, that it should not allow the shimmer state to scroll.

but I've removed and tested and it seems to be working without that as well.

so here is the thing, I've removed the entire className (LN: 100) as it is not required

PR #5889, opened 2025-11-09
fix(toast): hide static toast when dragging to prevent duplicates

### LINKED ISSUE #5883: [BUG] - Toast bugs. Dublicating toast, animation speed etc
### HeroUI Version

2.8.5

### Describe the bug

When you close a toast, duplicates still appear in the background, even if there's only one element to duplicate. Below is an example of how it should work.

### Heroui toast

[https://www.heroui.com/docs/components/toast](url)

remove dublicats, animation speed must be faster

### Expected behavior

I expect better toast :((

## With multiple toasts. When closing toast a little bit freezing, also have dublicats on the background

https://github.com/user-attachments/assets/76d83425-a2f6-4fd7-8186-93a9c3f5f899

## With single toast also has dublicat.

https://github.com/user-attachments/assets/17f981aa-7f26-423e-a5d4-fc09f5d5bf8a

### Operating System Version

Linux

### Browser

Chrome

hassanzadeh-mj:
- Set opacity to 0 for the static toast element when dragging starts
- Only apply opacity change to the dragged toast, not all toasts
- Fixes duplicate toast visibility issue during drag animation
- Resolves #5883

Closes # 

## Release Notes

* **New Features**
  * Added V3 Release Banner to documentation homepage highlighting HeroUI v3.0.0 (Beta).

* **Refactor**
  * Simplified toast animations by removing drag-based interactions.
  * Updated documentation site banner components and displays.
  * Removed inline version indicator from navigation bar.


packages/components/toast/src/toast.tsx
```diff
@@ -82,8 +83,16 @@ const Toast = forwardRef<"div", ToastProps>((props, ref) => {
...
+  const toastProps = getToastProps();
   const toastContent = (
-    <Component ref={domRef} {...getToastProps()}>
+    <Component
+      ref={domRef}
+      {...toastProps}
+      style={{
+        ...toastProps.style,
+        opacity: drag ? 0 : toastProps.style?.opacity,
+      }}
```
wingkwong: we don't define styles in component level. Please move it to theme.  See getToastProps

wingkwong (changes requested):

hassanzadeh-mj:
@wingkwong

To remove the underlying toast, a feature needs to be added to Framer Motion, because when dragging occurs, it takes a snapshot of the toast and displays it.
I’ve removed the drag functionality so that the toast can now be dismissed officially using a button.

packages/components/toast/src/toast.tsx
wingkwong: why is this resolved?

hassanzadeh-mj:
@wingkwong
Framer Motion does not provide any attribute or feature to remove the underlying image during drag, so I had to remove the drag functionality in order to allow the image to be removed officially via a button.

PR #5913, opened 2025-11-19
fix: isClearable handling

### LINKED ISSUE #5614: [BUG] - Make isClearable behavior more explicit in Input
### HeroUI Version

2.8.2

### Describe the bug

In the [use-input.ts](https://github.com/heroui-inc/heroui/blob/928383bcfc5f35a80010c833f655c1355349d0f2/packages/components/input/src/use-input.ts#L246) file, the following code enables `isClearable` implicitly when `onClear` is passed:

```ts
const isClearable = !!onClear || originalProps.isClearable;
```

This means that even if a developer does **not** want the clear button, it will still appear as soon as they provide an `onClear` handler.

_No response_

```js
import { Input } from "@heroui/input";

export default function Demo() {
  return (
    <Input
      label="Username"
      onClear={() => console.log("clear pressed")}
      // isClearable is NOT set here!
    />
  );
}
```

### Expected behavior

Change the behavior so that:

* `isClearable` is controlled **only by the explicit prop**.
* `onClear` is simply a callback executed when the clear button is clicked (if visible).

That is, replace with:

```ts
const isClearable = originalProps.isClearable ?? false;
```

This makes the API clearer and avoids unintended UI changes.

---

* More predictable and explicit API.
* Separation of concerns:

  * `isClearable` → UI visibility.
  * `onClear` → behavior when clicked.
* Consistent developer experience across components.

---

It may also make sense to apply the same principle in other components where similar props exist, such as **Textarea**, **Select**, etc. — so that `isClearable` (or analogous flags) are never auto-enabled by the presence of a handler.

### Screenshots or Videos

_No response_

### Operating System Version

MacOS

### Browser

Chrome

deepansh946:
Closes #5614

## 📝 Description

This PR fixes the implicit `isClearable` behavior in `Input` and `NumberInput` components. Previously, providing an `onClear` callback would automatically enable the clear button, even when `isClearable` was not explicitly set. This made the API unpredictable and violated separation of concerns.

**Before:**
- `isClearable` was implicitly enabled when `onClear` prop was provided
- Escape key clearing worked when either `isClearable` was true OR `onClear` was provided
- JSDoc comments incorrectly stated: "If `isClearable` is not explicitly set, the clear button will be shown when this prop is provided."

**Example of problematic behavior:**
```tsx
<Input
  label="Username"
  onClear={() => console.log("clear pressed")}
  // isClearable is NOT set, but clear button still appears!
/>
```

## 🚀 New behavior

**After:**
- `isClearable` is now controlled **only by the explicit prop**
- Escape key clearing only works when `isClearable` is explicitly `true`
- `onClear` is simply a callback executed when the clear button is clicked (if visible)
- JSDoc comments updated to reflect the new behavior

**Example of new explicit behavior:**
```tsx
<Input
  label="Username"
  onClear={() => console.log("clear pressed")}
  // Clear button will NOT appear unless isClearable={true} is set
/>

<Input
  label="Username"
  isClearable={true}
  onClear={() => console.log("clear pressed")}
  // Clear button WILL appear
/>
```

**Yes**

### Impact:
- Any code that relied on the implicit behavior (providing `onClear` without `isClearable`) will no longer show the clear button
- Developers must now explicitly set `isClearable={true}` to show the clear button

### Migration path:
1. Find all instances where `onClear` is provided without `isClearable`
2. Add `isClearable={true}` prop to enable the clear button
3. If you don't want the clear button but still need the callback, you can keep `onClear` without `isClearable` (the callback will still work if triggered programmatically)

**Before:**
```tsx
<Input onClear={handleClear} />
```

**After:**
```tsx
<Input isClearable onClear={handleClear} />
```

## Summary by CodeRabbit

* **Refactor**
  * Clear button now appears only when explicitly enabled via the isClearable prop; it no longer appears implicitly from an onClear handler.
  * Escape key clearing now only works when isClearable is enabled and the field is not read-only.

* **Tests**
  * Updated tests to assert the clear button is absent unless isClearable is set.

* **Chores**
  * Minor package version bumps recorded and internal docs clarified.


deepansh946:
@wingkwong I have fixed the isClearable implicit issue, please review it.

PR #5933, opened 2025-11-24
fix(link): resolve issue where useHref is not working correctly

### LINKED ISSUE #5431: HeroUIProvider's useHref is ignored due to mergeProps overriding linkProps.href
### Describe the bug
When you supply a custom `useHref` to `HeroUIProvider`, the generated href never reaches the rendered link. Internally, `mergeProps(focusProps, linkProps, otherProps)` merges the objects left → right. Because otherProps comes last—and also carries an href key—it replaces the href created inside `linkProps`, effectively silencing your `useHref`.

### Steps to reproduce
1. Wrap your app with <HeroUIProvider navigate={customNavigate} useHref={customUseHref}>
2. Render any Link/ButtonLink component that relies on that provider, or call useLink directly.
3. Observe the rendered anchor: useHref has no effect in HeroUIProvider due to href being overwritten

### LINKED ISSUE #5600: [BUG] - Absolute urls are considered relative when using useHref from react-router
### HeroUI Version

2.8.2

### Describe the bug

I'm using

```tsx
import { useHref, useNavigate } from 'react-router';

<HeroUIProvider navigate={navigate} useHref={useHref} locale="en-GB">
```

Then, when passing absolute url to menuitem href, it composes url like `https://mysite.com/a/b/https://google.com` when I'm at https://mysite.com/a/b

There should be a way to tell heroui not to pass the href value to useHref

_No response_

```tsx
<DropdownItem href="https://google.com"
```

### Expected behavior

Link pointing to `https://google.com`

### Screenshots or Videos

_No response_

### Operating System Version

- macos

### Browser

Chrome

### LINKED ISSUE #5925: [BUG] - useHref not work as expected
### HeroUI Version

2.8.5

### Describe the bug

I want to use useHref to add lang params prefix. but in rendered a tag have no prefix.

in provider
``` tsx
export default function Provider({
  children
}: Readonly<{
  children: React.ReactNode;
}>) {
  const { push, lang } = useRouter();
  const useHref = (href: string) => `/${lang}${href}`;

  return (
    <ThemeProvider attribute='class'>
      <HeroUIProvider navigate={push} useHref={useHref}>
        {children}
        <ToastProvider
          placement='top-center'
          toastProps={{
            variant: 'bordered',
            hideCloseButton: true
          }}
          maxVisibleToasts={3}
        />
      </HeroUIProvider>
    </ThemeProvider>
  )
}
```

in demo
``` tsx
'use client';

import {
  Link
} from '@heroui/react';

import useBreadcrumb from "@/hooks/use-breadcrumb";

export default function Dashboard() {
  useBreadcrumb([
    {
      label: 'Dashboard',
    }
  ]);
  
  return (
    <div>
      Dashboard
      <Link href="/admin/user">Users</Link>
    </div>
  )
}
```

rendered result

Image

useRouter is my custom hooks. the navigate result is expected. It's just that the href attribute is incorrect

_No response_

none

### Expected behavior

add lang parameter before the href

### Screenshots or Videos

_No response_

### Operating System Version

macOS

### Browser

Chrome

hasegawa-101:
- Closes #5925
- Closes #5431
- Closes #5600

## 📝 Description
This PR fixes a bug in the Link component where the useHref function provided by the HeroUIProvider was not being correctly applied.

##⛳️ Current behavior (updates)
Currently, in the useLink hook, the props objects are merged in an incorrect order. Specifically, otherProps (containing the original href) is merged after linkProps (containing the href transformed by useHref). This causes the transformed href to be overridden by the original one, resulting in the useHref functionality not working as expected.

##🚀 New behavior
The order of arguments in the mergeProps function within the useLink hook (packages/components/link/src/use-link.ts) has been corrected. By merging linkProps last, we ensure that the correctly transformed href from the useHref function takes precedence and is applied to the final rendered <a> tag. This resolves the bug and makes the Link component respect the useHref prop from the provider.

No

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where provider-supplied href transformations were not consistently applied to links.

* **Tests**
  * Added a test confirming provider-based href transformation is applied correctly.

* **Chores**
  * Added a patch-release changeset entry for the link package documenting the fix.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/.changeset/quick-pants-kick.md
+++ b/.changeset/quick-pants-kick.md
@@ -2,4 +2,4 @@
 "@heroui/link": patch
 ---
 
-Fix an issue where the useHref was not being applied correctly (#5925)
+Fix an issue where the useHref was not being applied correctly (#5925, #5431, #5600)


hasegawa-101:
@wingkwong  pushed the fix! Please check it again when you have a moment.

wingkwong:
@hasegawa-101 are you using [this example](https://codesandbox.io/p/sandbox/4238cz) to test? I downloaded it to my local and tried with `"@heroui/react": "https://pkg.pr.new/@heroui/react@5933"`. The issue is still reproducible.

wingkwong:
@hasegawa-101 lemme know once it is ready for review again. make sure you've tested it on that example first.

hasegawa-101:
@wingkwong Thanks. I've verified it in the codesandobox environment. Please review it again 🙏

[https://codesandbox.io/p/github/hasegawa-101/sparkling-frog/main](https://codesandbox.io/p/github/hasegawa-101/sparkling-frog/main)

wingkwong:
@hasegawa-101 the sandbox is not running

image

hasegawa-101:
@wingkwong 
Sorry for the back and forth. Could you please check it here?
 [https://stackblitz.com/~/github.com/hasegawa-101/sparkling-frog](https://stackblitz.com/~/github.com/hasegawa-101/sparkling-frog)

https://github.com/user-attachments/assets/6e21e00b-449a-4d76-b123-c41a0635f3ae

packages/components/link/src/use-link.ts
```diff
@@ -44,8 +44,12 @@ export type UseLinkProps = Props & AriaLinkProps;
 
 export function useLink(originalProps: UseLinkProps) {
   const globalContext = useProviderContext();
+  const resolvedProps =
+    globalContext?.useHref && originalProps.href
+      ? {...originalProps, href: globalContext.useHref(originalProps.href)}
+      : originalProps;
```
wingkwong: This logic seems to be covered in `useLinkProps` in `packages/hooks/use-aria-link/src/index.ts` already. Could you please cross check the react-spectrum implementation (`packages/@react-aria/link/src/useLink.ts`)? From what I see, they look same.

hasegawa-101:
@wingkwong 

Thanks for the feedback. I've reviewed the react-spectrum implementation as suggested.

You are correct that `useLinkProps` handles href transformation via `useHref`. However, there is an important structural dependency regarding `HeroUIProvider`.

Currently, `HeroUIProvider` allows setting `useHref` without `navigate`. In that case, `RouterProvider` is not rendered, so the transformation logic is not applied. The `resolvedProps` logic was added specifically to support this scenario where a user provides only `useHref` without `navigate`.

If we aim to align with the react-spectrum implementation, I believe we would need to require `navigate` to be set whenever `useHref` is used, rather than allowing `useHref` to be set independently.

I'd love to hear your thoughts on this 🙏

wingkwong:
@hasegawa-101 thanks for the findings. Lemme take a look when I got more time later.

wingkwong:
I also checked their implementation and compared ours. The expected behaviour is that `navigate` is required in order to use `RouterProvider` and `useHref` is optional along with it, meaning we cannot purely use `useHref` alone without `navigate`. So the existing design is correct. 

However, from those 2 linked issues, they also include `navigate` and `useHref`. In such case it should work without this PR changes. May need to investigate why here. In case you need a faster communication or discussion, free feel to DM me at discord (same user ID).

PR #5972, opened 2025-12-08
refactor: removing as-child

tianenpang:
Closes # 

## 📝 Description

This PR removes the asChild prop pattern from all HeroUI components. Previously, components used Radix UI's Slot to polymorphically render as different elements. We now use variant functions and direct BEM class application for styling flexibility.

Previously, HeroUI components supported an asChild prop that allowed rendering components as different elements:

```tsx
// Old pattern
<Button asChild>
  <Link href="/about">About</Link>
</Button>

<Card asChild>
  <a href="/details">Card Link</a>
</Card>
```

- Hard to support `renderProps` from react aria component
- Added complexity to component implementations
- Created inconsistent behavior across components
- Made type inference more difficult

## 🚀 New behavior

Components now use variant functions and direct BEM class application for polymorphic styling:

### Using variant functions:

```tsx
import { buttonVariants } from '@heroui/react';
import Link from 'next/link';

<Link
  className={buttonVariants({ variant: "primary" })}
  href="/about"
>
  About
</Link>
```

### Using BEM classes directly:

```tsx
import Link from 'next/link';

<Link className="button button--primary" href="/dashboard">
  Dashboard
</Link>
```

### For compound components:

```tsx
import { Link, linkVariants } from '@heroui/react';
import NextLink from 'next/link';

const slots = linkVariants({ underline: "hover" });

<NextLink className={slots.base()} href="/about">
  About Page
  <Link.Icon className={slots.icon()} />
</NextLink>
```

Yes.

### The Problem

After removing `asChild`, compound components still need a way to apply styles from Context to their child components. The `composeSlotClassName` utility was introduced to handle `className` composition for non-react-aria-components (regular HTML elements like `div, span, h3` etc.). In compound components, child components (like `Card.Header`, `Card.Title`, `Alert.Indicator`) need to:

1. Access style slots from Context (via `useContext`)
2. Merge the slot's `className` with user-provided `className`
3. Handle cases where slots might be `undefined`
4. Work with regular HTML elements that only accept string `className` (not render props)

### The Solution

The `composeSlotClassName` provides a safe way to compose `className` strings

## 📝 Additional Information

- Design Principles and Composition  Updated.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -216,11 +216,11 @@ const ComponentRoot = React.forwardRef<...>(({children, className, ...props}, re
   const slots = React.useMemo(() => componentVariants({...}), [...]);
 
   return (
-    <ComponentContext.Provider value={{slots}}>
+    <ComponentContext value={{slots}}>
       <ReactAriaComponent ref={ref} className={composeTwRenderProps(className, slots.base())}>
         {children}
       </ReactAriaComponent>
-    </ComponentContext.Provider>
+    </ComponentContext>
   );
 });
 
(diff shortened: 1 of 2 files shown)


daveycodez:
Hey how are we going to be able to do `<Button href={}>` ?

jrgarciadev (approved):
Excellent!

PR #6044, opened 2025-12-24
fix(component): button-group context only applies to direct children

diegonzs:
Closes # 

## 📝 Description

Fixed ButtonGroup context applying to nested buttons by marking only direct children with a `BUTTON_GROUP_CHILD` prop. Previously, all descendant buttons (including those inside modals, dropdowns, etc.) would inherit the group's size, variant, and disabled state, causing unintended styling conflicts.

ButtonGroup was applying its context (size, variant, isDisabled, fullWidth) to all descendant Button components, regardless of whether they were direct children or nested inside other components like modals or dropdowns. This caused buttons inside nested components to unexpectedly inherit the ButtonGroup's styling props.

## 🚀 New behavior

ButtonGroup now only applies context props to its direct child buttons. This is achieved by:
- Marking direct children with a `BUTTON_GROUP_CHILD` prop during rendering
- The Button component checks for this marker before using ButtonGroup context
- Nested buttons maintain their own styling and don't inherit from the group

The documentation has been updated to clarify that only direct child buttons receive ButtonGroup props.

No

## 📝 Additional Information

This fix ensures that buttons nested inside other components (like Modal, Dropdown, etc.) won't accidentally inherit ButtonGroup styling, which was causing visual inconsistencies and unexpected behavior.

### DIFF
--- a/packages/react/src/components/button/index.ts
+++ b/packages/react/src/components/button/index.ts
@@ -1,5 +1,7 @@
 import type {ComponentProps} from "react";
 
+import {BUTTON_GROUP_CHILD} from "../button-group";
+
 import {ButtonRoot} from "./button";
 
 /* -------------------------------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


jrgarciadev:
@tianenpang please review

jrgarciadev:
@diegonzs please fix the conflicts (when possible)

PR #6066, opened 2025-12-30
fix(icon-color): change Stackblitz and copy button when in gradient box

Pourghannad:


wingkwong:
can you share a before & after screenshot?

Issue #6039, opened 2025-12-23
[BUG] - v3 Dialog looks blurred


wingkwong:
### HeroUI Version

3.0.0-beta.3

### Describe the bug

originally reported from https://discord.com/channels/856545348885676062/1452471133327589456

- 3.0.0-beta.2 works fine
- occurs in google chrome

### Your Example Website or App

https://v3.heroui.com/docs/components/alert-dialog

### Steps to Reproduce the Bug or Issue

firefox 

<img width="1918" height="948" alt="Image" src="https://github.com/user-attachments/assets/5a35eb7f-c1f9-4d50-81ae-61ada35a6b5e" />

google chrome

<img width="1918" height="945" alt="Image" src="https://github.com/user-attachments/assets/47b2240d-317a-4b09-8770-0cc07dbb691a" />

### Expected behavior

not blurred

### Operating System Version

macOS

### Browser

Chrome

Issue #6113, opened 2026-01-15
[v3] The content and buttons rendered by the Modal component will be blurred.


hubingkang:
### HeroUI Version

@heroui/react

### Describe the bug

In the official website's Modal Demo, you can see that the rendered content becomes blurry.

I found that the rendering was blurry because the Modal Container and Button had `will-change: opacity, transform;` added. I just needed to change them to `will-change-auto`, and it was fine.

### Steps to Reproduce the Bug or Issue

In the official website's Modal Demo

### Expected behavior

As a default behavior, I think it should render a normal clear effect, rather than the blurry effect it has now. 

### Screenshots or Videos

<img width="581" height="262" alt="Image" src="https://github.com/user-attachments/assets/5d696315-f64e-4ccb-8005-089d3f21fa16" />

I added `will-change-auto` to the button on the right side of the image, which is obviously different when opening the Modal compared to the left side.

```jsx
 <Modal.Footer>
    <Button variant="primary" slot="close">
      取消
    </Button>
    <Button
      variant="primary"
      onClick={handleCreate}
      className="will-change-auto"
    >
      创建
    </Button>
  </Modal.Footer>
```

### Operating System Version

macOS

### Browser

Chrome

iSaulX:
Hey. I think this duplicated of #6039 I think I heard this is already fixed for the next coming update (3.0.0-beta4). 


If you think so, could you consider close this issue due it's duplicated?

PR #6136, opened 2026-01-22
fix(style): overlay content blur

tianenpang:
Closes https://github.com/heroui-inc/heroui/issues/6039, https://github.com/heroui-inc/heroui/issues/6113

## 📝 Description

This PR tightens `will-change` usage across component styles to align with CSS best practices: hints are scoped to real animation windows, avoiding long‑lived compositor hints on static/interactive elements while preserving smooth transitions.

Also fixed the content blur, keeping `will-change` permanently can cause compositing layers to accumulate and memory usage to rise, which ends up hurting performance.

N/A

## 🚀 New behavior

`will-change` is now limited to actual animation states (enter/exit) for overlays and to expanded state for accordion/disclosure panels, reducing unnecessary layer promotion while keeping animations smooth.

This also minimizes GPU/memory overhead and helps prevent long‑lived compositing side effects such as text/icon blur.

No.

## 📝 Additional Information

Ref to [Discord](https://discord.com/channels/856545348885676062/1452471133327589456)


jrgarciadev (approved):
thanks! @tianenpang

PR #6142, opened 2026-01-22
feat(separator): variant

wingkwong:
Closes # 

## 📝 Description

add `variant` to Separator. Available options are `default`, `secondary` and `tertiary`

## 🚀 New behavior

image

### DIFF
--- a/apps/docs/content/docs/react/components/(layout)/separator.mdx
+++ b/apps/docs/content/docs/react/components/(layout)/separator.mdx
@@ -1,7 +1,7 @@
 ---
 title: Separator
 description: Visually divide content sections
-
+icon: updated
 links:
   rac: Separator
   source: separator/separator.tsx


packages/styles/components/separator.css
```diff
@@ -17,6 +17,19 @@
   @apply h-full min-h-2 w-px;
 }
 
+/* Color variants */
+.separator--default {
+  @apply bg-separator;
+}
+
+.separator--secondary {
+  @apply bg-[var(--color-separator-secondary)];
```
jrgarciadev: @wingkwong do this instead

`bg-separator-secondary`

packages/styles/components/separator.css
```diff
@@ -17,6 +17,19 @@
...
+}
+
+.separator--tertiary {
+  @apply bg-[var(--color-separator-tertiary)];
```
jrgarciadev: same here `tertiary`

apps/docs/content/docs/react/components/(layout)/separator.mdx
jrgarciadev: `icon: updated`

PR #6144, opened 2026-01-23
docs: use consistent font name

stefanofa:
The example looked wrong since in the docs it is mentioned: `Note: The font names in CSS variables should match the PostScript names of your loaded fonts. Check your font package documentation or use the font names exactly as they appear in your useFonts hook.`

But actually different names were being used in `useFonts` hook and in `global.css`.


jrgarciadev:
Thank you! @stefanofa

PR #6215, opened 2026-02-08
fix(system-rsc): extendVariants rendering behavior with as

ITBoomBKStudio:
## ⚽️ Current behavior (updates)

When using the `as` prop with components created via `extendVariants`, the component was rendered as the value passed to `as` instead of rendering as the original base component.
This caused components to lose the original rendering semantics and styles.

## 🚀 New behavior

Components created with `extendVariants` now always render as the original base component.
The `as` prop no longer changes the rendered element, restoring the expected rendering behavior.

### DIFF
--- a/packages/core/system-rsc/__tests__/extend-variants.test.tsx
+++ b/packages/core/system-rsc/__tests__/extend-variants.test.tsx
@@ -2,8 +2,6 @@ import React from "react";
 import {render, screen} from "@testing-library/react";
 
 import {extendVariants} from "../src/extend-variants";
-// import {Button} from "../test-utils/extend-components";
-// import {Card} from "../test-utils/slots-component";
 import {Link, Button, Card, CardHeader, CardBody, CardFooter} from "../../react/src";
 
 const BaseButton = extendVariants(Button, {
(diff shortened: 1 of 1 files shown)


jrgarciadev:
@wingkwong please check

ITBoomBKStudio:
Hey @jrgarciadev and @wingkwong 👋

I ran into a confusing inconsistency with `extendVariants` and `ForwardedComponent`.

Depending on how the `as` prop is handled, the behavior differs between environments:
- One implementation works correctly in production but fails in Jest tests
- The other passes Jest tests but breaks rendering and styling in production

Below are the two versions that demonstrate the issue.

### Version that works in production

```ts
const ForwardedComponent = React.forwardRef((originalProps = {}, ref) => {
  const newProps = React.useMemo(
    () =>
      getClassNamesWithProps(
        {
          slots,
          variants,
          compoundVariants,
          props: originalProps,
          defaultVariants,
          hasSlots,
          opts,
        },
        [originalProps],
      ),
    [originalProps],
  );

  return React.createElement(BaseComponent, {
    ...originalProps,
    ...newProps,
    ref,
  });
});
```

### Version that works in Jest tests

```ts
const ForwardedComponent = React.forwardRef((originalProps = {}, ref) => {
  // Extract `as` prop if present
  const { as: Component = BaseComponent, ...restProps } = originalProps;

  const newProps = React.useMemo(
    () =>
      getClassNamesWithProps(
        {
          slots,
          variants,
          compoundVariants,
          props: restProps,
          defaultVariants,
          hasSlots,
          opts,
        },
        [restProps],
      ),
    [restProps],
  );

  return React.createElement(Component, {
    ...restProps,
    ...newProps,
    ref,
  });
});
```

In the second version, components render as the value passed to as instead of rendering as the original base component, which causes styling and rendering issues in production.

This PR restores consistent rendering behavior by ensuring components created with extendVariants always render as the original base component, while still allowing as to be used in the styling logic.

ITBoomBKStudio:
### Summary

I’ve completely rewritten the test file to better reflect the actual behavior and responsibilities of `extendVariants`.

In the process, I also fixed a serious issue with chained `extendVariants` inheritance, where compound variants and slot styles could not reliably reference variants defined in base components. With this fix, variant matching and slot resolution now behave consistently across multiple extension levels.

As a result, production behavior and test expectations are now aligned and predictable.

Some additional cleanup may still be needed in related files, but the core logic and tests are now in a good and stable state.

ITBoomBKStudio:
@jrgarciadev @wingkwong 
It would be really appreciated if this PR could be released as a hotfix.

At the moment, this issue causes broken behavior in one of my projects, and having this fix available sooner would help a lot.

PR #6356, opened 2026-03-25
fix(drawer): support controlled usage without trigger child

### LINKED ISSUE #6347: [BUG] - v3 Drawer cannot be controlled externally without a trigger child (state prop + useOverlayState does not open the drawer)
### HeroUI Version

@heroui/react@3.0.1

### Describe the bug

In HeroUI v3, the `Drawer` component cannot be opened programmatically without a trigger child element. The `DrawerRoot` wraps everything in React Aria's `DialogTrigger`, which expects its **first child to be a trigger** (Button) and the **second child to be the overlay** (Drawer.Backdrop). When no trigger child is present — which is the case for externally-controlled drawers — the `Drawer.Backdrop` is treated as the trigger instead of the overlay, so the drawer never opens.

This is a **v2 to v3 migration regression**. In v2, `<Drawer isOpen={isOpen} onOpenChange={onOpenChange}>` worked directly for controlled drawers without needing a trigger child. In v3, this pattern is broken.

## What I tried (none work)

1. **`state` prop with `useOverlayState`** — `<Drawer state={state}>` where `state = useOverlayState({ isOpen, onOpenChange })`. The `Drawer.Backdrop` renders with `display: none` (has the `hidden` CSS class) and never becomes visible.
2. **`isOpen`/`onOpenChange` on `Drawer.Backdrop`** — Same result: backdrop stays hidden.
3. **`defaultOpen` on `Drawer` root** — `<Drawer defaultOpen>` — Same result.
4. **`useEffect` calling `state.open()`** — Opening via the state object after mount. Same result.
## Root cause (from reading the source)

In `drawer.tsx`, `DrawerRoot` renders:

```tsx
<DrawerTriggerPrimitive {...mergeProps(props, controlledProps)}>
  {children}
</DrawerTriggerPrimitive>
```

Where `DrawerTriggerPrimitive` is `DialogTrigger` from `react-aria-components`. `DialogTrigger` requires its first child to be a pressable trigger element and its second child to be the overlay (`ModalOverlay`). Without a trigger as the first child, the overlay is never mounted as visible.

## Use case

Many apps use drawers in a **controlled, externally-driven way** — for example, a detail panel that opens when a URL parameter changes, or when a list item is selected. In these cases:
- The parent component manages `isOpen` state
- - There is no "trigger button" — the drawer opens in response to external events (URL changes, row clicks, etc.)
- - - The parent conditionally renders the drawer or passes `isOpen={true}` when it should be visible
This pattern was fully supported in v2 and is documented in the v3 migration guide under "Controlled State", but does not actually work in practice without a trigger child.

1. Create a controlled Drawer without a trigger child — only `Drawer.Backdrop` as a child of `Drawer`:
```tsx
import { Drawer, useOverlayState } from "@heroui/react";

function MyPanel({ isOpen, onClose }) {
  const state = useOverlayState({ isOpen, onOpenChange: (open) => { if (!open) onClose() } });

  return (
    <Drawer state={state}>
      <Drawer.Backdrop>
        <Drawer.Content placement="right">
          <Drawer.Dialog>
            <Drawer.Header><Drawer.Heading>Details</Drawer.Heading></Drawer.Header>
            <Drawer.Body>Content here</Drawer.Body>
          </Drawer.Dialog>
        </Drawer.Content>
      </Drawer.Backdrop>
    </Drawer>
  );
}

// Parent component:
function App() {
  const [isOpen, setIsOpen] = useState(false);
  return (
    <>
      <button onClick={() => setIsOpen(true)}>Open</button>
      <MyPanel isOpen={isOpen} onClose={() => setIsOpen(false)} />
    </>
  );
}
```

2. Click the "Open" button to set `isOpen` to `true`
3. 3. The Drawer renders in the DOM but `Drawer.Backdrop` has `display: none` (the `hidden` CSS class is applied)
4. 4. The drawer never becomes visible
Also tried: `defaultOpen`, `isOpen`/`onOpenChange` directly on `Drawer.Backdrop`, wrapping in `useEffect` calling `state.open()` — none work.

### Expected behavior

When using `<Drawer state={state}>` with `useOverlayState({ isOpen: true })` (or `isOpen`/`onOpenChange` on `Drawer.Backdrop`), the Drawer should open and be visible — even without a trigger child element.

The v3 migration guide shows a "Controlled State" example using `state` prop with `useOverlayState`, but this only works if a `Button` trigger is present as the first child of `Drawer`. Apps that control drawers externally (e.g., via URL params or parent state) cannot use the trigger pattern.

**Suggestion**: Either allow `Drawer.Backdrop` to work as a standalone controlled overlay (like `Modal.Backdrop` seems intended to), or provide a way to use `Drawer` without the `DialogTrigger` wrapper when in fully controlled mode.

### Screenshots or Videos

_No response_

### Operating System Version

macOS

### Browser

Chrome

oferitz:
## Summary
- When `state` prop is provided, skip `DialogTrigger` wrapper and provide state via `OverlayTriggerStateContext` directly
- `DialogTrigger` requires a pressable first child — without one, `Drawer.Backdrop` is misinterpreted as the trigger
- Added `ControlledWithoutTrigger` story demonstrating the fix

## Root Cause
`DrawerRoot` unconditionally wraps children in `DialogTrigger` (from react-aria-components). When no trigger child is present (controlled drawers opened by external events), the overlay never mounts as visible.

## Fix
Detect controlled mode (`state` prop) and render children inside `OverlayTriggerStateContext.Provider` instead of `DialogTrigger`. `ModalOverlay` reads overlay state from this context natively.

Fixes #6347

### DIFF
--- a/packages/react/src/components/drawer/drawer.stories.tsx
+++ b/packages/react/src/components/drawer/drawer.stories.tsx
@@ -3,12 +3,11 @@ import type {Meta} from "@storybook/react";
 import {Icon} from "@iconify/react";
 import React from "react";
 
+import {useOverlayState} from "../../hooks/use-overlay-state";
 import {Button} from "../button";
 import {Input} from "../input";
 import {Label} from "../label";
 import {TextField} from "../textfield";
-
-import {useOverlayState} from "../../hooks/use-overlay-state";
 
 import {Drawer} from "./index";
 
@@ -343,9 +342,9 @@ export const ControlledWithoutTrigger = () => {
               </Drawer.Header>
               <Drawer.Body>
                 <p>
-                  This drawer is opened externally via <code>useOverlayState</code> without a trigger
-                  child inside the Drawer root. This is useful for drawers driven by URL params,
-                  row selections, or other external events.
+                  This drawer is opened externally via <code>useOverlayState</code> without a
+                  trigger child inside the Drawer root. This is useful for drawers driven by URL
+                  params, row selections, or other external events.
                 </p>
               </Drawer.Body>
               <Drawer.Footer>
--- a/packages/react/src/components/drawer/drawer.stories.tsx
+++ b/packages/react/src/components/drawer/drawer.stories.tsx
@@ -8,6 +8,8 @@ import {Input} from "../input";
 import {Label} from "../label";
 import {TextField} from "../textfield";
 
+import {useOverlayState} from "../../hooks/use-overlay-state";
+
 import {Drawer} from "./index";
 
 export default {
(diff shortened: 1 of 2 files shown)


oferitz:
Hi @jrgarciadev, @wingkwong could you let me know why this PR was closed? The linked issue (#6347) is still open and the bug is still reproducible. Happy to address any feedback or adjust the approach if needed.

wingkwong:
@oferitz Hey this PR was automatically closed because the base branch `v3.0.2` was closed as we just released it. As we couldn't reopen it, could you please create a new PR and choose the base branch `v3.0.3`? Sorry for any inconvenience caused.

PR #6386, opened 2026-04-01
fix(calendar): close year picker when tapping the already-selected year

### LINKED ISSUE #6329: Calendar YearPicker: tapping the selected year should close the picker
## Problem

When the year picker is open and the user taps the already-selected year row, nothing happens — the picker stays open.

## Suggested fix

Tapping the currently selected year should close the picker and return to the month/day calendar, same as if the user had tapped the month+year heading.

## Why

It is a natural gesture: the user sees their year is already correct, taps it to confirm, and expects to go back. Currently they have to find and tap the heading to dismiss.

Gdhanush-13:
### What

Fixes #6329

### Problem

When the year picker is open and the user taps the already-selected year, the picker stays open instead of closing.

### Fix

Skip the redundant setFocusedDate call when the selected year already matches the focused year, and always call setIsYearPickerOpen(false). This ensures tapping the current year reliably closes the picker.

### DIFF
--- a/packages/react/src/components/calendar-year-picker/calendar-year-picker.tsx
+++ b/packages/react/src/components/calendar-year-picker/calendar-year-picker.tsx
@@ -365,9 +365,11 @@ const CalendarYearPickerGrid = ({
 
   const handleYearSelect = React.useCallback(
     (year: number) => {
-      const newDate = state.focusedDate.set({year});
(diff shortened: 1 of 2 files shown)


wingkwong:
Closing - the issue was reported in v2 and this PR is changing code in v3. The changes aren't related and the issue is not reproducible in v3.

PR #6397, opened 2026-04-03
fix(drawer): support controlled usage without trigger child

### LINKED ISSUE #6347: [BUG] - v3 Drawer cannot be controlled externally without a trigger child (state prop + useOverlayState does not open the drawer)

oferitz:
## Summary
- When `state` prop is provided, skip `DialogTrigger` wrapper and provide state via `OverlayTriggerStateContext` directly
- `DialogTrigger` requires a pressable first child — without one, `Drawer.Backdrop` is misinterpreted as the trigger
- Added `ControlledWithoutTrigger` story demonstrating the fix

## Root Cause
`DrawerRoot` unconditionally wraps children in `DialogTrigger` (from react-aria-components). When no trigger child is present (controlled drawers opened by external events), the overlay never mounts as visible.

## Fix
Detect controlled mode (`state` prop) and render children inside `OverlayTriggerStateContext.Provider` instead of `DialogTrigger`. `ModalOverlay` reads overlay state from this context natively.

Fixes #6347

> Re-opened from #6356, which was auto-closed when the `v3.0.2` branch was deleted after release.

### DIFF
--- a/packages/react/src/components/drawer/drawer.stories.tsx
+++ b/packages/react/src/components/drawer/drawer.stories.tsx
@@ -3,12 +3,11 @@ import type {Meta} from "@storybook/react";
 import {Icon} from "@iconify/react";
 import React from "react";
 
+import {useOverlayState} from "../../hooks/use-overlay-state";
 import {Button} from "../button";
 import {Input} from "../input";
 import {Label} from "../label";
 import {TextField} from "../textfield";
-
-import {useOverlayState} from "../../hooks/use-overlay-state";
 
 import {Drawer} from "./index";
 
@@ -343,9 +342,9 @@ export const ControlledWithoutTrigger = () => {
               </Drawer.Header>
               <Drawer.Body>
                 <p>
-                  This drawer is opened externally via <code>useOverlayState</code> without a trigger
-                  child inside the Drawer root. This is useful for drawers driven by URL params,
-                  row selections, or other external events.
+                  This drawer is opened externally via <code>useOverlayState</code> without a
+                  trigger child inside the Drawer root. This is useful for drawers driven by URL
+                  params, row selections, or other external events.
                 </p>
               </Drawer.Body>
               <Drawer.Footer>
--- a/packages/react/src/components/drawer/drawer.stories.tsx
+++ b/packages/react/src/components/drawer/drawer.stories.tsx
@@ -311,8 +311,6 @@ export const Controlled = () => {
 /**
  * Controlled drawer using `useOverlayState` + `Drawer` root without a trigger child.
  * This pattern is common for drawers opened by external events (URL changes, row clicks, etc.).
- * Before the fix, this would not work because `DrawerRoot` always wrapped children
- * in `DialogTrigger`, which requires a pressable trigger as its first child.
  */
 export const ControlledWithoutTrigger = () => {
   const [isOpen, setIsOpen] = React.useState(false);
--- a/packages/react/src/components/drawer/drawer.stories.tsx
+++ b/packages/react/src/components/drawer/drawer.stories.tsx
@@ -8,6 +8,8 @@ import {Input} from "../input";
 import {Label} from "../label";
 import {TextField} from "../textfield";
 
+import {useOverlayState} from "../../hooks/use-overlay-state";
+
 import {Drawer} from "./index";
 
 export default {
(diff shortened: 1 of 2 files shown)


wingkwong (commented):
I copied your `ControlledWithoutTrigger` story and run it on `3.0.3` branch and I didn't see the difference. After revisiting your reported issue statement, you mentioned the drawer is never visible but seems not the case. Also from our production documentation site, we have a [controlled state example](https://heroui.com/docs/react/components/drawer#controlled-state) that doesn't use trigger as well and I'm able to see the drawer.

image

Not sure if I understand your reported issue correctly or not. Could you please help double check again to see if it is still reproducible with latest version? If so, please share a minimal reproducible sandbox for us to test.

oferitz:
Hey @wingkwong — thanks for pushing back and asking me to re-verify.

Did a proper round of verification: built a minimal sandbox with the exact pattern from the issue on 3.0.3, and couldn't reproduce the bug — you're right, the controlled `<Drawer state={...}>` path works as expected.

Tracking the symptom down in our app, it turned out to be a leftover hack in our own wrapper — we were applying `display: none` (Tailwind `hidden`) to `Drawer.Backdrop` to suppress the built-in backdrop in favor of a portaled one. That worked in an earlier v3 beta when the panel was structured differently, but since `Drawer.Backdrop` now wraps `Drawer.Content`, it collapses the whole subtree. Our bug, not HeroUI's.

Closing this. Thanks for the careful review — and #6347 can probably be closed as well, it no longer reproduces on current v3.

PR #6410, opened 2026-04-08
Fix toast InvalidStateError when view transitions abort

aakarsh12x:
## Summary
- guard toast queue updates when `document.startViewTransition` throws or its transition promises reject
- fall back to a normal `flushSync` update when the browser aborts the transition during navigation or unmounts
- add regressions for synchronous `InvalidStateError` failures and `toast.promise` resolving after the subscriber unmounts

## Root cause
HeroUI wraps toast queue mutations in `document.startViewTransition(() => flushSync(fn))`.
When a route change, Strict Mode effect replay, or unmount interrupts that browser transition, the View Transition API can either throw synchronously or reject its transition promises with `InvalidStateError`. Because those promise rejections were left unhandled, `toast.success()` / `toast.promise()` could surface an uncaught DOMException even though the toast update itself was otherwise valid.

## Impact
This keeps the existing toast API intact while making the animation layer best-effort instead of fatal. Toasts still render and resolve normally, but aborted browser transitions no longer bubble as uncaught promise rejections.

## Validation
- `pnpm exec vitest run --config packages/react/vitest.config.ts src/components/toast/toast.test.tsx`
- `pnpm exec tsc --noEmit -p packages/react/tsconfig.json`
- `pnpm build:react`

### DIFF
--- a/packages/react/src/components/toast/toast-queue.ts
+++ b/packages/react/src/components/toast/toast-queue.ts
@@ -12,6 +12,46 @@ import {flushSync} from "react-dom";
 
 import {DEFAULT_RAC_MAX_VISIBLE_TOAST, DEFAULT_TOAST_TIMEOUT} from "./constants";
 
+type ViewTransitionLike = {
+  finished?: Promise<unknown>;
+  ready?: Promise<unknown>;
+  updateCallbackDone?: Promise<unknown>;
+};
+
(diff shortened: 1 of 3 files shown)


aakarsh12x:
Superseded by https://github.com/heroui-inc/heroui/pull/6411 so the branch/PR path no longer includes the previous prefix.

PR #6428, opened 2026-04-14
fix(a11y): add aria-labels to table icon-only action buttons

### LINKED ISSUE #6427: [Accessibility] Table component: Icon-only action buttons missing aria-label (WCAG 4.1.2)
## HeroUI Version
v3.0.2

## Describe the bug

The Table component documentation examples contain **20 icon-only action buttons** (view, edit, delete) that have **no accessible names**. These buttons contain only SVG icons with no `aria-label`, `aria-labelledby`, or visible text.

This is a **WCAG 2.1 Level A violation** under [Success Criterion 4.1.2: Name, Role, Value](https://www.w3.org/WAI/WCAG21/Understanding/name-role-value.html).

### Impact
- Screen readers announce these buttons as just **"button"** with no context
- Users relying on assistive technology cannot determine what each action does
- Developers copying these examples will create inaccessible applications

### Affected button patterns found:
- `button--icon-only button--sm button--ghost` (5 instances) = likely "view" actions
- `button--icon-only button--sm button--tertiary` (10 instances) = likely "edit" actions
- `button--icon-only button--sm button--danger-soft` (5 instances) = likely "delete" actions

All 20 buttons contain SVG icons but zero accessible text.

## Steps to Reproduce

1. Navigate to https://heroui.com/docs/react/components/table
2. Open browser DevTools
3. Inspect any icon-only action button in the table examples (edit, view, delete icons)
4. Check for `aria-label` — it is missing
5. Run a screen reader (VoiceOver/NVDA) and tab to these buttons, they announce only "button"

## Expected behavior

All icon-only buttons should include an `aria-label` describing their action:

```jsx
<Button aria-label="Edit user" variant="ghost" size="sm" isIconOnly>
  <EditIcon />
</Button>

<Button aria-label="View details" variant="tertiary" size="sm" isIconOnly>
  <EyeIcon />
</Button>

<Button aria-label="Delete user" variant="danger-soft" size="sm" isIconOnly>
  <TrashIcon />
</Button>
```

## Additional context

This was found during a manual accessibility audit of HeroUI v3 components. Since these are documentation examples, they serve as the reference implementation that developers copy,  making it critical that they demonstrate accessible patterns.

### Audit method
- Manual keyboard navigation testing
- Automated script checking for buttons without accessible names
- Chrome DevTools accessibility inspector

### Operating System
macOS

### Browser
Chrome

akinshaywai:
Closes heroui-inc/heroui#6427

Added aria-label attributes to all icon-only Button components in the table custom cells demo to fix WCAG 2.1 Level A violation (SC 4.1.2):
- Copy button: aria-label="Copy ID"
- View button: aria-label="View user"
- Edit button: aria-label="Edit user"
- Delete button: aria-label="Delete user"

Screen readers previously announced these as just "button" with no context.

Closes #


wingkwong (changes requested):
Based on your PR description, I'd expect there are only several lines of changes. Also your change won't pass the doc build. Please fix it.

PR #6432, opened 2026-04-15
fix(docs): add aria-labels to icon-only button demos

akinshaywai:
## 📝 Description

Icon-only buttons in the `Button` and `ButtonGroup` demo files had no `aria-label`, so screen readers announced them as plain "button" with no context. Since these examples are the reference implementation that developers copy, they should demonstrate the correct accessible pattern.

## ⛳️ Current behavior

All 6 icon-only demo buttons are announced as just "button" by screen readers.

## 🚀 New behavior

- `button/icon-only.tsx`: More options, Settings, Delete
- `button-group/with-icons.tsx`: Search, Add, Delete

Each button now has a descriptive `aria-label`.

No


wingkwong:
Closing - in favor of https://github.com/heroui-inc/heroui/pull/6437

PR #6448, opened 2026-04-21
fix(styles): prevent Dialog from being programmatically scrolled by focus

### LINKED ISSUE #6447: [BUG] - Modal.Dialog with scroll="inside" gets programmatically scrolled by focus scrollIntoView
### HeroUI Version

@heroui/react@3.0.3, @heroui/styles@3.0.3

### Describe the bug

When a `Modal` uses `size="cover"` (or any configuration where the Dialog's `scrollHeight > clientHeight`) and the default `scroll="inside"`, clicking any focusable element inside `Modal.Body` (Switch, Input, Button, …) causes the entire Dialog content to shift upward. `Modal.Footer` (Cancel/Save) leaves the bottom-right corner and a blank gap appears under the Dialog.

The Dialog itself should never scroll — only `Modal.Body` should. But React Aria's focus management calls `el.scrollIntoView(...)` on focus, which walks up the DOM looking for scrollable ancestors. Because `.modal__dialog--scroll-inside` uses `overflow: hidden`, the Dialog is detected as a scroll container and its `scrollTop` is set to a non-zero value, translating all its children (Body + Footer) upward.

**Live measurements from the minimal repro** (one click on a Switch inside the Body):

|                              | Before click | After click            |
| ---------------------------- | ------------ | ---------------------- |
| `Dialog.scrollTop`           | `0`          | **`220.5`**            |
| `Body.scrollTop`             | `0`          | `314`                  |
| `Dialog` bounding top/height | `{40, 810}`  | `{40, 810}` (unchanged)|
| `Body` bounding top          | `100`        | `-120.5` (Δ `-220.5`)  |
| `Footer` bounding top        | `790`        | `569.5` (Δ `-220.5`)   |

The Body/Footer delta is exactly `Dialog.scrollTop`, confirming the Dialog is being scrolled.

**Root cause**: `overflow: hidden` still creates a scroll container — the scrollbar is suppressed and the user can't scroll manually, but `scrollTop` is still writable via JS and `scrollIntoView` will target it.

**Likely fix**: change `.modal__dialog--scroll-inside` from `overflow-hidden` to `overflow-clip` in `packages/styles/components/modal.css` (line ~198). `overflow: clip` is visually identical but does not establish a scroll container, so `scrollTop` stays at `0` and `scrollIntoView` skips the element.

Browser support for `overflow: clip`: Chrome 90+, Firefox 81+, Safari 16+ — matches HeroUI v3's supported range.

Just use the following `modal.stories.tsx` to replace the existing file in `packages/react/src/components/modal`, and then run Storybook.

[modal.stories.tsx](https://github.com/user-attachments/files/26916752/modal.stories.tsx)

1. Open a Modal with `size="cover"` (`scroll="inside"` is the default).
2. Put enough content in `Modal.Body` to overflow (e.g. 40 paragraphs).
3. Add any focusable element (Switch / Input / Button) near the bottom of `Modal.Body`.
4. Click the focusable element.
5. Observe: `Modal.Footer` shifts upward, a blank gap appears at the bottom of the Dialog.
6. In DevTools: `document.querySelector('[data-slot="modal-dialog"]').scrollTop` is now non-zero.

### Expected behavior

Focusing an element inside `Modal.Body` should leave the Dialog visually static.
Only `Modal.Body` should scroll internally to bring the focused element into view.
`Modal.Footer` should remain pinned to the bottom of the Dialog.

### Screenshots or Videos

Before clicking switch: 

Image

After clicking switch:
Image

### Operating System Version

masOS

### Browser

Edge

anobaka:
Closes #6447 

## 📝 Description

Fix a visual regression in `Modal` where focusing any element inside `Modal.Body` (Switch, Input, Button, …) shifted the whole Dialog content upward, pushing `Modal.Footer` out of the bottom of the Dialog and leaving a blank gap.

The Dialog itself is never supposed to scroll — only `Modal.Body` scrolls internally. But `.modal__dialog--scroll-inside` used `overflow: hidden`, which still establishes a scroll container: the scrollbar is hidden and users can't scroll manually, but `scrollTop` is still writable via JS, and React Aria's focus management calls `element.scrollIntoView(...)` up the ancestor chain. The Dialog was being picked up as the nearest scrollable ancestor, so `Dialog.scrollTop` was set to a non-zero value, visually translating all its children (Body + Footer) upward.

With `size="cover"` + `scroll="inside"` (the default) and enough Body content to overflow:

- Clicking any focusable element inside `Modal.Body` shifts `Modal.Footer` upward.
- `document.querySelector('[data-slot="modal-dialog"]').scrollTop` becomes non-zero.
- `Body` and `Footer` `getBoundingClientRect().top` values both drop by exactly `Dialog.scrollTop` — confirming the Dialog (not just the Body) is scrolling.

Measurements from the new Storybook repro:

|                    | Before Switch click | After Switch click |
| ------------------ | ------------------- | ------------------ |
| `Dialog.scrollTop` | `0`                 | **`220.5`**        |
| `Body.scrollTop`   | `0`                 | `314`              |
| `Body` rect top    | `100`               | `-120.5`           |
| `Footer` rect top  | `790`               | `569.5`            |

## 🚀 New behavior

Change `.modal__dialog--scroll-inside` to use `overflow: clip` instead of `overflow: hidden`. `overflow: clip` is visually identical (clips overflow, no scrollbar) but does **not** establish a scroll container, so:

- `Dialog.scrollTop` stays at `0` — it can't be written by JS.
- `scrollIntoView` walks past the Dialog and correctly targets only `Modal.Body`.
- The Dialog stays visually static; only Body scrolls internally.

No

### DIFF
--- a/packages/styles/components/alert-dialog.css
+++ b/packages/styles/components/alert-dialog.css
@@ -145,7 +145,7 @@
   @apply flex w-full flex-col;
   @apply rounded-3xl bg-overlay shadow-overlay outline-none;
   @apply p-6;
-  @apply overflow-hidden;
+  @apply overflow-clip;
   @apply pointer-events-auto;
 
   &[data-placement="auto"] {
--- a/packages/styles/components/modal.css
+++ b/packages/styles/components/modal.css
@@ -195,7 +195,7 @@
  * Modifier: dialog--scroll-inside
  */
 .modal__dialog--scroll-inside {
-  @apply overflow-hidden;
+  @apply overflow-clip;
 }
 
 /**


jrgarciadev:
@wingkwong please check

anobaka:
Hi @winkwong, sorry to bother you. Just a gentle ping on this PR. Please let me know if there's anything else needed from my side

wingkwong (approved):

Issue #6497, opened 2026-05-08
[BUG] - `color="muted"` has no effect for the `Text` component


mwskwong:
### HeroUI Version

3.0.4

### Describe the bug

Setting `color="muted"` for `Text` and `Text.*` components have no effect, instead of changing the font color to `muted`.

### Your Example Website or App

https://heroui.com/docs/react/components/text#usage

### Steps to Reproduce the Bug or Issue

Issue can be observed in the doc itself. The text
> Smaller muted body copy for secondary descriptions.

should have a muted font color.

### Expected behavior

Setting `color="muted"` should change the font color of `Text` and `Text.*` component to `muted`

### Operating System Version

N/A

### Browser

Chrome

### Workaround
Make use of `text-muted` Tailwind class instead

Torsten85:
minor additional issue: the link to the docs of the new  `<Text/>` component is not in the correct alphabetical order.

wingkwong:
Fixed in 3.0.5. Please check out [breaking change](https://heroui.com/docs/react/releases/v3-0-5#%EF%B8%8F-breaking-changes).

PR #6505, opened 2026-05-11
refactor(text): rename Text to Typography

aosmcleod:
## Summary

The `Text` component's BEM class names share the `text-` prefix with Tailwind's `text-*` utility family (font-size, text-color, font-weight). `tailwind-merge` — invoked internally by `tailwind-variants` for every variant composition — treats `text--body`, `text--color-muted`, `text--weight-normal`, etc. as members of that family and dedupes them to the last one. The result is that several `<Text>` props **silently render nothing**.

see: https://github.com/heroui-inc/heroui/issues/6497

## Breaking change

- `<Text>` -> `<Typography>`
- BEM modifiers `text-*`


wingkwong:
As discussed internally, we will rename `Text` to `Typography`.

jrgarciadev (approved):

PR #6510, opened 2026-05-11
fix(toast): move tabindex from style object to HTML attribute

### LINKED ISSUE #6506: [BUG] - Toast: `tabindex` set as CSS style property has no effect — non-frontmost toasts remain keyboard-tabbable
### HeroUI Version

3.0.4

### Describe the bug

In `packages/react/src/components/toast/toast.tsx`, the `Toast` component sets `tabindex` inside the `style` useMemo object:

```ts
const style = useMemo<CSSProperties>(() => {
  return {
    // ...
    tabindex: isFrontmost ? 0 : -1,  // ← BUG: not a CSS property
    // ...
  };
}, [...]);
```

`tabindex` is an **HTML attribute**, not a CSS property. Placing it in the `style` object has zero effect on the DOM. As a result, non-frontmost (stacked/hidden) toasts remain keyboard-tabbable when they should be removed from the tab order.

### Root cause

The value should be passed as the `tabIndex` React prop on the `<ToastPrimitive>` element, not inside the `style` object.

### Suggested fix

```diff
  <ToastPrimitive
    ref={toastRef}
    aria-hidden={isHidden}
+   tabIndex={isFrontmost ? 0 : -1}
    className={...}
    style={style}
    toast={toast}
    {...rest}
  >
```

And remove `tabindex` from the `style` useMemo return value.

### Impact

- **Accessibility:** Screen reader and keyboard users can tab into hidden/stacked toasts that are visually obscured behind the frontmost toast.
- **Severity:** Every toast rendered by the default `ToastProvider` is affected.

_No response_

1. Trigger multiple toasts so that stacking occurs (more than `maxVisibleToasts`)
2. Press Tab repeatedly
3. Observe that focus moves into hidden/stacked toasts behind the frontmost one
4. Inspect the DOM: the `style` attribute contains `tabindex: 0` as a CSS property (which browsers ignore)

### Expected behavior

Only the frontmost toast should be keyboard-tabbable. Non-frontmost toasts should have `tabIndex={-1}` as an HTML attribute.

### Operating System Version

macOS

### Browser

Chrome

aosmcleod:
Closes #6506

## 📝 Description

The `Toast` component sets `tabindex` inside the `style` useMemo object. `tabindex` is an HTML attribute, not a CSS property — placing it in the `style` object has zero effect on the DOM.

As a result, non-frontmost (stacked/hidden) toasts remain keyboard-tabbable when they should be removed from the tab order.

- `tabindex: isFrontmost ? 0 : -1` is set inside the `style` object
- Browsers ignore unknown CSS properties, so this has no effect
- All toasts (including hidden/stacked ones) remain in the tab order

## 🚀 New behavior

- `tabIndex={isFrontmost ? 0 : -1}` is set as a proper React prop on `<ToastPrimitive>`
- The invalid `tabindex` entry is removed from the style useMemo
- Only the frontmost toast is keyboard-tabbable; stacked toasts are removed from tab order

No

## 📝 Additional Information

This is a one-line fix (move attribute from style to prop) with significant accessibility impact. Screen reader and keyboard users can no longer accidentally tab into hidden/stacked toasts.


wingkwong (changes requested):
See those 5 failing checks.

aosmcleod:
@wingkwong thanks for the review — fixed.

The TS error was that `ToastProps` from `react-aria-components` does not expose `tabIndex` as a typed prop, so `<ToastPrimitive tabIndex={…}>` failed typecheck (and consequently the Vercel + Build jobs as well). Moved to a `useEffect` that sets `tabIndex` imperatively on the underlying DOM node via the existing `toastRef`. Same behaviour — only the frontmost toast is in tab order — now type-safe.

All Actions checks now pass; the Vercel deploy checks are blocked on the "Authorization required to deploy" gate.

PR #6511, opened 2026-05-11
fix(toast): defer success/error toast after promise to avoid ViewTransition collision

### LINKED ISSUE #6406: [BUG] - Toast promises cause DOMException
### HeroUI Version

3.0.2

### Describe the bug

Whenever a toast promise is replaced it throws a DOMException `Uncaught (in promise) DOMException: Skipped ViewTransition due to another transition starting`

https://heroui.com/docs/react/components/toast#promise--loading

```ts
 <Button
            size="sm"
            variant="secondary"
            onPress={() => {
              const loadingId = toast("Uploading file...", {
                description: "Please wait while we upload your file",
                isLoading: true,
                timeout: 0,
              });

              setTimeout(() => {
                toast.close(loadingId);
                toast.success("File uploaded", {
                  description: "Your file has been uploaded successfully",
                });
              }, 3000);
            }}
          >
            Upload with loading
          </Button>
```

The `toast.success` call is the problem

### Expected behavior

No error

### Screenshots or Videos

_No response_

### Operating System Version

macOS

### Browser

Firefox

### LINKED ISSUE #6409: [BUG] - HeroUI Toast InvalidStateError
### HeroUI Version

3.0.2

### Describe the bug

# HeroUI Toast

## HeroUI version

- `@heroui/react@3.0.2`
- `@heroui/styles@3.0.2`

## Describe the bug

Calling `toast.success()` from `@heroui/react` (with `<Toast.Provider />` in the tree) sometimes produces an **uncaught promise rejection**:

`Uncaught (in promise) InvalidStateError: Transition was aborted because of invalid state`

The stack trace points to toast internals (e.g. `useToastState` → `reset` / `resume` via `setTimeout`), `toast-queue`, `flushSync`, and in development React’s `commitDoubleInvokeEffectsInDEV` / double-invocation of effects. The failure is **intermittent** (not on every toast).

## Actual behavior

Console shows `InvalidStateError: Transition was aborted because of invalid state` (often as an uncaught promise), with the stack involving toast state updates and React dev double-effects.

## Environment

| Package             | Version |
| ------------------- | ------- |
| `@heroui/react`     | 3.0.2   |
| `@heroui/styles`    | 3.0.2   |
| `react` / `react-dom` | 19.2.x |
| `next`              | 16.1.x  |

- **Browser:** (fill in, e.g. Chrome 131)
- **OS:** (fill in, e.g. Windows 11)

---

_No response_

## Steps to reproduce

1. Use **Next.js** App Router, **React 19**, **development** (Strict Mode on).
2. Mount `<Toast.Provider />` from `@heroui/react` (e.g. root layout / app providers).
3. On mutation success (e.g. TanStack Query `useMutation`), run `toast.success("…")` and then `router.push("…")` in the same flow, or use a `useEffect` keyed on `mutation.isSuccess` that does both.
4. Repeat saves / navigations until it reproduces (may take several attempts).

Minimal pattern:

```tsx
useEffect(() => {
  if (mutation.isSuccess) {
    toast.success("OK");
    router.push("/next-route");
  }
}, [mutation.isSuccess, router]);
```

## Expected behavior

Toasts should enqueue and display without throwing `InvalidStateError` or leaving an **unhandled** rejected promise in the console.

### Screenshots or Videos

Image

### Operating System Version

Windows

### Browser

Chrome

### LINKED ISSUE #6507: [BUG] - Toast: `toast.promise()` triggers two ViewTransitions in same microtask, causing guaranteed DOMException
### HeroUI Version

3.0.4

### Describe the bug

In `packages/react/src/components/toast/toast-queue.ts`, the `toast.promise()` method calls `queue.close(loadingId)` followed immediately by `toastFn.success(message)` in the same `.then()` callback:

```ts
promiseFn.then((data) => {
  queue.close(loadingId);          // triggers startViewTransition + flushSync
  return toastFn.success(message); // triggers ANOTHER startViewTransition + flushSync
});
```

Both `close()` and the variant methods (`success`, `danger`) go through the `wrapUpdate` function which calls `document.startViewTransition(() => flushSync(fn))`. The View Transitions API only allows **one active transition at a time** — starting a second one while the first is still active aborts the first with:

```
DOMException: Skipped ViewTransition due to another transition starting
```

This is not a race condition — it's a **guaranteed crash** on every resolved/rejected promise toast because both operations execute synchronously in the same microtask.

### Root cause

The `ToastQueue` constructor sets up `wrapUpdate` to use `document.startViewTransition`:

```ts
wrapUpdate: (fn: () => void) => {
  if ("startViewTransition" in document) {
    document.startViewTransition(() => {
      flushSync(fn);
    });
  } else {
    fn();
  }
},
```

When `promise()` resolves, it calls `close` then `success` synchronously — each triggers a separate ViewTransition.

### Suggested fix

Option A: Batch both operations into a single ViewTransition:
```ts
promiseFn.then((data) => {
  const successMessage = typeof options.success === "function" ? options.success(data) : options.success;
  // Single transition for both close + add
  document.startViewTransition(() => {
    flushSync(() => {
      queue.close(loadingId);
      queue.add({ title: successMessage, variant: "success" }, { timeout: DEFAULT_TOAST_TIMEOUT });
    });
  });
});
```

Option B: Defer the second operation with `queueMicrotask`:
```ts
promiseFn.then((data) => {
  queue.close(loadingId);
  queueMicrotask(() => {
    toastFn.success(successMessage);
  });
});
```

### Impact

- Every use of `toast.promise()` throws an unhandled DOMException in browsers that support View Transitions (Chrome 111+, Edge 111+).
- Related to #6406 (symptom report) but this identifies the architectural root cause.

https://heroui.com/docs/react/components/toast#promise--loading

1. Use `toast.promise(fetch(...), { loading: "...", success: "...", error: "..." })`
2. Let the promise resolve
3. Open browser console
4. Observe: `DOMException: Skipped ViewTransition due to another transition starting`

### Expected behavior

Promise toasts should transition from loading → success/error without throwing DOMExceptions.

### Operating System Version

macOS

### Browser

Chrome

aosmcleod:
- Closes #6507
- Closes #6406 
- Closes #6409

## 📝 Description

When `toast.promise()` resolves/rejects, it calls `queue.close(loadingId)` followed immediately by `toastFn.success(message)` in the same `.then()` callback. Both operations trigger `document.startViewTransition()` via the queue's `wrapUpdate`. The View Transitions API only allows one active transition — starting a second aborts the first with:

```
DOMException: Skipped ViewTransition due to another transition starting
```

This is a guaranteed crash on every resolved/rejected promise toast in browsers that support View Transitions (Chrome 111+, Edge 111+).

```ts
promiseFn.then((data) => {
  queue.close(loadingId);          // triggers startViewTransition
  return toastFn.success(message); // triggers ANOTHER startViewTransition → DOMException
});
```

## 🚀 New behavior

```ts
promiseFn.then((data) => {
  queue.close(loadingId);          // triggers startViewTransition
  setTimeout(() => {
    toastFn.success(message);      // deferred to next task — no collision
  }, 0);
});
```

The success/error toast is deferred to the next task via `setTimeout(fn, 0)` so the close transition completes before the new toast transition starts.

No

## 📝 Additional Information

- Related to #6406 (symptom report) — this PR fixes the architectural root cause.
- The `setTimeout` approach is the minimal fix. A more comprehensive solution could batch both operations into a single ViewTransition, but that would require deeper changes to the queue's internal API.


wingkwong (changes requested):
The issue is still reproducible in Firefox.

image

aosmcleod:
@wingkwong thanks for catching the Firefox case — re-investigated and went deeper this time.

Firefox 142+ (Oct 2025) now ships the View Transitions API, so the same `DOMException: Skipped ViewTransition due to another transition starting` reproduces there too. The `setTimeout(0)` happened to work in Chrome but Firefox's VT state machine has different timing — the second transition still fires while the first is in capture/setup.

Rewrote the fix at the queue level rather than the call site. `defaultWrapUpdate` now tracks the in-flight `ViewTransition` and awaits its `.finished` promise before starting the next:

```ts
let activeTransition: ViewTransition | null = null;

const defaultWrapUpdate = (fn: () => void): void => {
  if (typeof document === "undefined" || !("startViewTransition" in document)) {
    fn();
    return;
  }
  const runTransition = () => {
    const transition = document.startViewTransition(() => flushSync(fn));
    activeTransition = transition;
    transition.finished.finally(() => {
      if (activeTransition === transition) activeTransition = null;
    });
  };
  if (activeTransition) {
    activeTransition.finished.then(runTransition, runTransition);
  } else {
    runTransition();
  }
};
```

This serializes _every_ successive mutation through the queue, not just `toast.promise()` — any future call site doing close+add back-to-back is collision-free. The call-site `setTimeout` deferral in `createToastFunction` is reverted accordingly.

Verified locally against Chrome 111+, Edge 111+, and Firefox 142+.

wingkwong (changes requested):
Still able to reproduce in this PR build in Firefox (`151.0.1 (aarch64)`)

https://heroui-acdbiwx1d-heroui.vercel.app/docs/react/components/toast#promise--loading

image

wingkwong:
Please click re-request review once it is ready for review. Thanks.

aosmcleod:
@wingkwong dug into the Firefox repro more carefully and found a bug in the previous chain implementation. The original code only attached new ViewTransitions to the *first* active transition's `.finished` — when three or more mutations stack inside one microtask (which happens in \`toast.promise()\` while the loading toast's own add transition is still in flight), all stacked callbacks fire when that first transition finishes and the later ones unconditionally call `startViewTransition` again, colliding with each other.

Replaced with a tail-extending promise chain:

```ts
let transitionChain: Promise<unknown> = Promise.resolve();

const defaultWrapUpdate = (fn: () => void): void => {
  if (typeof document === "undefined" || !("startViewTransition" in document)) {
    fn();
    return;
  }
  const runNext = (): Promise<unknown> => {
    const transition = document.startViewTransition(() => {
      flushSync(fn);
    });
    return transition.finished.catch(() => {});
  };
  transitionChain = transitionChain.then(runNext, runNext);
};
```

Each new transition always awaits the *latest* one — chain stays strictly ordered no matter how many mutations stack.

Verified in Chromium (Playwright against the toast PromiseToast story): on the same flow that previously emitted 16 `Transition was skipped` / `AbortError` console events across one single-promise + four rapid stacked promise clicks, the new version emits **zero**. The Firefox issue you reported should be resolved by the same mechanism (the bug wasn't Firefox-specific, just more sensitive to the timing).

Ready for re-review.

wingkwong (approved):

PR #6512, opened 2026-05-11
fix(toast): clean up stale toastHeights entries when toasts are dismissed

### LINKED ISSUE #6508: [BUG] - Toast: `toastHeights` state never cleaned up — memory leak in long-running sessions
### HeroUI Version

3.0.4

### Describe the bug

In `packages/react/src/components/toast/toast.tsx`, the `ToastProvider` maintains a `toastHeights` state record that tracks the measured height of each toast by key:

```ts
const [toastHeights, setToastHeights] = React.useState<Record<string, number>>({});

const handleToastHeightChange = useCallback((key: string, height: number) => {
  setToastHeights((prev) => ({...prev, [key]: height}));
}, []);
```

When a toast is dismissed and removed from `visibleToasts`, its entry in `toastHeights` is **never deleted**. Over a long-running session with many toasts shown and dismissed, this object grows unbounded — a classic memory leak.

### Root cause

There is no cleanup mechanism. The `handleToastHeightChange` callback only adds/updates entries. Nothing removes entries when toasts leave the queue.

### Suggested fix

Add a cleanup effect that removes stale keys:

```ts
useEffect(() => {
  const visibleKeys = new Set(toastQueue.visibleToasts.map((t) => String(t.key)));
  setToastHeights((prev) => {
    const next: Record<string, number> = {};
    for (const key of Object.keys(prev)) {
      if (visibleKeys.has(key)) {
        next[key] = prev[key];
      }
    }
    return Object.keys(next).length === Object.keys(prev).length ? prev : next;
  });
}, [toastQueue.visibleToasts]);
```

Or alternatively, clean up in the `handleToastHeightChange` callback by checking if the key is still in `visibleToasts`.

### Impact

- **Memory:** In apps that show many toasts over time (e.g., real-time dashboards, chat apps), the `toastHeights` state object grows without bound.
- **Performance:** Each new entry triggers a state update that re-renders the entire toast region, and the growing object increases the cost of each subsequent render.
- **Severity:** Low immediate impact but degrades over time in long-running sessions.

_No response_

1. Mount `<Toast.Provider />`
2. In a loop, trigger 100+ toasts with short timeouts (e.g., `toast("msg", { timeout: 1000 })`)
3. After all toasts have dismissed, inspect React DevTools
4. The `toastHeights` state in `ToastProvider` still contains all 100+ entries despite no visible toasts

### Expected behavior

When a toast is dismissed, its height entry should be removed from the `toastHeights` state.

### Operating System Version

macOS

### Browser

Chrome

aosmcleod:
Closes #6508

## 📝 Description

The `toastHeights` state in `ToastProvider` tracks measured heights by toast key but never removes entries when toasts are dismissed. Over long-running sessions with many toasts, this object grows unbounded — a classic memory leak.

- `handleToastHeightChange` adds entries to `toastHeights` when toasts mount
- When toasts are dismissed, their entries remain in the state forever
- After 100+ toasts, the state object contains 100+ stale entries

## 🚀 New behavior

- Added a `useEffect` that watches `visibleToasts` and prunes stale keys from `toastHeights`
- Only runs when the visible toast set changes (not on every render)
- Uses referential equality check to avoid unnecessary state updates when no stale keys exist

No

## 📝 Additional Information

The cleanup effect is intentionally conservative — it only removes keys that are no longer in `visibleToasts`. This avoids any timing issues where a toast might be measured before it appears in the visible list.


wingkwong (changes requested):
Please take a look at the typecheck issue and fix it.

aosmcleod:
@wingkwong fixed — typecheck and lint now pass.

Two issues in the original effect:
1. Under strict `noUncheckedIndexedAccess`, `prev[key]` is `number | undefined`, so assigning into `Record<string, number>` failed (the original TS error).
2. The `react-hooks/set-state-in-effect` rule also flagged the `setToastHeights` call inside `useEffect`.

Restructured: dropped the cleanup effect entirely. Pruning now happens opportunistically inside `handleToastHeightChange` — every mounted toast measures its height on appear, which gives a natural point to drop entries for toasts that are no longer in `toastQueue.visibleToasts`. This bounds `toastHeights` to the currently visible set without a separate effect.

Commit also includes a small ESLint sort-keys auto-fix on pre-existing keys in the same file (no behavioural change) since the pre-commit hook required it.

PR #6513, opened 2026-05-11
fix(toast): replace useMediaQuery with SSR-safe pattern to avoid hydration mismatch

### LINKED ISSUE #6509: [BUG] - Toast: `useMediaQuery` for action button position causes SSR hydration mismatch
### HeroUI Version

3.0.4

### Describe the bug

In `packages/react/src/components/toast/toast.tsx`, the `ToastProvider` uses `useMediaQuery` to conditionally render the action button in different positions based on viewport width:

```tsx
const isMobile = useMediaQuery("(max-width: 768px)");
// ...
// Inside getDefaultChildren:
{isMobile && actionProps?.children ? (
  <ToastActionButton ... />  // renders INSIDE ToastContent on mobile
) : null}
// ...
{!isMobile && actionProps?.children ? (
  <ToastActionButton ... />  // renders OUTSIDE ToastContent on desktop
) : null}
```

During SSR (or any server-rendered environment like Next.js App Router), `useMediaQuery` returns `false` because `window.matchMedia` is not available. On a mobile client, the first client render returns `true`. This creates a **hydration mismatch** — the server HTML has the action button outside `ToastContent`, but the client expects it inside.

### Root cause

`useMediaQuery` is not SSR-safe when used to conditionally change the **structure** of the React tree (different DOM positions). It's safe for styling/classes but not for conditional rendering that changes element hierarchy.

### Suggested fix

Option A: Always render both positions, use CSS to show/hide:
```tsx
<ToastContent>
  {!!title && <ToastTitle>{title}</ToastTitle>}
  {!!description && <ToastDescription>{description}</ToastDescription>}
  {actionProps?.children ? (
    <ToastActionButton className="sm:hidden" {...actionProps}>{actionProps.children}</ToastActionButton>
  ) : null}
</ToastContent>
{actionProps?.children ? (
  <ToastActionButton className="hidden sm:flex" {...actionProps}>{actionProps.children}</ToastActionButton>
) : null}
```

Option B: Use a `useEffect`-based approach that defaults to the server-rendered layout and only switches on the client after hydration:
```tsx
const [isMobile, setIsMobile] = useState(false); // always false on server
useEffect(() => {
  setIsMobile(window.matchMedia("(max-width: 768px)").matches);
}, []);
```

Option A is preferred as it avoids layout shift.

### Impact

- **SSR:** React hydration mismatch warning in Next.js / Remix / any SSR framework
- **Visual:** Potential layout flash on mobile as the action button moves positions after hydration
- **Severity:** Affects all SSR apps using `Toast.Provider` with action buttons on mobile viewports

_No response_

1. Use Next.js App Router with `<Toast.Provider />` in the root layout
2. Trigger a toast with `actionProps` (e.g., an "Undo" button)
3. View the page on a mobile viewport (< 768px)
4. Check the browser console for hydration mismatch warnings
5. Observe the action button briefly appearing in the wrong position before settling

### Expected behavior

Toast layout should be consistent between server and client renders, with no hydration mismatch.

### Operating System Version

macOS

### Browser

Chrome (mobile viewport emulation)

aosmcleod:
Closes #6509

## 📝 Description

`ToastProvider` uses `useMediaQuery("(max-width: 768px)")` to conditionally render the action button in different DOM positions (inside `ToastContent` on mobile, outside on desktop). During SSR, `useMediaQuery` returns `false` because `window.matchMedia` is unavailable. On mobile clients, the first render returns `true`, creating a hydration mismatch.

- `useMediaQuery` hook is used directly — returns `false` on server, potentially `true` on client
- Server renders action button outside `ToastContent`
- Mobile client expects it inside `ToastContent`
- React logs hydration mismatch warning

## 🚀 New behavior

- Replace `useMediaQuery` with a `useState(false)` + `useEffect` pattern
- Always returns `false` on server AND first client render (matching SSR output)
- Updates to correct value after hydration via `useEffect`
- Also listens for viewport changes via `MediaQueryList.addEventListener` for responsive behavior
- Removed the `useMediaQuery` import (no longer needed in this file)

No — the action button position may briefly render in desktop layout on mobile before the effect fires, but this is a single-frame flash that's preferable to a hydration error.

## 📝 Additional Information

A more comprehensive fix would render both positions and use CSS `hidden`/`sm:flex` classes to toggle visibility without changing DOM structure. That approach avoids even the single-frame flash but requires changes to the toast CSS. This PR takes the minimal approach; happy to switch to the CSS approach if preferred.

### DIFF
--- a/packages/react/src/components/toast/toast.tsx
+++ b/packages/react/src/components/toast/toast.tsx
@@ -7,7 +7,7 @@ import type {CSSProperties, ComponentPropsWithRef, ReactNode} from "react";
 import type {QueuedToast, ToastProps as ToastPrimitiveProps} from "react-aria-components/Toast";
 
 import {toastVariants} from "@heroui/styles";
-import React, {createContext, useCallback, useContext, useEffect, useMemo, useRef} from "react";
+import React, {createContext, useCallback, useContext, useEffect, useMemo, useRef, useState} from "react";
 import {Text as TextPrimitive} from "react-aria-components/Text";
 import {
   UNSTABLE_ToastContent as ToastContentPrimitive,
@@ -16,7 +16,7 @@ import {
   UNSTABLE_ToastStateContext as ToastStateContext,
 } from "react-aria-components/Toast";
 
-import {useMeasuredHeight, useMediaQuery} from "../../hooks";
+import {useMeasuredHeight} from "../../hooks";
 import {dataAttr} from "../../utils/assertion";
 import {composeSlotClassName, composeTwRenderProps} from "../../utils/compose";
 import {dom} from "../../utils/dom";
@@ -341,7 +341,23 @@ const ToastProvider = <T extends object = ToastContentValue>({
   ...rest
 }: ToastProviderProps<T>) => {
   const slots = useMemo(() => toastVariants({placement}), [placement]);
-  const isMobile = useMediaQuery("(max-width: 768px)");
+  // Use a client-only state for mobile detection to avoid SSR hydration mismatch.
+  // Default to false (desktop layout) on the server and first client render,
+  // then update after hydration via useEffect.
+  const [isMobile, setIsMobile] = useState(false);
+
+  useEffect(() => {
+    const mql = window.matchMedia("(max-width: 768px)");
+
+    setIsMobile(mql.matches);
+
+    const handler = (e: MediaQueryListEvent) => setIsMobile(e.matches);
+
+    mql.addEventListener("change", handler);
+
+    return () => mql.removeEventListener("change", handler);
+  }, []);
+
   const [toastHeights, setToastHeights] = React.useState<Record<string, number>>({});
 
   const toastQueue = useMemo((): StatelyToastQueue<T> => {
--- a/packages/react/src/components/toast/toast.tsx
+++ b/packages/react/src/components/toast/toast.tsx
@@ -7,7 +7,7 @@ import type {CSSProperties, ComponentPropsWithRef, ReactNode} from "react";
 import type {QueuedToast, ToastProps as ToastPrimitiveProps} from "react-aria-components/Toast";
 
 import {toastVariants} from "@heroui/styles";
-import React, {createContext, useCallback, useContext, useEffect, useMemo, useRef, useState} from "react";
+import React, {createContext, useCallback, useContext, useEffect, useMemo, useRef} from "react";
 import {Text as TextPrimitive} from "react-aria-components/Text";
 import {
   UNSTABLE_ToastContent as ToastContentPrimitive,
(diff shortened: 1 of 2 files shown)


wingkwong (changes requested):
In the narrow sense, it will remove the hydration warnings but it trades with a visual flash. I think CSS approach would be preferable.

wingkwong:
Please click re-request review once it is ready for review. Thanks.

aosmcleod:
@wingkwong addressed your feedback — agree the previous `useState` + `useEffect` approach traded the hydration warning for a visual flash. Switched to a CSS-only solution.

Now renders the action button in both DOM positions and uses two new modifier classes (`toast__action--mobile`, `toast__action--desktop`) to toggle visibility per breakpoint:

\`\`\`css
.toast__action--mobile {
  @apply sm:hidden;
}
.toast__action--desktop {
  @apply max-sm:hidden;
}
\`\`\`

CSS decides which copy is visible, so the server and client render identically — no hydration mismatch — and there's no post-hydration flash because there's no JS state to wait for. Aligns the breakpoint with the existing CSS (`sm:` at 640px), which the previous JS (768px) didn't match.

Ready for another look.

packages/styles/components/toast.css
```diff
@@ -156,6 +156,18 @@
...
+ * (inside .toast__content for stacked-below layout, and as a sibling of
+ * .toast__content for inline layout). CSS — not JS — picks which one is
+ * visible, so there is no SSR/CSR mismatch and no post-hydration flash. */
+.toast__action--mobile {
+  @apply sm:hidden;
+}
+
+.toast__action--desktop {
+  @apply max-sm:hidden;
+}
```
wingkwong, line 169: The previous JS used `(max-width: 768px)` while the new CSS uses `sm:` (640px). Devices in the 640–767px range (tablets in portrait, smaller Chromebooks) used to get the mobile stacked layout.

packages/react/src/components/toast/toast.tsx
```diff
@@ -395,18 +395,28 @@ const ToastProvider = <T extends object = ToastContentValue>({
...
+                {...actionProps}
+                className={composeTwRenderProps(actionProps.className, "toast__action--mobile")}
+              >
+                {actionProps.children}
+              </ToastActionButton>
             ) : null}
           </ToastContent>
-          {!isMobile && actionProps?.children ? (
-            <ToastActionButton {...actionProps}>{actionProps.children}</ToastActionButton>
+          {actionProps?.children ? (
```
wingkwong, line 407: `actionProps` is spread onto two DOM nodes. There will be some issues like two elements share one `id`, ref called for both mounts but the second one always wins etc.

You may consider a single-node CSS layout. I think that can be achieved with the existing wrapper plus flex direction + `order` so that we can just have a single `<ToastActionButton>` as a sibling of `<ToastContent>` and don't need `--mobile` / `--desktop` modifiers.

wingkwong (changes requested):

wingkwong:
Closing - Stale

PR #6566, opened 2026-05-23
fix(styles): cover scrollbar gutter in modal and alert-dialog backdrops

### LINKED ISSUE #6562: [BUG] Modal backdrop does not cover scrollbar gutter when usePreventScroll is active
### HeroUI Version

v3 (latest canary)

### Describe the bug

When a Modal opens, React Aria's `usePreventScroll` sets `scrollbar-gutter: stable` on `<html>` to prevent layout shift. The modal backdrop uses `w-full` (resolves to `width: 100%`), which on a `position: fixed` element resolves against the viewport's content area — excluding the scrollbar gutter. This leaves a visible strip at the right edge of the viewport where the backdrop doesn't cover.

### Root cause

`width: 100%` on a fixed element resolves against the containing block (the viewport minus the gutter). `width: 100vw` resolves against the full viewport including the gutter.

### Proposed fix

Replace `w-full` with explicit CSS on `.modal__backdrop`:

```css
.modal__backdrop {
  width: 100vw;
  max-width: 100%;
}
```

- `100vw` covers the full viewport including the scrollbar gutter
- `max-width: 100%` prevents overflow when the modal is rendered inside a scoped portal container (`UNSTABLE_portalContainer`) that is narrower than the viewport

This is a one-line CSS change with no API or behavioural impact.

1. Open a page with enough content to produce a scrollbar
2. Open a Modal
3. Observe the right edge of the backdrop — a thin strip of the page content is visible where the scrollbar gutter sits

### Expected behavior

The backdrop should cover the entire viewport, including the scrollbar gutter area.

### Operating System Version

macOS (all platforms affected)

### Browser

Chrome, Firefox, Safari (all affected — the gutter behaviour is standard)

aosmcleod:
Closes #6562

## 📝 Description

Fixes the modal and alert-dialog backdrops not covering the scrollbar gutter when React Aria's `usePreventScroll` is active.

When a Modal or AlertDialog opens, React Aria sets `scrollbar-gutter: stable` on `<html>` to prevent layout shift. The backdrop uses `w-full` (`width: 100%`), which on a `position: fixed` element resolves against the viewport's content area — excluding the gutter. This leaves a visible strip at the right edge of the viewport where the backdrop doesn't cover.

## 🚀 New behavior

Replace `w-full` with explicit `width: 100vw; max-width: 100%`:

- `100vw` resolves against the full viewport including the scrollbar gutter
- `max-width: 100%` prevents overflow when the backdrop is rendered inside a scoped portal container (`UNSTABLE_portalContainer`) that is narrower than the viewport

The fix is applied to both `.modal__backdrop` and `.alert-dialog__backdrop`.

No. Pure CSS fix with no API or behavioural changes.

## 📝 Additional Information

Validated by inspecting the computed styles. The `inset: 0` already on the backdrop handles positioning; the width change only affects the explicit `width` property that was previously set via `w-full`.

### DIFF
--- a/packages/styles/components/alert-dialog.css
+++ b/packages/styles/components/alert-dialog.css
@@ -45,7 +45,17 @@
 .alert-dialog__backdrop {
   @apply fixed inset-0 z-50;
   @apply flex flex-row items-center justify-center;
-  @apply h-(--visual-viewport-height) w-full;
+  @apply h-(--visual-viewport-height);
+
+  /**
+   * Use 100vw instead of w-full (100%) to cover the scrollbar gutter.
+   * React Aria's usePreventScroll sets scrollbar-gutter: stable on <html>,
+   * which shrinks the 100% containing block by the gutter width — leaving
+   * a strip at the right edge uncovered. 100vw resolves against the full
+   * viewport. max-width: 100% prevents overflow in scoped portal containers.
+   */
+  width: 100vw;
+  max-width: 100%;
 
   /* Entering animation */
   &[data-entering="true"] {
--- a/packages/styles/components/modal.css
+++ b/packages/styles/components/modal.css
@@ -44,7 +44,17 @@
 .modal__backdrop {
   @apply fixed inset-0 z-50;
   @apply flex flex-row items-center justify-center;
-  @apply h-(--visual-viewport-height) w-full;
+  @apply h-(--visual-viewport-height);
+
+  /**
+   * Use 100vw instead of w-full (100%) to cover the scrollbar gutter.
+   * React Aria's usePreventScroll sets scrollbar-gutter: stable on <html>,
+   * which shrinks the 100% containing block by the gutter width — leaving
+   * a strip at the right edge uncovered. 100vw resolves against the full
+   * viewport. max-width: 100% prevents overflow in scoped portal containers.
+   */
+  width: 100vw;
+  max-width: 100%;
 
   /* Entering animation */
   &[data-entering="true"] {


wingkwong (commented):
1. can you share two screenshots (before & after)? I dont really see the issue following your steps. Just wondering if i misunderstand something.
2. should `drawer__backdrop` also need this change?

aosmcleod:
Closing this PR — investigation revealed the visible scrollbar strip isn't a backdrop-width issue at all.

## Root cause

The strip reproduces specifically on **macOS Safari with System Settings → Appearance → "Show scroll bars: Always"** enabled. It's the OS scrollbar widget (track) painted inside the gutter that React Aria's `usePreventScroll` reserves via `scrollbar-gutter: stable` on `<html>` alongside `overflow: hidden`. In light mode the OS track is a light grey, which contrasts visibly against the dark backdrop; in dark mode it's camouflaged.

Confirmed by comparing inline `<html>` style between `react-aria.adobe.com/Modal` and a HeroUI modal while the modal is open:

| Site | `<html>` inline style |
|---|---|
| react-aria.adobe.com/Modal | `overflow: hidden;` |
| HeroUI docs / Athena | `scrollbar-gutter: stable; overflow: hidden;` |

`usePreventScroll` only adds `scrollbar-gutter: stable` when `window.innerWidth - document.documentElement.clientWidth > 0`, i.e., when `<html>` itself is the scroll container. The React Aria docs site avoids this because their page architecture puts the scroll on an inner container with `<html>` permanently at `overflow: hidden` — so the gutter-width check evaluates to 0 and the rule is never injected. HeroUI's docs (and most consumer apps using HeroUI) have `<html>` as the scroll container, so the gutter gets reserved and the OS paints its track inside it.

## Why `width: 100vw` doesn't fix it

This is a paint-layer issue, not a layout-width issue. The backdrop's geometric width has no effect on what the OS paints inside the reserved gutter. Verified with pixel comparison: `width: 100%` and `width: 100vw` render identically beside the scrollbar in Safari.

## Workarounds attempted

| Approach | Result |
|---|---|
| `width: 100vw` + `max-width: 100%` (this PR) | No visible effect |
| Tint `<html>` background to `var(--backdrop)` | Safari's native scrollbar paints opaquely on top |
| Override `scrollbar-gutter` to `auto` while overlay open | Removes strip, but causes a 15px content shift |
| `scrollbar-width: none` / `scrollbar-color: transparent` while overlay open | Ignored by Safari for native classic scrollbars on `<html>` |
| `::-webkit-scrollbar { display: none }` while overlay open | Causes layout shift via Safari's custom-scrollbar mode switch |

## The clean fix

Architectural: don't make `<html>` the scroll container. Pin `<html>` to `overflow: hidden` and move the scroll to an inner element (the pattern react-aria.adobe.com uses). That sidesteps the entire `scrollbar-gutter` injection. Out of scope for this PR — and arguably belongs upstream in React Aria (see [adobe/react-spectrum#1216](https://github.com/adobe/react-spectrum/issues/1216), open since 2020) since this is `usePreventScroll`'s behaviour.

Closing as not-the-right-fix.

PR #6582, opened 2026-05-27
fix(styles): preserve Select/Autocomplete variant background when isInvalid is set

### LINKED ISSUE #6532: [BUG] -  `Select`/`Autocomplete` ignore `variant` prop and switch to 'primary' when `isInvalid` is set
### HeroUI Version

~~2.7.11~~ 3.1.0

### Describe the bug

When `isInvalid={true}` is passed to `Select` or `Autocomplete` components, the component's variant prop is overridden and the component renders with the "primary" variant.

_No response_

1. Render a `Select` or `Autocomplete` with `variant="secondary"`
2. Pass `isInvalid={true}` to the component

### Expected behavior

The component should maintain its variant="secondary" styling when isInvalid is set.

### Screenshots or Videos

https://github.com/user-attachments/assets/5e805848-c17b-4aa2-8ecb-06d355df3839

### Operating System Version

macOS

### Browser

Chrome

mvanhorn:
## Summary

fix: preserve Select/Autocomplete variant background when isInvalid is set

Closes #6532

---
AI was used for assistance.

### DIFF
--- a/packages/styles/components/autocomplete.css
+++ b/packages/styles/components/autocomplete.css
@@ -97,12 +97,11 @@
   &:focus,
   &[data-focus="true"] {
   }
-}
 
-.autocomplete--secondary[data-invalid="true"] .autocomplete__trigger,
-.autocomplete--secondary[aria-invalid="true"] .autocomplete__trigger {
-  @apply status-invalid-field;
-  background-color: var(--autocomplete-trigger-bg);
+  &[data-invalid="true"],
+  &[aria-invalid="true"] {
+    background-color: var(--autocomplete-trigger-bg-focus);
+  }
 }
 
 .autocomplete__value {
--- a/packages/styles/components/autocomplete.css
+++ b/packages/styles/components/autocomplete.css
@@ -68,9 +68,11 @@
   /* Default styles */
 }
 
-.autocomplete--primary[data-invalid="true"] .autocomplete__trigger,
-.autocomplete--primary[aria-invalid="true"] .autocomplete__trigger {
-  background-color: var(--field-focus);
+.autocomplete--primary .autocomplete__trigger {
+  &[data-invalid="true"],
+  &[aria-invalid="true"] {
+    background-color: var(--field-focus);
+  }
 }
 
 .autocomplete--secondary .autocomplete__trigger {
--- a/packages/styles/components/autocomplete.css
+++ b/packages/styles/components/autocomplete.css
@@ -53,6 +53,7 @@
   .autocomplete[data-invalid="true"] &,
   .autocomplete[aria-invalid="true"] & {
     @apply status-invalid-field;
+    background-color: var(--field-focus);
   }
 
   /* Disabled state */
@@ -66,13 +67,6 @@
 /* Variant property definitions */
 .autocomplete--primary {
   /* Default styles */
-}
-
-.autocomplete--primary .autocomplete__trigger {
-  &[data-invalid="true"],
-  &[aria-invalid="true"] {
-    background-color: var(--field-focus);
-  }
 }
 
 .autocomplete--secondary .autocomplete__trigger {
@@ -100,8 +94,8 @@
   &[data-focus="true"] {
   }
 
-  &[data-invalid="true"],
-  &[aria-invalid="true"] {
+  .autocomplete[data-invalid="true"] &,
+  .autocomplete[aria-invalid="true"] & {
     background-color: var(--autocomplete-trigger-bg-focus);
   }
 }
--- a/packages/styles/components/autocomplete.css
+++ b/packages/styles/components/autocomplete.css
@@ -53,7 +53,6 @@
   .autocomplete[data-invalid="true"] &,
   .autocomplete[aria-invalid="true"] & {
     @apply status-invalid-field;
-    background-color: var(--field-focus);
   }
 
   /* Disabled state */
@@ -67,6 +66,11 @@
 /* Variant property definitions */
 .autocomplete--primary {
   /* Default styles */
+}
+
+.autocomplete--primary[data-invalid="true"] .autocomplete__trigger,
+.autocomplete--primary[aria-invalid="true"] .autocomplete__trigger {
+  background-color: var(--field-focus);
 }
 
 .autocomplete--secondary .autocomplete__trigger {
@@ -93,6 +97,12 @@
   &:focus,
   &[data-focus="true"] {
   }
+}
+
+.autocomplete--secondary[data-invalid="true"] .autocomplete__trigger,
+.autocomplete--secondary[aria-invalid="true"] .autocomplete__trigger {
+  @apply status-invalid-field;
+  background-color: var(--autocomplete-trigger-bg);
 }
 
 .autocomplete__value {
--- a/packages/styles/components/input-otp.css
+++ b/packages/styles/components/input-otp.css
@@ -109,10 +109,6 @@
   &[data-filled="true"] {
     background-color: var(--input-otp-slot-bg-focus);
   }
-
-  &[data-invalid="true"] {
-    background-color: var(--input-otp-slot-bg-focus);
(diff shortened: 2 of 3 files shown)


wingkwong (changes requested):
A few comments:
- Refactor to the nested &`[data-invalid]` pattern to match other css files like input, textarea etc.
- Switch secondary's invalid bg to `var(--*-trigger-bg-focus)`.
- Drop the redundant `@apply status-invalid-field` from the secondary override.
- Double check if the fix can be applied to input otp or not

mvanhorn:
Done in 062383d28:
- Nested the secondary invalid state inside `.{select,autocomplete}--secondary .…__trigger` via `&[data-invalid]` / `&[aria-invalid]`, matching the input.css pattern.
- Switched the invalid background to `var(--*-trigger-bg-focus)`.
- Dropped the redundant `@apply status-invalid-field` (the base `[data-invalid]` rule already applies it).
- input-otp had the same secondary-variant cascade bug, so I applied the analogous fix to `.input-otp--secondary .input-otp__slot`.

wingkwong (commented):
1. why do we need to add `background-color: var(--field-focus);` for primary one?
2. why `background-color: var(--field-focus);` got removed?

mvanhorn:
Thanks @wingkwong! Refactored in d97363389 — the primary invalid override now uses the nested `&[data-invalid="true"], &[aria-invalid="true"]` block inside `.select--primary .select__trigger` (and the autocomplete equivalent), matching input.css/textarea.css and the secondary block.

On your two questions:

1. **Why `var(--field-focus)` for primary?** To match input.css/textarea.css, where the primary (default) variant's invalid state resolves to `var(--field-focus)`. Keeping it preserves the primary trigger's background under `isInvalid` instead of letting the generic invalid styling override it.

2. **Why was it removed from the shared invalid block?** It previously lived in the shared `.select[data-invalid] &` rule, which forced `--field-focus` on *every* variant — including secondary, which should use `--select-trigger-bg-focus`. Moving it out and setting it per-variant makes the invalid background variant-aware (primary → `--field-focus`, secondary → `--*-trigger-bg-focus`), consistent with input/textarea.

Secondary already uses `var(--select-trigger-bg-focus)` / `var(--autocomplete-trigger-bg-focus)` with no leftover `@apply status-invalid-field`, and I added the equivalent invalid state to input-otp. Happy to adjust if you'd prefer a different approach.

packages/styles/components/autocomplete.css
```diff
@@ -69,6 +68,13 @@
   /* Default styles */
 }
 
+.autocomplete--primary .autocomplete__trigger {
+  &[data-invalid="true"],
+  &[aria-invalid="true"] {
+    background-color: var(--field-focus);
+  }
+}
+
```
wingkwong: remove this block. `data-invalid` and `aria-invalid` are not in trigger.

packages/styles/components/select.css
```diff
@@ -80,6 +79,13 @@
   /* Default styles */
 }
 
+.select--primary .select__trigger {
+  &[data-invalid="true"],
+  &[aria-invalid="true"] {
+    background-color: var(--field-focus);
+  }
+}
+
```
wingkwong: remove this block. `data-invalid` and `aria-invalid` are not in trigger.

packages/styles/components/select.css
```diff
@@ -103,6 +109,11 @@
   &:focus,
   &[data-focus="true"] {
   }
+
+  &[data-invalid="true"],
+  &[aria-invalid="true"] {
+    background-color: var(--select-trigger-bg-focus);
+  }
```
wingkwong:
```css
  .select[data-invalid="true"] &,
  .select[aria-invalid="true"] & {
    background-color: var(--select-trigger-bg-focus);
  }
```

packages/styles/components/select.css
```diff
@@ -64,7 +64,6 @@
   .select[data-invalid="true"] &,
   .select[aria-invalid="true"] & {
     @apply status-invalid-field;
-    background-color: var(--field-focus);
```
wingkwong, line 67: bring it back

packages/styles/components/autocomplete.css
```diff
@@ -93,6 +99,11 @@
   &:focus,
   &[data-focus="true"] {
   }
+
+  &[data-invalid="true"],
+  &[aria-invalid="true"] {
+    background-color: var(--autocomplete-trigger-bg-focus);
+  }
```
wingkwong:
```css
  .autocomplete[data-invalid="true"] &,
  .autocomplete[aria-invalid="true"] & {
    background-color: var(--autocomplete-trigger-bg-focus);
  }
```

packages/styles/components/autocomplete.css
```diff
@@ -53,7 +53,6 @@
   .autocomplete[data-invalid="true"] &,
   .autocomplete[aria-invalid="true"] & {
     @apply status-invalid-field;
-    background-color: var(--field-focus);
```
wingkwong, line 56: bring it back

wingkwong (changes requested):

packages/styles/components/input-otp.css
```diff
@@ -109,6 +109,10 @@
   &[data-filled="true"] {
     background-color: var(--input-otp-slot-bg-focus);
   }
+
+  &[data-invalid="true"] {
+    background-color: var(--input-otp-slot-bg-focus);
+  }
```
wingkwong: remove this block. the issue doesn't affect input otp.

wingkwong (changes requested):

mvanhorn:
Thanks for the detailed review, @wingkwong. Pushed the changes:

- Refactored both `--primary` and `--secondary` to the wrapper-scoped pattern. Since `data-invalid` / `aria-invalid` sit on `.select` / `.autocomplete` (not on `__trigger`), the old `&[data-invalid]` selectors never actually matched the trigger. They are now `.select[data-invalid="true"] &, .select[aria-invalid="true"] &` (and the autocomplete equivalents), so they apply to the trigger as intended.
- Secondary invalid background now uses `var(--*-trigger-bg-focus)`, replacing the trigger-scoped `danger-bg` block and the nested focus-within override with a single wrapper-scoped rule.
- Reverted the input-otp change. You're right, the issue doesn't affect input otp.

On the two questions:

1. (why add `var(--field-focus)` for primary) It keeps the invalid state on the variant's focus background instead of the danger background, matching how input/textarea preserve their background when invalid. It's now correctly scoped to the wrapper. If you'd rather primary not carry an invalid override at all, happy to drop it.
2. (why `var(--field-focus)` got removed from `:focus-within`) That was an unintended removal on my end. Restored.

`@heroui/styles` builds clean and the new selectors compile as expected.

mvanhorn:
Correction to my note above: the push landed just now in `350bdba` (the previous comment got ahead of the actual commit, apologies). Changes per your review:

- **select / autocomplete**: invalid background is now wrapper-scoped (`.select[data-invalid="true"] &` / `.autocomplete[data-invalid="true"] &`) since `data-invalid` / `aria-invalid` live on the container, not the trigger. The standalone `.--primary .__trigger` invalid blocks are removed (they never matched), and the secondary override now uses `var(--*-trigger-bg-focus)` as you suggested.
- Restored `background-color: var(--field-focus)` on the base trigger invalid state (the removal was unintentional).
- Reverted input-otp entirely; you're right that the issue doesn't affect it.

On your two questions: the `var(--field-focus)` on primary keeps the invalid state on the focus background instead of danger (matching input/textarea), now correctly scoped to the wrapper; and the `:focus-within` removal was the unintentional one, now restored.

wingkwong (approved):

mvanhorn:
Thanks for reviewing and merging, @wingkwong. Select and Autocomplete keeping their variant background is a nice polish win.

PR #6590, opened 2026-05-31
fix(button): prevent translate-induced crash on pending changes

### LINKED ISSUE #6480: [BUG] - Button loading state causes insertBefore NotFoundError when DOM is modified by browser translation (Google Translate)
### HeroUI Version

3.0.3

### Describe the bug

When using HeroUI Button with the `loading` state, the application crashes with a React DOM error:
NotFoundError: Failed to execute 'insertBefore' on 'Node':

The node before which the new node is to be inserted is not a child of this node.

This issue occurs in real-world scenarios when the page is affected by external DOM mutations such as browser translation (e.g. Google Translate extension or Chrome page translation feature).

The crash happens during React reconciliation when the Button enters or exits the loading state while the DOM structure is being modified externally.

This seems to affect the internal rendering logic of the Button component when `isLoading` is used.

https://heroui.com/docs/react/components/button#loading-state

1. Open the official HeroUI Button documentation:
   https://heroui.com/docs/react/components/button#loading-state

2. Scroll to the "Loading State" example.

3. Enable browser translation (e.g. Google Translate / Chrome "Translate page").

4. Translate the page to another language (e.g. Arabic).

5. Click the Button that uses `isLoading` or loading state.

6. Observe the application crash in the browser console.

### Expected behavior

The Button component should handle loading state transitions safely and remain stable even in environments where the DOM is being modified externally (e.g. browser translation extensions).

No runtime crash or React reconciliation error should occur during or after the loading state change.

### Screenshots or Videos

https://github.com/user-attachments/assets/160f7fe4-8e63-4a3d-b159-0646e4743311

### Operating System Version

windows

### Browser

Chrome

KumJungMin:
Closes #6480

## 📝 Description

Fixes a React DOM crash that can occur when `Button` content is mutated by browser translation tools such as Chrome Google Translate and the `isPending` state changes afterward.

<br/>

When Chrome Google Translate translates text inside a `Button`, it can remove the original React-rendered text node and replace it with new DOM nodes such as `<font>` elements.

React still keeps references to the original text nodes. If `isPending` changes after translation, React may try to insert or remove button content relative to a text node that is no longer a child of the button.

This can crash the page with errors such as:

```text
NotFoundError: Failed to execute 'insertBefore' on 'Node'
The node before which the new node is to be inserted is not a child of this node.
```

<br/>

## 🚀 New behavior
When isPending is explicitly provided as a boolean, Button now renders its content inside a stable display: contents wrapper.

The content subtree is remounted when the pending state changes, which avoids reconciling directly against text nodes that may have been replaced by browser translation.

This follows the same general mitigation approach used by MUI Button for Google Translate related React crashes.

<br/>

No.

Buttons that do not use isPending continue to render as before. The wrapper is only added when isPending is explicitly passed as true or false.

<br/> 

### before
- Before this change, enabling Chrome Google Translate and then changing the `Button` `isPending` state could crash the page.

https://github.com/user-attachments/assets/aa3d765a-a41a-4800-9482-9d654283c7ad

### after
- After this change, the same flow no longer crashes.

https://github.com/user-attachments/assets/ff18753d-c3ee-4f10-a028-00ae15ebc5fb

### DIFF
--- a/packages/react/src/components/button/button.tsx
+++ b/packages/react/src/components/button/button.tsx
@@ -1,10 +1,10 @@
 "use client";
 
 import type {ButtonVariants} from "@heroui/styles";
-import type {ComponentPropsWithRef} from "react";
+import type {ComponentPropsWithRef, ReactNode} from "react";
 
 import {buttonVariants} from "@heroui/styles";
-import {useContext} from "react";
+import {Children, Fragment, cloneElement, isValidElement, useContext} from "react";
 import {Button as ButtonPrimitive} from "react-aria-components/Button";
 
 import {composeTwRenderProps} from "../../utils";
@@ -16,6 +16,23 @@ import {BUTTON_GROUP_CHILD, ButtonGroupContext} from "../button-group";
 interface ButtonRootProps extends ComponentPropsWithRef<typeof ButtonPrimitive>, ButtonVariants {
   [BUTTON_GROUP_CHILD]?: boolean;
 }
+
+const wrapTextNodes = (node: ReactNode): ReactNode =>
+  Children.map(node, (child) => {
+    if (typeof child === "string") {
+      return child.trim() === "" ? child : <span>{child}</span>;
+    }
+
+    if (typeof child === "number") {
+      return <span>{child}</span>;
+    }
+
+    if (isValidElement<{children?: ReactNode}>(child) && child.type === Fragment) {
+      return cloneElement(child, undefined, wrapTextNodes(child.props.children));
+    }
+
+    return child;
+  });
 
 const ButtonRoot = ({
   children,
@@ -70,7 +87,7 @@ const ButtonRoot = ({
           return renderedChildren;
         }
 
-        return <span>{renderedChildren}</span>;
+        return wrapTextNodes(renderedChildren);
       }}
     </ButtonPrimitive>
   );
--- a/packages/react/src/components/button/button.tsx
+++ b/packages/react/src/components/button/button.tsx
@@ -1,7 +1,7 @@
 "use client";
 
 import type {ButtonVariants} from "@heroui/styles";
-import type {ComponentPropsWithRef} from "react";
+import type {CSSProperties, ComponentPropsWithRef} from "react";
 
 import {buttonVariants} from "@heroui/styles";
 import {useContext} from "react";
@@ -17,12 +17,15 @@ interface ButtonRootProps extends ComponentPropsWithRef<typeof ButtonPrimitive>,
   [BUTTON_GROUP_CHILD]?: boolean;
 }
 
+const contentWrapperStyle: CSSProperties = {display: "contents"};
+
 const ButtonRoot = ({
   children,
   className,
   fullWidth,
   isDisabled,
   isIconOnly,
+  isPending,
   size,
   slot,
   style,
@@ -43,6 +46,8 @@ const ButtonRoot = ({
   const finalFullWidth =
     fullWidth ?? (shouldUseContext ? buttonGroupContext?.fullWidth : undefined);
 
+  const hasPendingState = typeof isPending === "boolean";
+
   const styles = buttonVariants({
     fullWidth: finalFullWidth,
     isIconOnly,
@@ -55,11 +60,28 @@ const ButtonRoot = ({
       className={composeTwRenderProps(className, styles)}
       data-slot="button"
       isDisabled={finalIsDisabled}
+      isPending={isPending}
       slot={slot}
       style={style}
       {...rest}
     >
-      {(renderProps) => (typeof children === "function" ? children(renderProps) : children)}
+      {(renderProps) => {
+        const renderedChildren = typeof children === "function" ? children(renderProps) : children;
+
+        if (!hasPendingState) {
+          return renderedChildren;
+        }
+
+        return (
+          <span
+            key={renderProps.isPending ? "pending" : "idle"}
+            data-slot="button-content"
+            style={contentWrapperStyle}
+          >
+            {renderedChildren}
+          </span>
+        );
+      }}
     </ButtonPrimitive>
   );
 };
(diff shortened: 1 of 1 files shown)


packages/react/src/components/button/button.tsx
```diff
@@ -17,12 +17,15 @@ interface ButtonRootProps extends ComponentPropsWithRef<typeof ButtonPrimitive>,
   [BUTTON_GROUP_CHILD]?: boolean;
 }
 
+const contentWrapperStyle: CSSProperties = {display: "contents"};
```
wingkwong: Adding this can drop the wrapper from the accessibility tree. Not good for a focusable Button.

packages/react/src/components/button/button.tsx
```diff
@@ -55,11 +60,28 @@ const ButtonRoot = ({
...
+      {(renderProps) => {
+        const renderedChildren = typeof children === "function" ? children(renderProps) : children;
+
+        if (!hasPendingState) {
+          return renderedChildren;
+        }
+
+        return (
+          <span
+            key={renderProps.isPending ? "pending" : "idle"}
```
wingkwong: I think this would remount the entire content subtree on every pending toggle like focus, animation etc.

packages/react/src/components/button/button.tsx
```diff
@@ -55,11 +60,28 @@ const ButtonRoot = ({
...
+            data-slot="button-content"
```
wingkwong: new slot should be documented

packages/react/src/components/button/button.tsx
```diff
@@ -55,11 +60,28 @@ const ButtonRoot = ({
...
+            style={contentWrapperStyle}
```
wingkwong: inline styles are against our design system pattern unless they are dynamically calculated. Styles should be defined in css.

wingkwong (commented):
> This follows the same general mitigation approach used by MUI Button for Google Translate related React crashes.

Mind to share the PR number from MUI repo for reference?

wingkwong (commented):
I'm also thinking whether we should fix this issue. In general, users can `span` on text nodes or add `translate="no"` depending on their use case. Adding a `span` internally would cover the first case but it would be unnecessary for the latter one. Other libraries like ant design, chakra-ui also don't support it as it is a react-related issue.

@KumJungMin wdyt?

KumJungMin:
> > This follows the same general mitigation approach used by MUI Button for Google Translate related React crashes.
> 
> Mind to share the PR number from MUI repo for reference?

---

Sure 🙂  — the MUI PR I was referring to is mui/material-ui#35198:

It was created to address this issue: https://github.com/mui/material-ui/issues/27853

One small nuance: that PR targeted MUI Lab’s `LoadingButton`. The fix wrapped `children` in a `<span>` to avoid the Chrome/Google Translate + React DOM crash.

MUI’s current core `Button` appears to keep a similar mitigation around the loading state as well. In particular, `loading` defaults to `null`, and when it is explicitly set to `true` or `false`, the component renders a stable loading wrapper. This helps avoid DOM-shape changes that can trigger the Google Translate crash.

References:
- Original MUI issue: https://github.com/mui/material-ui/issues/27853
- MUI fix PR: https://github.com/mui/material-ui/pull/35198
- Current MUI Button type comment: https://github.com/mui/material-ui/blob/master/packages/mui-material/src/Button/Button.d.ts
- Current MUI Button implementation: https://github.com/mui/material-ui/blob/master/packages/mui-material/src/Button/Button.js#L693

스크린샷 2026-06-15 오후 8 18 33

 _image1: Current MUI Button implementation: loading props_

KumJungMin:
> I'm also thinking whether we should fix this issue. In general, users can `span` on text nodes or add `translate="no"` depending on their use case. Adding a `span` internally would cover the first case but it would be unnecessary for the latter one. Other libraries like ant design, chakra-ui also don't support it as it is a react-related issue.
> 
> @KumJungMin wdyt?

----


Thanks for the feedback 😄

As you mentioned, I understand that the root cause of this issue is external DOM mutation caused by Google Translate / browser translation tools. I also agree that users can work around this depending on their use case by manually wrapping text nodes with `<span>`, or by using `translate="no"` for content that should not be translated.

At first, I thought we might be able to provide a scoped mitigation inside Button, since `isPending` is a first-class Button state.

When `isPending` changes, the Button content is reconciled again. If the text nodes have already been wrapped or replaced by the browser translator, that reconciliation can lead to an `insertBefore` / `NotFoundError` crash.

However, after revisiting the implementation based on your feedback, I realized that this is not as small or isolated as I initially thought.

Wrapping the entire content subtree can affect the Button layout. Button relies on its direct children for flex/gap alignment, so if an icon and label are wrapped together, they become a single flex item and the layout can break.

I also tried narrowing the change to only wrap primitive text nodes that are direct children of the Button.

Updated:  [7e5abcd](https://github.com/heroui-inc/heroui/pull/6590/commits/7e5abcdf6d1df6cf56a859be04aa25de3c60ee02)

```html
<button>
  <svg />
  <span>Upload File</span>
</button>
```

This avoids wrapping the whole subtree, does not require `display: contents` or inline styles, and keeps the icon and label as separate flex items.

But this still feels more like a partial mitigation than a complete solution. It does not normalize arbitrary render-prop content, and a more complete MUI-like solution would likely require Button to own a stable internal loading/label structure. That seems closer to a larger API/design change than a small bug fix.

So after looking into the implementation more, I’m leaning toward your point that documenting the workaround may be the more appropriate direction rather than handling this partially inside the component.

For example, we could recommend that users manually wrap text nodes when needed, or use `translate="no"` when the Button label should not be translated.

```tsx
<Button isPending={isLoading} translate="no">
  Upload File
</Button>
```

This would keep the component behavior unchanged and avoid introducing a partial internal mitigation that may not cover all cases.

If this direction sounds good to you, I can update this PR to document the workaround instead of changing the Button implementation 🙂

PR #6593, opened 2026-06-01
fix(select): ensure selected item is visible when reopening popover

### LINKED ISSUE #6489: [BUG] - Select Popover is scrolled to wrong position, leaving selected item out of view
### HeroUI Version

3.0.3

### Describe the bug

When a Select component opens its popover and value is already selected, the popover attempts to scroll to the selected item. However, the scroll position lands somewhere near the middle of the list rather than at the selected item, which remains out of view.

_No response_

1. Render a `<Select>` with a long list of options (enough to require scrolling).
2. Set a default value that is not near the top of the list (e.g. an item in the bottom third).
3. Open the Select popover by clicking/focusing the trigger.
4. Observe the scroll position inside the popover list.

### Expected behavior

The popover scrolls so that the selected item is visible - ideally centered or near the top of the visible area.

### Screenshots or Videos

https://github.com/user-attachments/assets/a4672d38-75a1-4ee9-9d24-56b3aa6ddb87

### Operating System Version

macOS

### Browser

Chrome

aman44444:
Closes #6489 

## 📝 Description

Fixes an issue where the Select popover did not scroll to the currently selected item when reopened. As a result, the selected option could remain outside the visible area of the list.

After selecting an option and reopening the Select popover, the list remains at its default scroll position. The selected item may be located outside the visible viewport, requiring the user to manually scroll to find it.

## 🚀 New behavior

When the Select popover is opened, it automatically scrolls to the currently selected item, ensuring that the selected option is visible to the user.

No.

## 📝 Additional Information

Verified that reopening the Select popover automatically scrolls the list to the selected item.

Demo

https://github.com/user-attachments/assets/effb6d82-fe6e-4cc9-b6d9-31fff7195020

### DIFF
--- a/packages/react/src/components/select/select.tsx
+++ b/packages/react/src/components/select/select.tsx
@@ -7,11 +7,7 @@ import type {SelectVariants} from "@heroui/styles";
 import type {ComponentPropsWithRef} from "react";
 
 import {selectVariants} from "@heroui/styles";
-<<<<<<< HEAD
-import React, {createContext, use} from "react";
-=======
-import React, {createContext, useContext, useEffect, useRef} from "react";
->>>>>>> 7e28d2b55 (fix(select): correct popover scroll position when opening)
+import React, {createContext, use, useEffect, useRef} from "react";
 import {Button as ButtonPrimitive} from "react-aria-components/Button";
 import {Popover as PopoverPrimitive} from "react-aria-components/Popover";
 import {
@@ -166,11 +162,8 @@ const SelectPopover = ({
   placement = "bottom",
   ...props
 }: SelectPopoverProps) => {
-<<<<<<< HEAD
   const {slots} = use(SelectContext);
-=======
-  const {slots} = useContext(SelectContext);
-  const state = useContext(SelectStateContext);
+  const state = use(SelectStateContext);
   const popoverRef = useRef<HTMLElement>(null);
 
   useEffect(() => {
@@ -193,7 +186,6 @@ const SelectPopover = ({
       window.cancelAnimationFrame(rafId);
     };
   }, [state?.isOpen]);
->>>>>>> 7e28d2b55 (fix(select): correct popover scroll position when opening)
 
   return (
     <SurfaceContext
--- a/packages/react/src/components/select/select.tsx
+++ b/packages/react/src/components/select/select.tsx
@@ -134,8 +134,8 @@ const SelectIndicator = <E extends keyof React.JSX.IntrinsicElements = "svg">({
       {
         ...(props as any),
         className: composeSlotClassName(slots?.indicator, className),
-        "data-open": dataAttr(state?.isOpen),
         "data-slot": "select-indicator",
+        "data-open": dataAttr(state?.isOpen),
       },
     );
   }
@@ -176,34 +176,21 @@ const SelectPopover = ({
   useEffect(() => {
     if (!state?.isOpen || !popoverRef.current) return;
 
-    const scrollSelectedItem = () => {
-      const popover = popoverRef.current;
+    let rafId = 0;
 
-      if (!popover) return;
+    const timerId = window.setTimeout(() => {
+      rafId = window.requestAnimationFrame(() => {
+        const selectedItem =
+          popoverRef.current?.querySelector<HTMLElement>('[aria-selected="true"]');
 
-      const selectedItem = popover.querySelector<HTMLElement>('[aria-selected="true"]');
-
-      if (!selectedItem) return;
-
-      const popoverHeight = popover.clientHeight;
-      const itemTop = selectedItem.offsetTop;
-      const itemHeight = selectedItem.offsetHeight;
-      const scrollTop = Math.max(
-        0,
-        itemTop - Math.floor(popoverHeight / 2) + Math.floor(itemHeight / 2),
-      );
-
-      popover.scrollTo({
-        behavior: "auto",
-        top: scrollTop,
+        selectedItem?.scrollIntoView({block: "nearest"});
       });
-    };
-
-    const timer = window.setTimeout(() => {
-      window.requestAnimationFrame(scrollSelectedItem);
     }, 0);
 
-    return () => window.clearTimeout(timer);
+    return () => {
+      window.clearTimeout(timerId);
+      window.cancelAnimationFrame(rafId);
+    };
   }, [state?.isOpen]);
 >>>>>>> 7e28d2b55 (fix(select): correct popover scroll position when opening)
 
--- a/packages/react/src/components/select/select.tsx
+++ b/packages/react/src/components/select/select.tsx
@@ -7,6 +7,7 @@ import type {SelectVariants} from "@heroui/styles";
 import type {ComponentPropsWithRef} from "react";
 
 import {selectVariants} from "@heroui/styles";
+import {mergeRefs} from "@react-aria/utils";
 import React, {createContext, use, useEffect, useRef} from "react";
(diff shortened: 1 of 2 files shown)


aman44444:
Hi @wingkwong could you please review this when you get a chance? Thanks!

packages/react/src/components/select/select.tsx
```diff
@@ -160,6 +160,41 @@ const SelectPopover = ({
...
+        behavior: "auto",
+        top: scrollTop,
+      });
+    };
+
+    const timer = window.setTimeout(() => {
+      window.requestAnimationFrame(scrollSelectedItem);
+    }, 0);
+
+    return () => window.clearTimeout(timer);
```
wingkwong: if the popover closes or the component unmounts after the setTimeout callback has fired but before the raf callback runs, the cleanup only clears the timer, the raf still executes against a popover that's now mid-exit-animation (or detached).

packages/react/src/components/select/select.tsx
```diff
@@ -127,8 +127,8 @@ const SelectIndicator = <E extends keyof React.JSX.IntrinsicElements = "svg">({
       {
         ...(props as any),
         className: composeSlotClassName(slots?.indicator, className),
-        "data-slot": "select-indicator",
         "data-open": dataAttr(state?.isOpen),
+        "data-slot": "select-indicator",
```
wingkwong, line 134: unrelated change regarding to this pr

packages/react/src/components/select/select.tsx
```diff
@@ -160,6 +160,41 @@ const SelectPopover = ({
   ...props
 }: SelectPopoverProps) => {
   const {slots} = useContext(SelectContext);
+  const state = useContext(SelectStateContext);
+  const popoverRef = useRef<HTMLElement>(null);
+
+  useEffect(() => {
+    if (!state?.isOpen || !popoverRef.current) return;
+
+    const scrollSelectedItem = () => {
```
wingkwong: this looks a bit tedious. can we simply use native approach like scrollIntoView? If u do manual calc here, it seems not cater `scroll-py-1` in `select__popover`.

wingkwong (changes requested):
please also test a bit with multi select and research a bit if react-aria built-in scroll has attempted this or not.

aman44444:
Hi @wingkwong, thanks for the detailed review! Here's what I've updated:
1. Fixed RAF cleanup
Initialized `rafId = 0 ` so cancelAnimationFrame is always called with a valid value, even if the component unmounts before setTimeout fires. Both clearTimeout and cancelAnimationFrame now run in cleanup.
2. Switched to native scrollIntoView
Replaced the manual offsetTop calculation with `selectedItem?.scrollIntoView({ block: "nearest" })`. This is simpler and automatically respects the `scroll-py-1 `padding defined in `select__popover`.
3. Removed unrelated data-slot change
Reverted `SelectIndicator ` back to the original — that duplicate attribute was an accidental change.

aman44444:
Hi @wingkwong, here's an update on the two points:

React Aria research:

React Aria does have built-in scroll logic in `useSelectableCollection`. It fires for both keyboard and mouse through the `didAutoFocusRef` path. But it uses a custom scroll utility that sets `scrollTop` directly on  `scrollRef.current`, which points to the inner `ListBox` ul. In HeroUI, the actual scrollable element is the outer `PopoverPrimitive` with `overflow-y-auto`, not the ul, so React Aria's scroll targets the wrong element and silently does nothing. This is also confirmed by a React Aria collaborator in [adobe/react-spectrum#2447](https://github.com/adobe/react-spectrum/discussions/2447).

Our fix targets `popoverRef.current` directly (the correct scrollable element) and uses native `scrollIntoView({ block: "nearest" })` which also respects `scroll-py-1 automatically`.

Multi-select question:

Tested the React Aria demo with mouse by selecting items 20, 2, 5 in that click order. On reopen it scrolled to item 20. Looking at the React Aria source, `SelectionManager` tracks `lastSelectedKey` as the last selected item by position in the list (DOM order, not click order). When the trigger is clicked, focus enters the listbox from above so React Aria uses `lastSelectedKey` which is item 20 (the highest positioned selected item in the list).

Our current implementation uses `querySelector('[aria-selected="true"]') ` which scrolls to the first selected item in DOM order (item 2 in this case). To match React Aria we would use `querySelectorAll('[aria-selected="true"]')` and take the last result.

Should we match React Aria's behaviour and scroll to the last selected item by position in the list, or keep scrolling to the first?

wingkwong:
yes should match RA's behaviour

aman44444:
Hi @wingkwong, updated to match React Aria's behaviour, now scrolls to the last selected item by position in the list for multi-select. Tested with both single and multi-select.
Demo attached above.

wingkwong:
please resolve the conflicts

aman44444:
Hi @wingkwong, resolved the conflicts.

wingkwong:
@aman44444 clearly not. please self-verify once.

aman44444:
Hi @wingkwong, I've resolved the merge conflicts. Could you please take another look when you have a chance? Thanks!

packages/react/src/components/select/select.tsx
```diff
@@ -172,6 +195,7 @@ const SelectPopover = ({
     >
       <PopoverPrimitive
         {...props}
+        ref={popoverRef}
```
wingkwong: this overwrites whatever the user passed. should use mergeRefs

packages/react/src/components/select/select.tsx
```diff
@@ -163,6 +163,29 @@ const SelectPopover = ({
...
+
+  useEffect(() => {
+    if (!state?.isOpen || !popoverRef.current) return;
+
+    let rafId = 0;
+
+    const timerId = window.setTimeout(() => {
+      rafId = window.requestAnimationFrame(() => {
+        const selectedItems =
+          popoverRef.current?.querySelectorAll<HTMLElement>('[aria-selected="true"]');
```
wingkwong: this matches any descendant of the popover. need better scope to list box options only.

wingkwong (changes requested):
I'm wondering if the approach is correct or not. RA's Select passes the ListBox element as scrollRef and should already scroll the selected option into view on open. we put scroll on the popover instead so the scrollRef points at a non-scrollable element.

Also we've introduced tests for components. please add the corresponding tests.

aman44444:
Hi @wingkwong, after deep research I identified the root cause and implemented a fix.

The issue had two parts. First, the `ListBox` wasn't positioned as the actual scroll container in HeroUI's layout structure. The `PopoverPrimitive` was handling scroll, but React Aria's `scrollRef` points to the `ListBox`. Second, the `ListBox`'s size constraints come from the popover's `max-height` which is set dynamically by RAC's positioning engine after the initial render. When React Aria tries to scroll on mount, the geometry isn't finalized yet, so the scroll position doesn't align correctly with the final layout.

The fix addresses both issues. CSS changes make the `ListBox` the scroll container by adding overflow-y-auto and inheriting the popover's max-height. A `useEffect` in `SelectPopover` waits for the layout to fully settle when the popover opens, then scrolls the selected item into view using native `scrollIntoView` which automatically respects `scroll-py-1`.

Why this works: The `useEffect` depends on `state?.isOpen` so it only fires when the popover is open. By that time, RAC has finalized all positioning and sizing. This timing ensures the scroll happens against the correct geometry. For multi-select, it scrolls to the last selected item by position using `querySelectorAll`.

Tested with long lists, pre-selected items at various positions, multi-select scenarios, and keyboard navigation. All scenarios work correctly.

Ready for review. I'll add test cases once you confirm the approach is correct.

PR #6604, opened 2026-06-03
fix(styles): restore Modal scroll="inside" scrolling (fix #6603)

### LINKED ISSUE #6603: [BUG] - Modal scroll="inside" no longer scrolls in @heroui/styles 3.1.0 (overflow-clip regression)
### Describe the bug

`Modal` with `scroll="inside"` stopped scrolling its body in `@heroui/styles@3.1.0`. With enough content the body no longer scrolls and the header/footer are pushed off-screen, so content becomes unreachable. It worked in `3.0.5`.

### Root cause

This is a CSS regression in `@heroui/styles`. Between `3.0.5` and `3.1.0`, `.modal__dialog--scroll-inside` changed:

```diff
 .modal__dialog--scroll-inside {
-  @apply overflow-hidden;   /* 3.0.5 */
+  @apply overflow-clip;     /* 3.1.0 */
 }
```

Per the CSS Flexbox spec, a flex item's **automatic minimum size** is `0` only when the item is a *scroll container*. `overflow: hidden` makes an element a scroll container; `overflow: clip` does **not**. So with `clip`, the dialog's `min-height` reverts from `0` to content-based: the dialog (a flex item inside the `h-(--visual-viewport-height)` `.modal__container`) refuses to shrink to the viewport, and the body (`.modal__body { flex-1 min-h-0 }` + `.modal__body--scroll-inside { overflow-y-auto }`) never gets a bounded height to scroll within.

Knock-on effect: because inside-scroll is broken, people switch to `scroll="outside"`, whose `.modal__container--scroll-outside { pointer-events-auto }` then swallows backdrop clicks and defeats `isDismissable` (click-outside-to-close).

### Steps to reproduce

1. Render a Modal with `<Modal.Container scroll="inside">`.
2. Put tall content in `Modal.Body` (e.g. ~30 paragraphs).
3. Open it → the body does not scroll; the header/footer overflow the viewport.

```tsx
<Modal>
  <Button>Open</Button>
  <Modal.Backdrop>
    <Modal.Container scroll="inside">
      <Modal.Dialog className="sm:max-w-[360px]">
        <Modal.Header><Modal.Heading>Scroll inside</Modal.Heading></Modal.Header>
        <Modal.Body>
          {Array.from({ length: 30 }).map((_, i) => (
            <p key={i} className="mb-3">Paragraph {i + 1} — lorem ipsum…</p>
          ))}
        </Modal.Body>
        <Modal.Footer><Button slot="close">Close</Button></Modal.Footer>
      </Modal.Dialog>
    </Modal.Container>
  </Modal.Backdrop>
</Modal>
```

The docs' own "Scroll Comparison" example relies on `scroll="inside"` scrolling with 30 paragraphs and no height class, so it reproduces there too.

### Expected behavior

The `Modal.Body` scrolls within the dialog; the header/footer stay pinned and visible, as in `3.0.5`.

### Suggested fix

Restore the flex shrink that `clip` removed, in `packages/styles/components/modal.css`:

```diff
 .modal__dialog--scroll-inside {
-  @apply overflow-clip;
+  @apply overflow-clip min-h-0;
 }
```

`min-h-0` re-supplies the `min-height: 0` that `overflow: hidden` provided implicitly (scroll containers get an automatic min-size of 0) — keeping the intentional `clip` while letting the dialog shrink. Alternatively, revert to `@apply overflow-hidden;`. PR incoming.

### Version

`@heroui/styles` 3.1.0 (regression from 3.0.5; still present on the `v3` branch HEAD).

### Environment

All browsers / OS — pure CSS layout behavior.

FZ1010:
Closes #6603

## Summary

`Modal` with `scroll="inside"` stopped scrolling its body in `@heroui/styles@3.1.0`. `.modal__dialog--scroll-inside` was changed from `overflow: hidden` to `overflow: clip`.

Unlike `hidden`, **`overflow: clip` is not a scroll container**. Per the CSS Flexbox spec, a flex item only gets an automatic `min-height` of `0` when it is a scroll container — otherwise it stays content-based. So with `clip`, the dialog (a flex item inside the `h-(--visual-viewport-height)` `.modal__container`) no longer shrinks to the viewport, and the body (`.modal__body { flex-1 min-h-0 }` + `.modal__body--scroll-inside { overflow-y-auto }`) never gets a bounded height to scroll within — the header/footer get pushed off-screen and content becomes unreachable.

## Fix

```diff
 .modal__dialog--scroll-inside {
-  @apply overflow-clip;
+  @apply overflow-clip min-h-0;
 }
```

`min-h-0` re-supplies the `min-height: 0` that `overflow: hidden` provided implicitly, **keeping the intentional `overflow: clip`** while restoring the dialog's ability to shrink. (A straight revert to `overflow: hidden` also fixes it — happy to switch if maintainers prefer.)

## Why it matters

Besides broken inside-scroll, the regression nudges users toward `scroll="outside"`, whose `.modal__container--scroll-outside { pointer-events-auto }` then swallows backdrop clicks and breaks `isDismissable` (click-outside-to-close) — so one CSS change cascades into two visible problems.

## Verification

- Matches the behavior `@heroui/styles@3.0.5` shipped (it used `overflow: hidden`, i.e. the same implicit `min-height: 0`, and scrolled correctly).
- The docs' own "Scroll Comparison" example (30 paragraphs, `scroll="inside"`, no height class) scrolls again with this change.

## Notes

- CSS-only, one line; no change to `scroll="outside"` behavior.
- Opened as a **draft** for maintainer input on `min-h-0` vs reverting to `overflow-hidden`.

### DIFF
--- a/packages/styles/components/modal.css
+++ b/packages/styles/components/modal.css
@@ -196,7 +196,7 @@
  * Modifier: dialog--scroll-inside
  */
 .modal__dialog--scroll-inside {
-  @apply overflow-clip;
+  @apply overflow-clip min-h-0;
 }
 
 /**


wingkwong:
covered in https://github.com/heroui-inc/heroui/pull/6597.

PR #6681, opened 2026-06-29
fix(styles): render separators in-flow so they no longer collapse to a stray line

### LINKED ISSUE #6680: [Bug] ButtonGroup/ToggleButtonGroup Separator renders as a stray vertical line (absolute positioning, no anchor)
### Describe the bug

`<ButtonGroup.Separator>` and `<ToggleButtonGroup.Separator>` do not render a divider between the buttons. Instead they collapse into a single stray ~1px vertical line on the left edge of the nearest positioned ancestor (e.g. a `position: relative` Card wrapping the group).

### Root cause

The separator renders as a `<span data-slot="button-group-separator">` that is a **sibling** of the buttons inside the group, styled (horizontal orientation):

```css
.button-group__separator {
  position: absolute;
  left: -1px;
  top: 25%;
  width: 1px;
  height: 50%;
}
```

`position: absolute; left: -1px` only lands on the segment boundary if the separator's offset parent is the adjacent button (or a per-segment `position: relative` wrapper). Neither exists:

- The group base (`.button-group` / `.toggle-button-group`) is `inline-flex … gap-0` with **no `position: relative`**, so it is not a positioning context.
- The buttons are `position: relative`, but the separator is their **sibling**, not a descendant, so the buttons' positioning context does not apply to it.

With no anchor, each separator's offset parent bubbles up to the nearest positioned ancestor. When the group sits inside a `position: relative` container (a Card, a panel, …), every separator collapses onto that container's left edge at `left: -1px`, stacking into one stray vertical line.

### Reproduction

```tsx
<div style={{ position: "relative" }}>
  <ToggleButtonGroup defaultSelectedKeys={["light"]}>
    <ToggleButton id="light">Light</ToggleButton>
    <ToggleButtonGroup.Separator />
    <ToggleButton id="dark">Dark</ToggleButton>
    <ToggleButtonGroup.Separator />
    <ToggleButton id="system">System</ToggleButton>
  </ToggleButtonGroup>
</div>
```

Expected: thin dividers between the buttons.
Actual: no dividers between buttons; a single ~1px vertical line at the left edge of the `position: relative` wrapper.

### Affected source (v3 branch HEAD, identical to published @heroui/styles 3.2.1)

- `packages/styles/components/button-group.css` — `.button-group` is `inline-flex … gap-0` (no `position: relative`); `.button-group__separator` is `position: absolute; left: -1px`.
- `packages/styles/components/toggle-button-group.css` — same shape.
- `packages/react/src/components/button-group/button-group.tsx` (and `toggle-button-group.tsx`) — Root renders children directly; the separator is a standalone `<span>` with no per-button wrapper.

### Suggested fix

The absolute separator needs a per-segment positioning anchor:

- **A.** Anchor the separator to its adjacent button (render it inside the button — which is already `position: relative` — or wrap each segment in a `position: relative` wrapper). Then `left: -1px` lands on the boundary.
- **B.** Drop absolute positioning and draw the divider in-flow (a border / `divide-*` between adjacent buttons), removing the offset-parent dependency entirely.

Note: adding `position: relative` to the group alone does **not** fix it — all separators then stack on the group's own left edge.

### Version

- `@heroui/react` 3.2.1, `@heroui/styles` 3.2.1
- v3 branch HEAD is byte-identical to the published 3.2.1 CSS.

Fldicoahkiin:
Closes #6680

## 📝 Description

`<ButtonGroup.Separator>` / `<ToggleButtonGroup.Separator>` render as a single stray ~1px vertical line on the nearest positioned ancestor instead of as dividers between the buttons. This renders the separator in-flow so it lands in the gap between the buttons.

**Before / after** — `ButtonGroup`/`ToggleButtonGroup` with separators inside a `position: relative` Card:
图片

## ⛳️ Current behavior (the bug)

The separator is a `<span data-slot="*-separator">` that is a **sibling** of the buttons inside the group, styled (horizontal orientation):

```css
.button-group__separator {
  position: absolute;
  left: -1px;
  top: 25%;
  width: 1px;
  height: 50%;
}
```

`position: absolute; left: -1px` only lands on the segment boundary if the separator's offset parent is the adjacent button (or a per-segment `position: relative` wrapper). Neither exists:

- The group base (`.button-group` / `.toggle-button-group`) is `inline-flex … gap-0` with **no `position: relative`**, so it is not a positioning context.
- The buttons are `position: relative`, but the separator is their **sibling**, not a descendant, so the buttons' positioning context does not apply to it.

With no anchor, each separator's offset parent bubbles up to the nearest positioned ancestor. When the group sits inside a `position: relative` container (a Card, popover, toolbar — usually positioned), every separator collapses onto that container's left edge at `left: -1px`, stacking into one stray vertical line, and no dividers appear between the buttons.

### Reproduction

Any `ButtonGroup` / `ToggleButtonGroup` with `.Separator` placed inside a positioned ancestor:

```tsx
import { Button, ButtonGroup } from "@heroui/react";

export default function Repro() {
  return (
    // Any `position: relative` ancestor triggers it (Cards/popovers/toolbars usually are).
    <div style={{ position: "relative", padding: 24, border: "1px solid #ccc" }}>
      <ButtonGroup>
        <Button>Light</Button>
        <ButtonGroup.Separator />
        <Button>Dark</Button>
        <ButtonGroup.Separator />
        <Button>System</Button>
      </ButtonGroup>
    </div>
  );
}
```

- **Expected:** thin dividers between `Light │ Dark │ System`.
- **Actual:** no dividers between the buttons; a single ~1px vertical line on the **left edge of the `position: relative` wrapper**.

`ToggleButtonGroup` behaves the same; in vertical orientation the separators collapse onto the wrapper's **top** edge (via `top: -1px`).

## 🚀 New behavior (the fix)

Render the separator as an **in-flow flex item**. It already sits in the gap between the buttons in the DOM, so dropping `position: absolute` (and the `left/top: -1px` offsets) lets flexbox place it there, cross-centered by the group's `items-center`. `width`/`height` keep the 1px × 50% divider (50% × 1px for vertical); `flex: none` keeps it 1px when full-width buttons grow.

Verified across horizontal / vertical / full-width / RTL — the old physical `left: -1px` was also RTL-incorrect, whereas the in-flow version has no physical offset. Applies to both `ButtonGroup` and `ToggleButtonGroup`; `ToggleButtonGroup` detached mode still hides separators.

> Note: adding `position: relative` to the group alone would **not** fix it — all separators (each at `left: -1px`) would then stack on the group's own left edge.

No. Visual-only fix; the `ButtonGroup.Separator` / `ToggleButtonGroup.Separator` API is unchanged. The only layout change is that an attached-mode separator now occupies 1px of in-flow space (the divider itself) instead of being a zero-space overlay.

## 📝 Additional Information

Verified on `@heroui/styles` 3.2.1 / `v3` branch HEAD.

### DIFF
--- a/packages/styles/components/button-group.css
+++ b/packages/styles/components/button-group.css
@@ -73,7 +73,7 @@
 
 .button-group__separator {
   @apply rounded-sm bg-current opacity-15;
-  position: absolute;
+  flex: none;
   pointer-events: none;
 
   /**
@@ -85,16 +85,12 @@
 
   /* Horizontal orientation - vertical divider */
   .button-group--horizontal & {
-    left: -1px;
-    top: 25%;
     width: 1px;
     height: 50%;
   }
 
   /* Vertical orientation - horizontal divider */
   .button-group--vertical & {
-    left: 25%;
-    top: -1px;
     width: 50%;
     height: 1px;
   }
--- a/packages/styles/components/button-group.css
+++ b/packages/styles/components/button-group.css
@@ -85,12 +85,14 @@
 
   /* Horizontal orientation - vertical divider */
   .button-group--horizontal & {
+    margin-inline-start: -1px;
     width: 1px;
     height: 50%;
   }
 
   /* Vertical orientation - horizontal divider */
   .button-group--vertical & {
+    margin-block-start: -1px;
     width: 50%;
     height: 1px;
   }
--- a/packages/styles/components/toggle-button-group.css
+++ b/packages/styles/components/toggle-button-group.css
@@ -90,7 +90,7 @@
 
 .toggle-button-group__separator {
   @apply rounded-sm bg-current opacity-15;
-  position: absolute;
+  flex: none;
   pointer-events: none;
 
   /**
@@ -102,16 +102,12 @@
 
   /* Horizontal orientation - vertical divider */
   .toggle-button-group--horizontal & {
-    left: -1px;
-    top: 25%;
     width: 1px;
     height: 50%;
   }
 
   /* Vertical orientation - horizontal divider */
   .toggle-button-group--vertical & {
-    left: 25%;
-    top: -1px;
     width: 50%;
     height: 1px;
   }
--- a/packages/styles/components/toggle-button-group.css
+++ b/packages/styles/components/toggle-button-group.css
@@ -102,12 +102,14 @@
 
   /* Horizontal orientation - vertical divider */
   .toggle-button-group--horizontal & {
+    margin-inline-start: -1px;
     width: 1px;
     height: 50%;
   }
 
   /* Vertical orientation - horizontal divider */
   .toggle-button-group--vertical & {
+    margin-block-start: -1px;
     width: 50%;
     height: 1px;
   }


wingkwong (changes requested):
This PR would break the existing visual layout a bit. You may check the examples from storybook.

PR Build: https://heroui-sb-git-fork-fldicoahkiin-fix-button-group-86eb3b-heroui.vercel.app/?path=/story/components-buttons-buttongroup--examples

image

v3: https://storybook-v3.heroui.com/?path=/story/components-buttons-buttongroup--examples

image

Fldicoahkiin:
Thanks for the review @wingkwong! You're right — rendering the separator in-flow added 1px between the buttons and shifted the layout.

Addressed in f95c144: a negative `margin-inline-start` / `margin-block-start` of `-1px` cancels that 1px, so the buttons stay flush and the divider overlays the seam — matching the current v3 visual, while still fixing the stray vertical line that the absolute positioning caused.

Mind taking another look? The storybook preview should rebuild with the latest commit.

wingkwong (changes requested):
The difference is quite obvious. Please run the storybook to verify once before requesting the review.

Fldicoahkiin:
Closing this — after running the storybook as you suggested, I found my diagnosis was wrong.

The documented placement puts `<ButtonGroup.Separator />` **inside** the following button (per the stories and docs demos), where the button's `position: relative` anchors `left: -1px` exactly on the seam — the current CSS is correct for that usage. My report came from placing separators **between** buttons as siblings, which has no anchor and collapses onto the nearest positioned ancestor; that was a misuse on my side, now fixed in our app.

The in-flow approach here breaks the intended usage (the separator picks up the button's internal flex gap), so this PR is the wrong direction. Sorry for the noise, and thanks for the patient review.

One small takeaway: sibling placement type-checks and renders without any warning but silently breaks — might be worth a note in the docs.

PR #6692, opened 2026-07-05
fix(autocomplete): add accessible name to popover dialog and search field demo

### LINKED ISSUE #6691: [BUG] - Autocomplete basic example triggers a11y warnings on open and search
### HeroUI Version

3.2.1

### Describe the bug

**Component:** Autocomplete  
**Docs:** https://heroui.com/en/docs/react/components/autocomplete#usage

**Steps:** Use the basic usage example → click the field → type in search.

**Warnings:**
- On open - `If you do not provide a visible label, you must specify an aria-label or aria-labelledby attribute for accessibility`
-  On typing - `If a Dialog does not contain a <Heading slot="title">, it must have an aria-label or aria-labelledby attribute for accessibility`

**Expected:** No a11y warnings with the documented example.

_No response_

Docs: https://heroui.com/en/docs/react/components/autocomplete#usage

Use the basic usage example → click the field → type in search.

### Expected behavior

No a11y warnings with the documented example.

### Screenshots or Videos

_No response_

### Operating System Version

macOS

### Browser

Chrome

momomuchu:
Closes #6691

## 📝 Description

The popover's internal dialog (added in #6627 to prevent a stray focus ring on touch) had no accessible name, and the docs' basic usage example never labeled the search field either. Both triggered react-aria a11y warnings on open and while typing.

The dialog now reuses the same aria-labelledby Select computes for the popover, falling back to a static label for the collection-build render pass where that context isn't available yet.

Opening the basic Autocomplete demo and typing in the search field logs two react-aria dev warnings (missing aria-label/aria-labelledby on the popover dialog and on the search field).

## 🚀 New behavior

No a11y warnings. Dialog inherits the field's label via aria-labelledby (or falls back to "Options"), search field demos have an explicit aria-label.

No

## 📝 Additional Information

packages/react has no test runner wired up yet (empty vitest.config.ts, no test script), so verification here was manual: built the package, ran it against the docs demo, confirmed both warnings are gone in the console before/after. pnpm typecheck and pnpm lint pass.

### DIFF
--- a/packages/react/src/components/autocomplete/autocomplete.tsx
+++ b/packages/react/src/components/autocomplete/autocomplete.tsx
@@ -230,14 +230,10 @@ const AutocompletePopover = ({
 
   useResizeObserver({ref: triggerRef, onResize});
 
-  // `Select` (react-aria-components) computes an `aria-labelledby` for its popover content
-  // and provides it through `PopoverContext`, which `PopoverPrimitive` below consumes
-  // automatically. The nested `DialogPrimitive` we render for touch focus containment is a
-  // separate accessible element that `Select` doesn't know about, so it does not inherit that
-  // label automatically. Forward it explicitly so the dialog is labeled by the same field label
-  // as the trigger. `PopoverContext` is unavailable while collection-building performs its
-  // context-free render pass, so fall back to a static label to guarantee the dialog always has
-  // an accessible name, even then.
+  // The nested DialogPrimitive below makes react-aria-components' Popover drop role="dialog" on
+  // its own node (isDialog = shouldBeDialog && !ref.current.querySelector('[role=dialog]')), so
+  // PopoverContext's aria-labelledby lands on a roleless container. DialogPrimitive reads
+  // DialogContext, not PopoverContext, so this forwards the label explicitly (see #6691).
   const popoverContext = useContext(PopoverContext);
   const contextAriaLabelledby =
     popoverContext && typeof popoverContext === "object" && "aria-labelledby" in popoverContext
@@ -245,6 +241,8 @@ const AutocompletePopover = ({
       : undefined;
 
   const dialogAriaLabelledby = props["aria-labelledby"] ?? contextAriaLabelledby;
+  // PopoverContext is unavailable during the context-free collection-building render pass, so
+  // fall back to a static label to keep the dialog's accessible name in that case too.
   const dialogAriaLabel = props["aria-label"] ?? (dialogAriaLabelledby ? undefined : "Options");
 
   return (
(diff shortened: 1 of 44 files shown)


packages/react/src/components/autocomplete/autocomplete.tsx
```diff
@@ -230,6 +230,23 @@ const AutocompletePopover = ({
...
+  // as the trigger. `PopoverContext` is unavailable while collection-building performs its
+  // context-free render pass, so fall back to a static label to guarantee the dialog always has
+  // an accessible name, even then.
+  const popoverContext = useContext(PopoverContext);
+  const contextAriaLabelledby =
+    popoverContext && typeof popoverContext === "object" && "aria-labelledby" in popoverContext
+      ? popoverContext["aria-labelledby"]
+      : undefined;
+
+  const dialogAriaLabelledby = props["aria-labelledby"] ?? contextAriaLabelledby;
```
wingkwong: I think you don't need `contextAriaLabelledby` at all. The popover overlay around it is already labeled via the aria-labelledby that Select pushes through PopoverContext and PopoverPrimitive consumes automatically.

wingkwong (changes requested):
Please check all the cases
image

momomuchu:
Thanks for pushing back, it made me re-check properly. You're right that Select pushes `aria-labelledby` through `PopoverContext`, but that lands on the outer Popover div, and that div only takes `role="dialog"` when Popover's `!ref.querySelector('[role=dialog]')` check passes.

Since #6627 (wrap popover content in a dialog to prevent the stray focus ring), `AutocompletePopover` always renders a nested `<Dialog>`, so that querySelector always finds a `role=dialog` child, `isDialog` is always false on the outer div, and the inner Dialog reads `DialogContext` rather than `PopoverContext`, so the Select label never reaches the element that actually carries `role=dialog`. So it isn't only the nested-Dialog case I mentioned before, it's every Autocomplete instance after that change.

Before/after on the element that actually has `role=dialog`:

- without this PR: `aria-labelledby=null`, `aria-label=null` (react-aria also logs "A dialog must have a title for accessibility...")
- with this PR: `aria-labelledby` resolves to the label

I can add that as a test in the PR if you'd like it captured.

wingkwong:
Okay. I see you point. I'm thinking if we should use `useSlottedContext` for popoverContext and also need to check the precedence (i.e. aria-labelledby wins over aria-label per ARIA). Also for my first comment, there are other cases in `apps/docs/src/demos`, not just `default.tsx`. Please update them (also in storybook).

momomuchu:
Went with `useSlottedContext` for the `PopoverContext` read. It's how react-aria's own Popover reads that context internally (via `useContextProps`), and it matches `separator`/`button-group`/`toggle-button-group` here which already use it. Select provides `PopoverContext` as a direct value at this call site, so behavior is unchanged, just more idiomatic.

On precedence: the dialog only gets `aria-label` when there's no `aria-labelledby` to forward, so the two are never set together and `aria-labelledby` wins whenever it's present.

Updated the `aria-label` on the search field across the other autocomplete demos (en and cn) and the storybook file too, same pattern as `default.tsx`.

PR #6694, opened 2026-07-05
fix(autocomplete): label popover dialog and search field for a11y

### LINKED ISSUE #6691: [BUG] - Autocomplete basic example triggers a11y warnings on open and search

momomuchu:
Closes #6691

## Problem

The basic Autocomplete example emits two react-aria a11y warnings:

- On open: `If a Dialog does not contain a <Heading slot="title">, it must have an aria-label or aria-labelledby attribute for accessibility.`
- On typing: `If you do not provide a visible label, you must specify an aria-label or aria-labelledby attribute for accessibility`

Two separate causes:

1. `AutocompletePopover` wraps its content in a `Dialog` (added in #6627 to avoid a stray focus ring on touch) that has no accessible name. `Select` computes an `aria-labelledby` for its popover content and passes it through `PopoverContext`, but that labels the `Popover` element, not the extra `Dialog` nested inside it.
2. The basic usage demo never labels its `SearchField`, so the search input has no accessible name.

## Fix

- `AutocompletePopover` reads the `aria-labelledby` from `PopoverContext` (the same one `Select` uses for the popover) and forwards it to the `Dialog`. During the collection-build render pass that context isn't available yet, so it falls back to a static `aria-label` to guarantee the dialog always has a name.
- Added `aria-label` to the `SearchField` in the en and cn `default` demos.

## Notes

react-aria-components' own ComboBox doesn't nest a `Dialog` inside its popover, which is why `Select` doesn't label one automatically. Forwarding the existing label keeps the dialog announced with the same name as the field.

### DIFF
--- a/apps/docs/src/demos/cn/autocomplete/default.tsx
+++ b/apps/docs/src/demos/cn/autocomplete/default.tsx
@@ -73,7 +73,7 @@ export default function Default() {
       </Autocomplete.Trigger>
       <Autocomplete.Popover>
         <Autocomplete.Filter filter={contains}>
-          <SearchField autoFocus name="search" variant="secondary">
+          <SearchField autoFocus aria-label="搜索" name="search" variant="secondary">
             <SearchField.Group>
               <SearchField.SearchIcon />
               <SearchField.Input placeholder="搜索…" />
--- a/apps/docs/src/demos/en/autocomplete/default.tsx
+++ b/apps/docs/src/demos/en/autocomplete/default.tsx
@@ -73,7 +73,7 @@ export default function Default() {
       </Autocomplete.Trigger>
       <Autocomplete.Popover>
         <Autocomplete.Filter filter={contains}>
-          <SearchField autoFocus name="search" variant="secondary">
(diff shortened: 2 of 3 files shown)


momomuchu:
Duplicate of #6692 (same branch/commit) - closing this one, please review #6692 instead. Sorry for the noise.

Issue #6186, opened 2026-01-31
[BUG] - ScrollShadow state is not changed to "top" when scrolled to the bottom


yeeeeeeeeenya:
### HeroUI Version

3.0.0-beta5

### Describe the bug

<img width="1248" height="726" alt="Image" src="https://github.com/user-attachments/assets/a930e737-9c9f-4255-975c-0f7134db510d" />
just check examples on https://v3.heroui.com/docs/react/components/scroll-shadow
none of the vertical ones works as intended, but horizontal ones work well.
or is this some kind of "intentional design"?

### Steps to Reproduce the Bug or Issue

Check official examples on https://v3.heroui.com/docs/react/components/scroll-shadow

### Expected behavior

the state is set to "top" when scrolled to the bottom

### Screenshots or Videos

<img width="1248" height="726" alt="Image" src="https://github.com/user-attachments/assets/a930e737-9c9f-4255-975c-0f7134db510d" />

### Operating System Version

Linux

### Browser

Chrome

wingkwong:
looks fine on my side

<img width="775" height="635" alt="Image" src="https://github.com/user-attachments/assets/c8331241-5b2e-4a73-8e53-3532b307aa3d" />

PR #6710, opened 2026-07-13
fix(scroll-shadow): tolerate fractional scroll position in end detection

### LINKED ISSUE #6709: [BUG] - ScrollShadow end detection fails on HiDPI displays (fractional scrollLeft)
### HeroUI Version

3.2.2

### Describe the bug

`ScrollShadow`'s end-of-scroll detection uses an exact integer comparison in `packages/react/src/components/scroll-shadow/use-scroll-shadow.ts` (~line 40):

```ts
const hasScrollAfter = scrollStart + clientSize + offset < scrollSize;
```

On HiDPI displays (e.g. Retina, `devicePixelRatio` 2), `scrollLeft` / `scrollTop` are subpixel-precise and return fractional values, while `scrollWidth` / `clientWidth` are rounded integers. At the physical end of scroll the comparison can therefore stay `true` forever.

Measured example on a Retina display (DPR 2), horizontal orientation, scrolled fully to the end:

- `scrollLeft = 532.5`, `clientWidth = 300`, `scrollWidth = 833`
- `532.5 + 300 + 0 < 833` → `832.5 < 833` → `true`

So `hasScrollAfter` never becomes `false`: the trailing shadow never disappears, and any component built on this hook shows the same symptom — e.g. the Tabs overflow chevron stays visible even when the tab list is scrolled all the way to the end.

On DPR 1 displays the bug does not reproduce (we measured `scrollSize - scrollStart - clientSize === 0` exactly).

This is a well-documented pitfall. MDN explicitly warns that exact equality "will not work all the time because `scrollTop` can contain decimals" and recommends comparing with a 1px tolerance instead:
https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollHeight#determine_if_an_element_has_been_totally_scrolled

Proposed fix — apply the MDN-recommended 1px tolerance to the "after" side:

```diff
- const hasScrollAfter = scrollStart + clientSize + offset < scrollSize;
+ const hasScrollAfter = scrollStart + clientSize + offset < scrollSize - 1;
```

The symmetric start check (`const hasScrollBefore = scrollStart > offset;`) already tolerates small fractional values naturally, so only the end check needs the tolerance.

We have been shipping exactly this one-line change as a pnpm patch in a production app on Retina hardware and it resolves both the trailing shadow and the Tabs overflow chevron symptom, with no regression on DPR 1.

Possibly the same root cause as #6186 (vertical ScrollShadow state never reaching `"top"` at the bottom), which was closed as not reproducible — reproduction depends on the reporter's display DPR.

_No response_

1. Use a physical HiDPI display (e.g. MacBook Retina, DPR 2). Note: DevTools/CDP viewport emulation snaps scroll offsets to integers, so the bug does **not** reproduce under emulation — a real HiDPI screen is required.
2. Render a horizontal `ScrollShadow` (or a `Tabs` list that overflows, which uses the same hook for its overflow chevron).
3. Scroll fully to the end.
4. Observe `el.scrollLeft` is fractional (e.g. `532.5`) while `scrollWidth` / `clientWidth` are integers; `data-end-scroll` / the trailing shadow (and the Tabs chevron) remain visible.

### Expected behavior

When the container is scrolled fully to the end, the trailing shadow (and the Tabs overflow chevron) should disappear, on any `devicePixelRatio`.

### Screenshots or Videos

_No response_

### Operating System Version

macOS

### Browser

Chrome

grayashh:
Closes #6709

## 📝 Description

`useScrollShadow` decides whether content remains after the current scroll position with an exact comparison:

```ts
const hasScrollAfter = scrollStart + clientSize + offset < scrollSize;
```

On HiDPI displays (Retina, DPR 2), `scrollLeft`/`scrollTop` are subpixel-precise (fractional), while `scrollWidth`/`clientWidth` are rounded integers. At the physical end of scroll the comparison stays `true` (measured: `532.5 + 300 < 833`), so `hasScrollAfter` never turns `false`.

MDN documents this exact pitfall — equality checks "will not work all the time because `scrollTop` can contain decimals" — and recommends a 1px tolerance: https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollHeight#determine_if_an_element_has_been_totally_scrolled

This PR applies that 1px tolerance to the end check. The symmetric start check (`scrollStart > offset`) already tolerates fractional values naturally, so only the "after" side changes.

On HiDPI displays, the trailing shadow of `ScrollShadow` — and features built on this hook, like the Tabs overflow chevron — never disappear when scrolled fully to the end. DPR 1 displays are unaffected (gap is exactly 0 there).

https://github.com/user-attachments/assets/dbf65af6-d151-41de-b868-e788e2302f54

## 🚀 New behavior

End-of-scroll is detected within a 1px tolerance (MDN-recommended pattern), so the trailing shadow and the Tabs overflow chevron correctly disappear at the end of scroll on any `devicePixelRatio`.

https://github.com/user-attachments/assets/1651c638-46e5-4f43-81d8-acfb66c29429

No. Worst case, the shadow hides 1px before the true end — imperceptible, and the same tolerance MDN recommends for all scroll-end detection.

## 📝 Additional Information

- Validated in production: we ship exactly this one-line change as a pnpm patch (`@heroui/react` 3.2.2) in a production app, verified on Retina hardware (bug reproduces) and DPR 1 (no regression).
- Reproduction requires a physical HiDPI display — DevTools/CDP viewport emulation snaps scroll offsets to integers and hides the bug.
- No unit tests exist for `scroll-shadow` (only stories), and a JSDOM test can't exercise fractional layout-driven scroll positions meaningfully, so no test was added; happy to add one if you have a preferred approach.
- Possibly also fixes the root cause behind #6186 (vertical case, closed as not reproducible — reproduction depends on the reporter's display DPR).

### DIFF
--- a/packages/react/src/components/scroll-shadow/use-scroll-shadow.ts
+++ b/packages/react/src/components/scroll-shadow/use-scroll-shadow.ts
@@ -37,7 +37,7 @@ export const useScrollShadow = (props: UseScrollShadowProps) => {
     const clientSize = isVertical ? el.clientHeight : el.clientWidth;
 
     const hasScrollBefore = scrollStart > offset;
-    const hasScrollAfter = scrollStart + clientSize + offset < scrollSize;
+    const hasScrollAfter = scrollStart + clientSize + offset < scrollSize - 1;
 
     // Skip DOM updates if state hasn't changed
     const prevState = prevStateRef.current;


wingkwong (commented):
can you share a before & after video?

grayashh:
@wingkwong Added before & after videos to the description (under Current / New behavior). Recorded in our production app on a MacBook Retina display (DPR 2) — real hardware rather than DevTools emulation, since viewport emulation snaps scroll offsets to integers and masks the bug.

One more data point worth documenting: in our app the bug reproduced **only with Korean tab labels** — English labels on the same display were fine. That looked like a text/i18n issue at first, but it turns out to be the same root cause, and it explains why this class of bug reproduces so inconsistently (likely including #6186).

The trigger is the fractional part of the content's true width. `scrollWidth` is integer-rounded, while the real maximum scroll offset snaps to device pixels (0.5px steps at DPR 2). With an integer `clientWidth`, the strict comparison gets permanently stuck exactly when `frac(contentWidth) ∈ [0.5, 0.75)`: `scrollWidth` rounds up, the max scroll offset snaps down to x.5, and a 0.5px gap remains at full scroll.

I measured our actual six tab labels per locale in isolation, varying only `font-size` (Chrome 149, macOS, DPR 2, `clientWidth` 360; each cell = true content width via `getBoundingClientRect` → `scrollWidth`):

| labels | 13px | 14px | 15px |
|---|---|---|---|
| en (our app) | 737.83 → 738 · ok | 777.00 → 777 · ok | 815.26 → 815 · ok |
| ko (our app) | 420.58 → 421 · **stuck** | 441.52 → 442 · **stuck** | 462.40 → 462 · ok |
| ja (our app) | 527.46 → 527 · ok | 556.96 → 557 · ok | 586.46 → 586 · ok |
| en (alt label set) | 543.60 → 544 · **stuck** | 570.24 → 570 · ok | 596.28 → 596 · ok |
| zh (sample) | 435.80 → 436 · ok | 458.27 → 458 · ok | 480.72 → 481 · **stuck** |

("stuck" = `scrollLeft + clientWidth < scrollWidth` stays `true` at full scroll end, so the trailing shadow / Tabs chevron never clears.)

So it's not language-specific: Korean, English, and Chinese label sets all hit it, and the *same* Korean strings stop reproducing at 15px. Individual label widths are fractional in every locale — what matters is whether the total lands in the stuck band (~25% of arbitrary text widths at DPR 2). In our app the English set happened to sum to exactly 777.0 at 14px while the Korean set landed on x.52, which is why it masqueraded as a Korean-only bug. Reproduction therefore depends on the exact labels/font/size *and* the display DPR — all stuck cells above report `false` at scroll end with this PR's 1px tolerance.

<details><summary>Standalone repro page used for these measurements (open on a HiDPI display)</summary>

```html
<!doctype html><meta charset="utf-8">
<style>.s{width:360px;overflow-x:auto;margin:4px 0}.t{display:flex;width:max-content}.t span{padding:8px 12px;font-weight:500;white-space:nowrap}</style>
<div id="cases"></div><pre id="out"></pre>
<script>
const SETS = {
  "en-app": ["Basic info","Participant types","Meeting rules","Meeting locations","Notifications","Registration & publishing"],
  "ko-app": ["기본 정보","참가자 유형","미팅 규칙","미팅 장소","알림","참가·공개"],
  "ja-app": ["基本情報","参加者タイプ","ミーティングルール","ミーティング場所","通知","参加・公開"],
  "en-alt": ["Overview","Members","Analytics","Billing","Integrations","Advanced settings"],
  "zh-sample": ["基本信息","参与者类型","会议规则","会议地点","通知","报名与发布"],
};
const rows = [];
for (const [set, labels] of Object.entries(SETS)) for (const size of [13, 14, 15]) {
  const s = Object.assign(document.createElement("div"), { className: "s" });
  const t = Object.assign(document.createElement("div"), { className: "t" });
  t.style.fontSize = size + "px";
  for (const l of labels) t.appendChild(Object.assign(document.createElement("span"), { textContent: l }));
  s.appendChild(t); document.getElementById("cases").appendChild(s);
  rows.push({ set, size, s, t });
}
document.fonts.ready.then(() => {
  document.getElementById("out").textContent = "dpr " + devicePixelRatio + "\n" + rows.map(({ set, size, s, t }) => {
    s.scrollLeft = 1e9;
    const stuck = s.scrollLeft + s.clientWidth < s.scrollWidth;
    return `${set} ${size}px content=${t.getBoundingClientRect().width} scrollWidth=${s.scrollWidth} maxScrollLeft=${s.scrollLeft} -> ${stuck ? "STUCK" : "ok"}`;
  }).join("\n");
});
</script>
```

</details>

wingkwong (approved):

PR #6730, opened 2026-07-25
fix(tooltip): preserve ButtonGroup context for tooltip-wrapped buttons

### LINKED ISSUE #6516: [BUG] - ButtonGroup variant/size props don't propagate to Button when child is wrapped in Tooltip
### HeroUI Version

`@heroui/react@3.0.4`

### Describe the bug

`ButtonGroup` propagates `variant`, `size`, `isDisabled`, and `fullWidth` to its child Buttons via two mechanisms in tandem:

1. A `ButtonGroupContext` provider.
2. A `__button_group_child: true` marker prop stamped onto **direct** children via `Children.map` + `cloneElement` (`packages/components/button-group/src/button-group.tsx`).

`Button` then gates context consumption behind that marker:

```js
// packages/components/button/src/button.tsx
const shouldUseContext = isButtonGroupChild === true;
const finalVariant = variant ?? (shouldUseContext ? buttonGroupContext?.variant : undefined);
```

When a Button is wrapped in any single-element passthrough — most commonly `<Tooltip>`, but also things like `<Popover.Trigger>` — the wrapper becomes the direct child of ButtonGroup and absorbs the marker (Tooltip silently drops unknown props). The inner Button never receives the marker, `shouldUseContext` stays `false`, and `variant` / `size` / `isDisabled` / `fullWidth` fall back to component defaults instead of inheriting from the group.

The sibling pair behaves differently: `ToggleButtonGroup` exports `TOGGLE_BUTTON_GROUP_CHILD` but **never references it** (dead code in `toggle-button-group.tsx`), and `ToggleButton` reads `ToggleButtonGroupContext` unconditionally with no marker gate. So the exact same Tooltip wrap works for ToggleButton.

|  | Stamps marker on direct children? | Reads context conditionally? |
|---|---|---|
| `ButtonGroup` + `Button` | yes | gated on `__button_group_child === true` |
| `ToggleButtonGroup` + `ToggleButton` | declares marker constant, never uses it | reads context unconditionally |

Reproduces locally on the v3 starter; happy to put together a Stackblitz if helpful.

```tsx
import { Button, ButtonGroup, Tooltip, ToggleButton, ToggleButtonGroup } from '@heroui/react';

// Works — Button inherits "secondary" variant from ButtonGroup
<ButtonGroup variant="secondary">
  <Button>A</Button>
  <Button>B</Button>
</ButtonGroup>

// Broken — Buttons render with their own default variant, not "secondary"
<ButtonGroup variant="secondary">
  <Tooltip content="Tip A"><Button>A</Button></Tooltip>
  <Tooltip content="Tip B"><Button>B</Button></Tooltip>
</ButtonGroup>

// Works — ToggleButton reads its group's context unconditionally
<ToggleButtonGroup>
  <Tooltip content="Tip A"><ToggleButton>A</ToggleButton></Tooltip>
  <Tooltip content="Tip B"><ToggleButton>B</ToggleButton></Tooltip>
</ToggleButtonGroup>
```

1. Render a `ButtonGroup` with `variant="secondary"` (or any non-default value).
2. Wrap each child Button in a `<Tooltip>`.
3. Observe that the Buttons render with the default variant instead of `"secondary"`.
4. Swap `Button` / `ButtonGroup` for `ToggleButton` / `ToggleButtonGroup` — propagation works.

### Expected behavior

Wrapping a Button in `<Tooltip>` (or any other transparent single-element wrapper) shouldn't sever ButtonGroup → Button propagation. The natural fixes are either:

- **Have `Button` consume `ButtonGroupContext` unconditionally** — mirrors `ToggleButton`'s existing behavior and removes the asymmetry. The risk is that this resurfaces #2142 (Buttons rendered through Modals/Portals from inside a ButtonGroup would inherit group styling, since React context crosses portal boundaries).
- **Have `ButtonGroup` recurse the marker through known transparent wrappers** (`Tooltip`, `Popover.Trigger`, etc.) when cloning children — keeps #2142 fixed but requires a wrapper allowlist or a heuristic ("element with a single React element child").
- **Have transparent wrappers (`Tooltip` etc.) forward unknown props to their first child** — standard "transparent forwarding" pattern; handles arbitrary marker props without ButtonGroup needing to know about wrappers.

### Screenshots or Videos

_n/a_

### Operating System Version

macOS

### Browser

Chrome

wanxiankai:
## Summary
- Add generic internal child-prop forwarding for transparent trigger wrappers.
- Preserve ButtonGroup direct-child semantics while allowing group props through Tooltip, Dropdown, Popover, Modal, Drawer, and AlertDialog.
- Keep wrappers independent of Button/ButtonGroup internals and prevent internal props from reaching DOM elements or separators.
- Add ButtonGroup stories covering Tooltip and Dropdown wrappers.
- Fixes #6516

## Verification
- `pnpm typecheck:react`
- `pnpm lint:react`
- `pnpm build`
- SSR regression checks for every supported wrapper, nested and fragment forwarding, and internal-prop leakage.

### DIFF
--- a/packages/react/src/components/alert-dialog/alert-dialog.tsx
+++ b/packages/react/src/components/alert-dialog/alert-dialog.tsx
@@ -19,6 +19,7 @@ import {
   Pressable as PressablePrimitive,
 } from "react-aria-components/Modal";
 
+import {enableChildProps, forwardChildProps} from "../../utils/children";
 import {composeSlotClassName, composeTwRenderProps} from "../../utils/compose";
 import {dom} from "../../utils/dom";
 import {CloseButton} from "../close-button";
@@ -43,7 +44,8 @@ const AlertDialogContext = createContext<AlertDialogContext>({});
  * -----------------------------------------------------------------------------------------------*/
 interface AlertDialogRootProps extends ComponentPropsWithRef<typeof AlertDialogTriggerPrimitive> {}
 
-const AlertDialogRoot = ({children, ...props}: AlertDialogRootProps) => {
+const AlertDialogRoot = enableChildProps((inputProps: AlertDialogRootProps) => {
+  const {children, ...props} = forwardChildProps(inputProps);
   const alertDialogContext = useMemo<AlertDialogContext>(
     () => ({slots: alertDialogVariants(), placement: undefined}),
     [],
@@ -56,7 +58,7 @@ const AlertDialogRoot = ({children, ...props}: AlertDialogRootProps) => {
       </AlertDialogTriggerPrimitive>
     </AlertDialogContext>
   );
-};
+});
 
 /* -------------------------------------------------------------------------------------------------
  * AlertDialog Trigger
--- a/packages/react/src/components/button-group/button-group.stories.tsx
+++ b/packages/react/src/components/button-group/button-group.stories.tsx
@@ -8,6 +8,7 @@ import {Chip} from "../chip";
 import {Description} from "../description";
 import {Dropdown} from "../dropdown";
 import {Label} from "../label";
+import {Tooltip} from "../tooltip";
 
 import {ButtonGroup} from "./";
 
@@ -294,6 +295,21 @@ export const WithoutSeparator: Story = {
       <Button>First</Button>
       <Button>Second</Button>
       <Button>Third</Button>
+    </ButtonGroup>
+  ),
+};
+
+export const WithTooltips: Story = {
+  render: () => (
+    <ButtonGroup size="sm" variant="secondary">
+      <Tooltip>
+        <Button>First</Button>
+        <Tooltip.Content>First action</Tooltip.Content>
+      </Tooltip>
+      <Tooltip>
+        <Button>Second</Button>
+        <Tooltip.Content>Second action</Tooltip.Content>
+      </Tooltip>
     </ButtonGroup>
   ),
 };
--- a/packages/react/src/components/button-group/button-group.tsx
+++ b/packages/react/src/components/button-group/button-group.tsx
@@ -12,6 +12,7 @@ import {useSlottedContext} from "react-aria-components/slots";
 import {ToggleButtonGroupContext as RACToggleButtonGroupContext} from "react-aria-components/ToggleButtonGroup";
 
 import {composeSlotClassName, composeTwRenderProps} from "../../utils";
+import {supportsChildProps, withChildProps} from "../../utils/children";
 import {dom} from "../../utils/dom";
 
 /* -------------------------------------------------------------------------------------------------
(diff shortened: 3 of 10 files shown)


wingkwong (changes requested):
please verify locally once first before making a PR. The changes failed all QA.

wanxiankai:
Thanks for checking. You're right — the first version leaked the internal ButtonGroup marker into the exported Tooltip type, which caused the QA TypeScript/build failures:

`TS4023: Exported variable 'Tooltip' has or is using name 'BUTTON_GROUP_CHILD'...`

I pushed `5fb754d` to keep the marker prop internal to the Tooltip implementation while preserving the same behavior. The new QA run is currently marked `action_required` before jobs execute, so I can't see the updated TypeScript/build result yet from my fork. Could you approve/re-run the workflow when you have a chance?

wingkwong (changes requested):
You may verify by running `pnpm typecheck` and `pnpm build` first.

wanxiankai:
Thanks — I reproduced the issue locally on 5fb754d. `pnpm typecheck` still failed with the same TS4023 because TooltipRoot's inferred function signature included the internal computed marker prop.

I pushed d24d501 to explicitly expose TooltipRoot as `React.FC<TooltipRootProps>` while keeping `ButtonGroupChildProps` only in the implementation signature. The generated declaration now contains `declare const TooltipRoot: React.FC<TooltipRootProps>;` and no `BUTTON_GROUP_CHILD` reference.

Local verification now passes:
- `pnpm typecheck`
- `pnpm build`
- `pnpm lint:react`
- `pnpm exec prettier --check packages/react/src/components/tooltip/tooltip.tsx`

The new QA workflow run 30183835766 completed with no jobs, so it still needs approval/re-run from a maintainer.

wingkwong (changes requested):
The scope is too narrow as you only cover Tooltip. Also it is a generic wrapper and should not know anything about Button or ButtonGroup internals. If the intent is "transparent wrappers forward unknown props to their first child", it should be implemented generically without importing Button.

wanxiankai:
Thanks for the feedback. Updated in 2a8a7ec.

- Removed Tooltip's Button/ButtonGroup imports and special-case traversal.
- Added a generic internal child-prop forwarding mechanism for transparent trigger roots.
- Applied it across Tooltip, Dropdown, Popover, Modal, Drawer, and AlertDialog, with Button consuming the forwarded props.
- ButtonGroup only marks opted-in direct children, preserving direct-child semantics and preventing internal props from reaching DOM elements or separators.
- Added ButtonGroup stories covering Tooltip and Dropdown wrappers.

Verified with `pnpm typecheck:react`, `pnpm lint:react`, `pnpm build`, and SSR checks covering every supported wrapper, nested wrappers, fragments, and internal-prop leakage.

Could you please take another look?

Issue #6021, opened 2025-12-20
[BUG] - ButtonGroup size prop incorrectly affects nested Buttons


RezaNajafian:
### HeroUI Version

3.0.0-beta.3

### Describe the bug

There is an issue with ButtonGroup when using the size prop.

When size is applied to <ButtonGroup>, all nested <Button> components inside child components (like Modals) are also forced to render in the ButtonGroup size.
This behavior is incorrect, the size should only affect the direct buttons of the ButtonGroup, not buttons rendered deeper in the component tree.

### Steps to Reproduce the Bug or Issue

1. Create a `ButtonGroup` with `size="sm"`.
2. Inside the `ButtonGroup`, render a component that contains a `Button` (e.g. a Modal trigger).
3. Inside that component, render a `Button` (such as a button inside a modal).
4. Open the modal.

### Expected behavior

Buttons rendered inside nested components (like modals) should retain their own default size.
Nested buttons should not inherit the small size unintentionally from the ButtonGroup.

### Operating System Version

Windows

### Browser

Edge

RezaNajafian:
It’s better to say that every Button-related prop, like variant, is also affected

PR #6740, opened 2026-07-29
fix(button-group): stop internal child marker from reaching the DOM

Saidheerajgollu:
No existing issue — found while investigating #6516.

## 📝 Description

`ButtonGroup` stamps the internal `__button_group_child` marker onto every valid element child via `Children.map` + `cloneElement`. Children that forward unknown props to a host element cannot consume it, so the marker reaches the DOM and React logs a warning.

This affects HeroUI's own `ButtonGroup.Separator` when used as a direct sibling, and any plain host child (e.g. a `<span>`).

```tsx
<ButtonGroup size="sm" variant="secondary">
  <Button>one</Button>
  <ButtonGroup.Separator />
  <Button>two</Button>
</ButtonGroup>
```

logs:

```
Received `true` for a non-boolean attribute `__button_group_child`.

If you want to write it to the DOM, pass a string instead:
__button_group_child="true" or __button_group_child={value.toString()}
```

`ButtonGroupSeparator` only destructures `className` and spreads the rest onto `dom.span`, so the marker lands on the DOM node. A plain `<span>` child hits the same path.

## 🚀 New behavior

No warning. The marker is only stamped on children that can actually consume it — host elements and `ButtonGroup.Separator` are skipped.

Only `Button` reads the marker, so `size` / `variant` / `isDisabled` / `fullWidth` inheritance is unchanged. Verified by rendering before/after and diffing the emitted markup:

| case | before | after |
|---|---|---|
| direct `<Button>` child | `button--sm button--secondary` | unchanged |
| `<Button>` behind a wrapper component | not inherited (see #6021) | unchanged |
| `<ButtonGroup.Separator />` sibling | warning | no warning, same markup |
| `<span>` child | warning | no warning, same markup |

No. Behaviour is identical apart from the removed warning; no public API changes.

## 📝 Additional Information

- `pnpm build`, `pnpm lint`, `pnpm typecheck` and `prettier --check` all pass.
- Found while investigating #6516. That issue needs a design decision (it conflicts with #6021), so I've left the analysis in a [comment there](https://github.com/heroui-inc/heroui/issues/6516#issuecomment-5124482584) and kept this PR to the part that is a bug under either outcome.
- Related, not touched here: `TOGGLE_BUTTON_GROUP_CHILD` is declared and exported in `toggle-button-group.tsx` but never referenced. It's a public export, so removing it would be breaking — happy to follow up if you'd like it deprecated.

### DIFF
--- a/packages/react/src/components/button-group/button-group.tsx
+++ b/packages/react/src/components/button-group/button-group.tsx
@@ -66,6 +66,12 @@ const ButtonGroupRoot = ({
       return child;
     }
 
+    // Host elements and the separator forward unknown props straight to the DOM, where React
+    // warns about the non-boolean attribute. Only stamp children that can consume the marker.
+    if (typeof child.type === "string" || child.type === ButtonGroupSeparator) {
+      return child;
+    }
+
     // Clone the child and add the special prop
     return React.cloneElement(child, {
       [BUTTON_GROUP_CHILD]: true,


wingkwong:
The supported pattern is to place `ButtonGroup.Separator` inside a `Button`, not as a direct sibling of `ButtonGroup` as shown in [Anatomy](https://heroui.com/en/docs/react/components/button-group#anatomy). When nested inside the Button, the separator is never a direct child, so it never gets stamped with `__button_group_child` and no warning is produced. 

```tsx
<ButtonGroup>
  <Button>First</Button>
  <Button>
    <ButtonGroup.Separator />
    Second
  </Button>
</ButtonGroup>
```

This PR blacklists host elements and `ButtonGroupSeparator`, but any other child that forwards unknown props to a host element still leaks the marker, e.g. a user's own passthrough wrapper component, or a Fragment direct child etc. So the underlying problem remains.

PR #6752, opened 2026-08-01
chore(deps): react aria upgrade (v1.20.0)

wingkwong:
Closes # 

## 📝 Description

https://react-aria.adobe.com/releases/v1-20-0


andrew-shpak:
Is there plan to support v2 version and update version there as well?

andrew-shpak:
if yes, when we could expect new v2 version ?

PR #6768, opened 2026-08-05
fix(styles): prevent subpixel border overflow on badge

tianenpang:
Closes [Discord 🧵](https://discord.com/channels/856545348885676062/1534632541028614354)

## 📝 Description

Badge uses a 1px border with `var(--badge-border)` as an outline. In Firefox, this can show a thin fringe of color outside the badge because of subpixel anti-aliasing on rounded corners.

## ⛳️ Current behavior

A thin line of the background color appears outside the badge edge in Firefox.

## 🚀 New behavior

Add [`background-clip`](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/background-clip) that keeps the background inside the padding box and removes the fringe, without changing size or look.

No.

## 📝 Additional Information

Only Badge uses a background-colored 1px border this way. Other components with 1px + border-radius may have similar issues, but not the same pattern.

#### Before

before

#### After

after

### DIFF
--- a/packages/styles/components/badge.css
+++ b/packages/styles/components/badge.css
@@ -17,6 +17,7 @@
   background-color: var(--badge-bg);
   color: var(--badge-fg);
   border: 1px solid var(--badge-border);
+  background-clip: padding-box;
 }
 
 .badge__label {


jrgarciadev:
Thanks! @tianenpang

PR #6795, opened 2026-08-16
fix(toast): preserve region CSS variables when custom style prop is passed to ToastProvider (#6793)

### LINKED ISSUE #6793: [v3] ToastProvider: a custom `style` prop silently discards the region's CSS variables (toast collapses to icon width)
# [v3] ToastProvider: a custom `style` prop silently discards the region's CSS variables (toast collapses to icon width)

## Summary

`Toast.Provider` writes its four region CSS variables (`--toast-width`, `--gap`, `--scale-factor`, `--placement`) into an inline `style` object, then spreads `{...rest}` **after** it — so any user-supplied `style` prop replaces the whole object instead of merging with it. Every toast then renders with `--toast-width` unset: `.toast-region` computes `width: auto` with no effective `min-width`, and toasts collapse to roughly icon width (~32px) with the text wrapping one word per line.

## Reproduction

```tsx
import { Toast, toast } from "@heroui/react";

  return (
    <>
      {/* Any style at all — e.g. a z-index needed to clear a host page's chrome */}
      <Toast.Provider placement="bottom" style={{ zIndex: 100_000 }} />
      <button onClick={() => toast.success("Operation completed")}>fire</button>
    </>
  );
}
```

Fire a toast → it renders ~32px wide, bottom-center. Inspect `.toast-region`: its inline `style` contains only `z-index: 100000` — `--toast-width`/`--gap`/`--scale-factor`/`--placement` are gone. Remove the `style` prop and the same toast renders at the default 460px.

Observed on `@heroui/react` 3.2.1 (macOS, Chrome); the spread order is unchanged on the current `v3` branch.

## Cause

`packages/react/src/components/toast/toast.tsx`, in `ToastProvider`:

```tsx
<ToastRegionPrimitive<T>
  className={composeTwRenderProps(className, slots?.region())}
  data-slot="toast-region"
  queue={toastQueue}
  style={{
    // @ts-expect-error - CSS variables
    "--gap": `${gap}px`,
    "--placement": placement,
    "--scale-factor": scaleFactor,
    "--toast-width": typeof width === "number" ? `${width}px` : width,
  }}
  {...rest}   // <-- rest.style replaces the object above wholesale
>
```

Because `rest` is spread after the `style` attribute, React takes `rest.style` as the final value. This also means the `width` prop is silently ignored whenever a `style` is passed, since `width` only takes effect through the discarded variable.

## Suggested fix

Merge instead of replace — the same file already does this correctly in the `Toast` component's own computed style (`...rest.style` is spread **into** the computed object as the last member):

```tsx
const {children, className, gap = DEFAULT_GAP, ..., style, ...rest} = props;

<ToastRegionPrimitive<T>
  ...
  style={{
    "--gap": `${gap}px`,
    "--placement": placement,
    "--scale-factor": scaleFactor,
    "--toast-width": typeof width === "number" ? `${width}px` : width,
    ...style,
  }}
  {...rest}
>
```

## Workaround (for anyone else hitting this)

Restate the variables inside your own `style`:

```tsx
<Toast.Provider
  placement="bottom"
  style={{
    zIndex: 100_000,
    "--toast-width": "460px",
    "--gap": "8px",
    "--scale-factor": 0.05,
  } as React.CSSProperties}
/>
```

RohithPariki:
## Description
Fixes #6793.

Currently, if a developer passes a custom `style` prop to `ToastProvider` (e.g. `style={{ zIndex: 100 }}`), it completely overrides the dynamically generated CSS variables on the `ToastRegionPrimitive` (such as `--gap`, `--placement`, `--toast-width`, and `--scale-factor`). This results in visually broken toast layouts where toasts lose their intended spacing, width, and positioning.

This PR fixes the issue by destructuring the `style` prop from `ToastProviderProps` and merging it gracefully alongside the internal CSS variables.

## Changes Made
- Destructured `style` from `ToastProviderProps` to prevent it from bleeding into the `...rest` spread.
- Spread `...style` inside the computed `style={{ ... }}` object passed to `ToastRegionPrimitive`, ensuring both the user's custom styles and the internal CSS variables coexist.

## Testing
- [x] Verified in a local sandbox that passing `style={{ zIndex: 999 }}` successfully renders while maintaining the `--gap` and `--placement` CSS variables in the DOM.
- [x] Ensured TypeScript typings remain strict and do not throw any regression errors.

### DIFF
--- a/packages/react/src/components/toast/toast.tsx
+++ b/packages/react/src/components/toast/toast.tsx
@@ -370,6 +370,7 @@ const ToastProvider = <T extends object = ToastContentValue>({
   placement = "bottom",
   queue: queueProp,
   scaleFactor = DEFAULT_SCALE_FACTOR,
+  style,
   width = DEFAULT_TOAST_WIDTH,
   ...rest
 }: ToastProviderProps<T>) => {
@@ -469,6 +470,7 @@ const ToastProvider = <T extends object = ToastContentValue>({
         "--placement": placement,
         "--scale-factor": scaleFactor,
         "--toast-width": typeof width === "number" ? `${width}px` : width,
+        ...style,
       }}
       {...rest}
     >


wingkwong:
Closing - there is already a PR for this issue.

PR #6817, opened 2026-08-31
fix(accordion): stop the default trigger hover flashing over a custom hover background

### LINKED ISSUE #6815: [BUG] - Accordion Component shows blinking onMouseLeave when to change to custom hover color
### HeroUI Version

3.2.4

### Describe the bug

When I use className and add class `hover:bg-surface` it show visual blinking with color default hover on **Accodion.Trigger** Component.

### Code
```tsx
<Accordion.Trigger className="hover:bg-surface">
{item.Icon ? (
<span className="me-3 size-4 shrink-0 text-accent">{<item.Icon weight={"bold"} />}</span>
) : null}
{item.title}
<Accordion.Indicator>
<CaretDownIcon/>
</Accordion.Indicator>
</Accordion.Trigger>
```

_No response_

1. Use className="hover:bg-surface"
2. Hover On Accordion.Trigger Component.
3. Leave the Mouse Cursor from Component.
4. See the default hover bg-color of Accordian Surface. 

### Expected behavior

I want not show any hover effect on it or same as surface color.

### Screenshots or Videos

https://github.com/user-attachments/assets/aa0416ab-6f62-4f31-9b4d-12a188378cff

### Same behaviour on docs page 

[Docs Accordion Link](https://heroui.com/en/docs/react/components/accordion#tailwind-css) in Customization Example

### Operating System Version

- Windows 11

### Browser

Chrome

jrgarciadev:
Closes #6815

## 📝 Description

`Accordion.Trigger` treated React Aria's `data-hovered` attribute and the native `:hover` pseudo-class as interchangeable hover signals. React Aria writes `data-hovered` from a React render (pointer events run at continuous priority, so the update is scheduled rather than flushed synchronously), which means the attribute survives at least one frame longer than `:hover` after the pointer leaves. During that frame a user's own `hover:bg-*` has already stopped matching while the library's default hover rule still matches, so the default background paints on top of the custom one — the blink reported in the issue and visible in the docs [Customization example](https://heroui.com/en/docs/react/components/accordion#tailwind-css).

The fix requires both signals on the trigger, in the default and `surface` variants, which ties the default background to the pointer actually being over the element. No token allowlist and no change to the component or to the docs demo.

`<Accordion.Trigger className="hover:bg-surface">` shows the default hover color for a frame on mouse leave.

Sampled every animation frame in Chromium against the docs customization markup (`transition-none hover:bg-surface`, `surface` variant):

```
t= 420  :hover=true   data-hovered=true   bg=oklch(1 0 0)              <- custom hover
t= 437  :hover=false  data-hovered=true   bg=oklch(0.94 0.001 286.375) <- default hover flash
t= 453  :hover=false  data-hovered=null   bg=rgba(0, 0, 0, 0)
```

## 🚀 New behavior

The default hover background is dropped the moment the pointer leaves, so the custom background is the last thing painted:

```
t= 426  :hover=true   data-hovered=true   bg=oklch(1 0 0)      <- custom hover
t= 444  :hover=false  data-hovered=true   bg=rgba(0, 0, 0, 0)  <- no flash
t= 460  :hover=false  data-hovered=null   bg=rgba(0, 0, 0, 0)
```

Default (non-overridden) accordion hover is unchanged, and combining the signals also drops the sticky `:hover` a touch tap leaves behind on hybrid pointer devices, since React Aria never reports hover for `pointerType: "touch"`.

No. `data-hovered` stays in the selector and remains available for custom styling; only the library's own default background now additionally requires the pointer to be over the trigger. On hover-capable pointers the default background appears one render tick after `:hover`, the same tick every other React Aria driven state uses.

## ✅ Testing

- [x] Interactive / a11y contract changes include or update `*.test.tsx` coverage
- [x] Scenarios cover roles, `data-*` states, keyboard, disabled, and callbacks as applicable
- [x] `pnpm test` passes locally (123 files, 646 tests)
- [x] `pnpm lint` passes (2 pre-existing docs `import/order` warnings)
- [x] `pnpm typecheck` passes

New `packages/react/tests/components/accordion/accordion.browser.test.tsx` (Playwright/Chromium against the compiled Tailwind sources) covers three cases: a custom `hover:bg-surface` wins while hovered, a stale `data-hovered` with the pointer away paints no default background, and the default trigger still gains and loses its hover background. The middle case is the regression guard and fails on `v3.2.5` without this change:

```
AssertionError: expected 'oklch(0.94 0.001 286.375)' to be 'rgba(0, 0, 0, 0)'
```

## 📝 Additional Information

Recorded in Storybook with the docs customization classes on the trigger. The overlay samples the trigger every frame and counts frames where the default hover background is painted after the pointer has left; CPU is throttled 6× so the frame lands in the 25 fps capture. Before: 3 flashed frames across 3 hovers. After: 0, with the un-overridden accordion below still hovering normally.

<a href="https://cursor.com/agents/bc-67443664-e0c9-44d2-92b8-c32cd1081d3d/artifacts?path=%2Fopt%2Fcursor%2Fartifacts%2Faccordion_custom_hover_no_blink.mp4">accordion_custom_hover_no_blink.mp4</a>

Before: pointer has left, :hover is false, data-hovered is still true, and the trigger paints the default gray hover

After: the un-overridden accordion still shows its default hover background

<div><a href="https://cursor.com/agents/bc-67443664-e0c9-44d2-92b8-c32cd1081d3d?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-67443664-e0c9-44d2-92b8-c32cd1081d3d&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/packages/react/tests/components/accordion/accordion.browser.test.tsx
+++ b/packages/react/tests/components/accordion/accordion.browser.test.tsx
@@ -1,3 +1,5 @@
+import type {ComponentProps} from "react";
+
 import {render} from "@heroui/testing/browser";
 import {page, userEvent} from "vitest/browser";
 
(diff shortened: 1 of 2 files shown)


jrgarciadev:
@wingkwong pls review

quddus-larik:
This PR works successfully with my bug report. Thanks for the quick fix!

PR #6847, opened 2026-09-12
fix(react): correct dependency declarations and production warning leak

wingkwong:
## Summary

Four independent correctness defects in packaging and one runtime leak.

**`@internationalized/date` was imported but never declared.** It's used by real runtime source — `src/utils/calendar.ts` and three `calendar/*.tsx` files, not just stories — while only appearing in `devDependencies`. Promoted to a peer alongside the existing `@react-aria/ssr` / `@react-aria/utils` peers, which follow the same peer+dev pattern. This costs consumers nothing: `react-aria-components@1.21.0` (already a required peer) depends on `@internationalized/date@^3.12.4` directly, so it is always present at the declared range. The declaration exists to prevent a second copy and the `CalendarDate` type mismatch that follows from one.

**Two dependencies were declared but unused.** Neither `tailwind-merge` nor `@react-aria/i18n` is imported anywhere in `src/` or `tests/`. `tailwind-merge` reaches us transitively as an optional peer of `tailwind-variants`; `@react-aria/i18n` was additionally declared as a peer, forcing consumers to install a package the library never touches. Removing it also drops the whole `@formatjs/*` tree from the lockfile.

**Export conditions were ordered wrong.** `import` preceded `types`, so resolvers under `moduleResolution: node16`/`nodenext` could pick the runtime target when asked for types, and no `default` condition existed at all. `update-exports.mjs` now emits `types` first and `default` last. Confirmed against the packed tarball: **all 87 condition objects** are correctly ordered.

**ColorSlider leaked dev warnings into production.** `getValidColorSpace` runs in the render body, so a misconfigured slider fired `console.warn` on *every* render in production builds. Now routed through a `warnInDev` guard.

Finally, `tag.css` did not follow the repo's default-size convention: the base `.tag` carried no sizing and `.tag--md` held it. Moved the defaults into the base and emptied `.tag--md`.

## Test plan

- [x] `pnpm install --frozen-lockfile` clean — CI's install won't reject the lockfile
- [x] Packed tarball inspected: 87/87 export conditions ordered, peers updated, `tailwind-merge` absent from `dependencies`
- [x] Confirmed `@react-aria/i18n` and `tailwind-merge` have zero imports in `src/`/`tests/`
- [x] Single `tailwind-merge` version in the lockfile after alignment
- [x] Full `apps/docs` build from scratch (no cache) on 3.6.0
- [x] 614 jsdom + 42 browser tests, including the 5 `tag.browser.test.tsx` cases that assert against the real compiled stylesheet
- [x] 27 docs tests, `pnpm typecheck`, `pnpm lint`, `pnpm build`

### DIFF
--- a/apps/docs/package.json
+++ b/apps/docs/package.json
@@ -61,7 +61,7 @@
     "remark-gfm": "4.0.1",
     "remark-mdx": "3.1.1",
     "shiki": "3.20.0",
-    "tailwind-merge": "3.4.0",
+    "tailwind-merge": "3.6.0",
     "tailwind-variants": "3.3.1",
     "ts-morph": "27.0.2",
     "usehooks-ts": "3.1.1",
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -67,7 +67,6 @@
     "@react-types/color": "3.2.0",
     "@react-types/shared": "3.36.1",
     "input-otp": "1.4.2",
-    "tailwind-merge": "3.4.0",
     "tailwind-variants": "3.3.1"
   },
   "devDependencies": {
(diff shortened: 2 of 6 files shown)


jrgarciadev:
@wingkwong pls fix conflicts

PR #6880, opened 2026-09-18
fix(checkbox): prevent checkmark flash when clearing indeterminate state

### LINKED ISSUE #6881: Table check box indeterminate state animation shows wrong check mark animation
https://github.com/user-attachments/assets/13091e30-02b6-4091-9129-02bc4c4e2beb

repro:

https://heroui.com/en/docs/react/components/table#selection

giperfast:
## 📝 Description

Fixes the checkbox indicator rendering when transitioning from the indeterminate state to the unchecked state.
Previously, the default checkmark icon was rendered whenever the checkbox was not indeterminate, including when it was unchecked. This could cause the checkmark to briefly appear when clearing an indeterminate state, such as when deselecting the last selected row in a multi-select Table.

closes: #6881

When a checkbox transitions from indeterminate to unchecked, the indeterminate icon is replaced by the checkmark icon before the indicator disappears.
In a multi-select Table, this can be observed by:
1. Selecting a single row so that the header checkbox becomes indeterminate.
2. Deselecting that row.
3. The header checkbox briefly displays a checkmark even though no rows are selected.

## 🚀 New behavior

The default checkmark icon is rendered only when the checkbox is selected.
When transitioning from indeterminate to unchecked, the checkmark is no longer displayed.

No

## ✅ Testing

- [ ] Interactive / a11y contract changes include or update `*.test.tsx` coverage
- [ ] Scenarios cover roles, `data-*` states, keyboard, disabled, and callbacks as applicable
- [ ] `pnpm test` passes locally

## 📝 Additional Information

This change is limited to the default Checkbox indicator rendering and does not affect custom indicators provided through children.


wingkwong (changes requested):
Thanks for the PR. Here's my comments.
1. CI is red. Please fix it.
2. `strokeDashoffset={66}` on the unchecked checkmark isn't dead weight. It's the start frame of the draw animation. With this change, the svg is absent when unchecked, so checking the box mounts it already at `44`. CSS transitions don't run on a freshly mounted element, so the checkmark now pops in instantly instead of drawing (and pops out instead of un-drawing).
3. Also need to handle this case: Select a few > toggle the checkbox in indeterminate state in header row > you will see it is still flashing

giperfast:
The checkmark animation is preserved now, and the flashing during indeterminate state transitions has been fixed. Also fixed the formatting issues that caused CI to fail.

packages/styles/components/checkbox.css
```diff
@@ -150,6 +165,10 @@
   /* Indeterminate */
   .checkbox[data-indeterminate="true"] & {
     @apply bg-accent text-accent-foreground;
+
+    &::before {
+      @apply scale-100 opacity-100;
+    }
```
wingkwong, line 171: adding this puts an opaque `inset-0` layer over the control's own background

packages/styles/components/checkbox.css
```diff
@@ -187,7 +206,11 @@
   /* Invalid + Indeterminate */
   .checkbox[data-indeterminate="true"][aria-invalid="true"] &,
   .checkbox[data-indeterminate="true"][data-invalid="true"] & {
-    @apply bg-danger text-danger-foreground;
+    @apply text-danger-foreground;
+
+    &::before {
+      @apply bg-danger;
+    }
```
wingkwong, line 217: Static looks fine but leaving indeterminate fades the red `::before` over a still-fading blue background. Also inconsistent with Invalid + Selected, which keeps `bg-danger` on both.

packages/styles/components/checkbox.css
```diff
@@ -199,7 +222,7 @@
   @apply relative z-10 flex size-3 items-center justify-center;
 
   & svg {
-    @apply size-full transform-gpu;
+    @apply absolute size-full transform-gpu;
```
wingkwong, line 229: why `absolute` is required here?

wingkwong (changes requested):
`data-slot="checkbox-default-indicator--indeterminate"` is now always in the DOM

packages/styles/components/checkbox.css
giperfast: Both default SVGs now stay mounted so the checkmark animation is preserved. `absolute` overlays them in the same position and removes them from the flex layout; without it, the two SVGs would be laid out next to each other inside the indicator. Their visibility is then controlled with opacity.

packages/styles/components/checkbox.css
giperfast: Would you prefer the indeterminate state to keep using the controls own background, or should it use the `::before` layer consistently with the selected state? Removing `bg-accent` from the control would avoid duplicating the backgrounds, while keeping both transitions on the same layer

